Issue #19, opened 2024-12-13
Font & Layout Settings  -> Misc : Chinese Font Name are not supported.


hustwenchao:
![image](https://github.com/user-attachments/assets/56a8d07e-51b0-4c4c-913a-519f52b8187c)

hustwenchao:
I just pr my solution. Please check it.

chrox:
Awesome. Already merged.

PR #20, opened 2024-12-13
Custom CSS Add Non-ASCII characters Supporting

hustwenchao:
#19


chrox:
@hustwenchao I noticed that your commit is not signed. Could you setup the commit signing locally for the potential following commits? You can refer to this [Github documentation about commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification).

Issue #16, opened 2024-12-13
Feature Request: config option to disable click navigation


kratosmy:
I don't really want to use mouse to control my reading and sometimes a unintentional click will cause the page changes.

AdkinsHan:
i met the same question
there seems a conflict between mouse click and scrolling pages
when i double click a word intend to select the word but it scroll to other paragraph

chrox:
Implemented in #23. You can toggle the 'Disable Click-to-Flip' option in the Misc tab of the Config Dialog to disable click navigation. Closing this now.

![image](https://github.com/user-attachments/assets/a551bd5e-2c47-4bba-8ac4-0b38af0f8112)

PR #23, opened 2024-12-15
Add 'Disable Click' feature to view settings

zangjiucheng:
## Description
This PR introduces the functionality to disable single-click actions in the ebook reader. The following changes have been made:

- Added a new setting disableClick to the BookLayout interface and DEFAULT_BOOK_LAYOUT.
- Updated FoliateViewer.tsx to check for the disableClick setting and prevent single-click actions if enabled.
- Enhanced ViewMenu.tsx to include a menu item that allows users to toggle the disableClick setting.
- Implemented state management for disableClick in the ViewMenu component and ensured changes are persisted in the view settings.

## Related Issues
#16 

## Testing
- Verified that the disableClick setting can be toggled through the ViewMenu.
- Confirmed that single-click actions are disabled when the setting is enabled and re-enabled when the setting is disabled.

## Known Issue
- The disable setting for the edge area (corner) is not functioning as intended.

### DIFF
--- a/apps/readest-app/src/app/reader/components/ViewMenu.tsx
+++ b/apps/readest-app/src/app/reader/components/ViewMenu.tsx
@@ -28,7 +28,6 @@ const ViewMenu: React.FC<ViewMenuProps> = ({
   const { themeMode, isDarkMode, themeCode, updateThemeMode } = useTheme();
   const [isScrolledMode, setScrolledMode] = useState(viewSettings!.scrolled);
   const [isInvertedColors, setInvertedColors] = useState(viewSettings!.invert);
-  const [isDisableClick, setIsDisableClick] = useState(viewSettings!.disableClick);
   const [zoomLevel, setZoomLevel] = useState(viewSettings!.zoomLevel!);
 
   const zoomIn = () => setZoomLevel((prev) => Math.min(prev + 10, 200));
@@ -36,7 +35,6 @@ const ViewMenu: React.FC<ViewMenuProps> = ({
   const resetZoom = () => setZoomLevel(100);
   const toggleScrolledMode = () => setScrolledMode(!isScrolledMode);
   const toggleInvertedColors = () => setInvertedColors(!isInvertedColors);
-  const toggleDisableClick = () => setIsDisableClick(!isDisableClick);
 
   const openFontLayoutMenu = () => {
     setIsDropdownOpen?.(false);
@@ -85,14 +83,6 @@ const ViewMenu: React.FC<ViewMenuProps> = ({
     setViewSettings(bookKey, viewSettings!);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [zoomLevel]);
-
-  useEffect(() => {
-    const view = getView(bookKey);
-    if (!view) return;
-    viewSettings!.disableClick = isDisableClick;
-    setViewSettings(bookKey, viewSettings!);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isDisableClick]);
 
   return (
     <div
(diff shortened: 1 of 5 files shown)


apps/readest-app/src/app/reader/components/ViewMenu.tsx
```diff
@@ -135,6 +145,13 @@ const ViewMenu: React.FC<ViewMenuProps> = ({
         onClick={toggleScrolledMode}
       />
 
+      <MenuItem
+          label='Disable Click'
+          shortcut='Shift+D'
+          icon={isDisableClick ? <MdCheck size={20} /> : undefined}
+          onClick={toggleDisableClick}
+      />
+
```
chrox: How about moving the config item to the config dialog just after the "Animation" settings [here](https://github.com/chrox/readest/blob/main/apps/readest-app/src/app/reader/components/settings/MiscPanel.tsx#L78-L93)?

It's an infrequently used function, so I want to keep it from drawing too much attention from users.

chrox: You can add a section called "Behavior" with an toggle of "Disable Click-to-Flip" just as the "Paging Animation".

zangjiucheng: Thank you for your guidance. I will adhere to your instructions. As a beginner programmer, I appreciate any potential issues you may foresee :) Thanks!

apps/readest-app/src/app/reader/components/settings/MiscPanel.tsx
```diff
@@ -73,6 +74,12 @@ const MiscPanel: React.FC<{ bookKey: string }> = ({ bookKey }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [animated]);
 
+  useEffect(() => {
+    viewSettings!.disableClick = isDisableClick;
+    setViewSettings(bookKey, viewSettings!);
```
chrox: Also need this as the `animated` logic: 
```
if (isFontLayoutSettingsGlobal) {
      settings.globalViewSettings.disableClick = disableClick;
      setSettings(settings);
}
```
otherwise it only applies to current Book.

chrox: And `viewSettings` is assured to be valid above:
```
const viewSettings = getViewSettings(bookKey)!;
```
so no need to  re-ensure it like this `viewSettings!.disableClick = isDisableClick;` just
`viewSettings.disableClick = isDisableClick;` should be OK. It's better to use the `JavaScript and TypeScript Nightly` extension in VSCode. Generally it works fine to deduce when to use the non-null assertion(!).

chrox (changes requested):

apps/readest-app/src/app/reader/components/settings/MiscPanel.tsx
zangjiucheng: Thanks :)

apps/readest-app/src/app/reader/components/settings/MiscPanel.tsx
```diff
@@ -71,7 +74,7 @@ const MiscPanel: React.FC<{ bookKey: string }> = ({ bookKey }) => {
       getView(bookKey)?.renderer.removeAttribute('animated');
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [animated]);
+  }, [animated,isDisableClick]);
```
chrox: Better to have two useEffect blocks, one for `animated` and one for `isDisableClick` in order to de-couple the two logics.

chrox (changes requested):

apps/readest-app/src/app/reader/components/settings/MiscPanel.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { use, useEffect, useState } from 'react';
```
chrox: This `use` is miss-imported by the VSCode. It's not used actually.

chrox (changes requested):

chrox (approved):

zangjiucheng:
Thanks for your help to point out those issue :)

chrox:
Good. Already merged. :)

PR #37, opened 2024-12-25
Add back navigation button and fix dev hydration failed.

Ray-D-Song:
- The auth page lacks an explicit back button, making it impossible to return on the Tauri App.
- @supabase/auth-ui-react cause hydration error in development mode.

image


chrox (approved):
LGTM

PR #46, opened 2024-12-26
Add i18n support

chrox:


chrox:
There is also a wiki guide to use i18n translation in Readest: https://github.com/readest/readest/wiki/Readest-i18n-Guide.

PR #80, opened 2025-01-02
Fix typos discovered by codespell

cclauss:
https://github.com/codespell-project/codespell


houqp:
could be a good idea to add the codespell check to CI too :)

chrox:
> could be a good idea to add the codespell check to CI too :)

This could be awesome to add codespell check in CI.

Issue #53, opened 2024-12-27
FR: Support changing between horizontal and vertical writing direction for books in CJK languages


chrox:
**Does your feature request involve difficulty completing a task? Please describe.**
It might be useful to support changing between horizontal and vertical writing direction for books in CJK languages. Currently only epub books with the `writing-mode: vertical-rl` defined in the CSS are supported.

**Describe the solution you'd like**
Add an option to switch between horizontal and vertical writing direction for books in CJK languages.

PR #91, opened 2025-01-03
Add vertical/horizontal switch in Layout configs for CJK books

chrox:


chrox:
This should closes #53.

PR #133, opened 2025-01-09
Better animation for sidebar transitions

xhuajin:
1. Fix Sidebar BookNoteItem hover animation: text like 'edit' and 'delete' will display before background arrive the area.
2. Add Fade in and out animation for SideBar


xhuajin:
I recorded what it looks like now

PixPin_2025-01-09_17-41-16

chrox:
LGTM. I will review it soon.

chrox:
@xhuajin This looks nice. Since the translation key is used as English translation. It's better to use the `TOC` instead of `Toc` as the key and English translation since `TOC` is more readable. Except that all is fine.

xhuajin:
Thank you for liking the changes. Now it's all `Table of Contents` instead of `Toc`.

PR #135, opened 2025-01-09
enforce numeric inputs on NumberInput and prevent NaN text

eg9y:
Cases:

- set to 0 when removing 1 digit number
- don't set any input when typing non-numeric character
- prevent 0 from being the last digit (left-most)

Note:
- Does not allow negative numbers and decimals


chrox:
LGTM.

PR #153, opened 2025-01-14
feat: provide reading progress indicator for each books in the bookshelf

eg9y:
image

This is one of the PRs for the reading progress features
- **Percentage reading progress per book**
  - For each book in the gallery, show the % reading progress


apps/readest-app/src/hooks/useBookConfigLoader.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { Book } from "@/types/book";
+import type { EnvConfigType } from "@/services/environment";
+
+/**
+ * Custom hook to handle batch loading of book configurations
+ * @param libraryBooks - Array of books to load configs for
+ * @param envConfig - Environment configuration
+ * @param settings - System settings
+ */
+export const useBookConfigLoader = (
```
chrox: How about storing the `progress` number for each book directly in `library.json`? This way, we can avoid reading the configuration file for each book. This approach can significantly improve performance, especially when the library contains several hundred books, as it reduces the overhead when opening the application.

chrox (changes requested):

apps/readest-app/src/hooks/useBookConfigLoader.ts
chrox: There’s no need to process previously opened books, we can simply save the progress to the library store whenever the config file for a book is saved. After this feature is released only newly opened books have the progress indicator. I think this is acceptable since most users of readest still have a relatively short reading history and it will be easy to manually reopen each book in order to have a complete progress indicators.

eg9y:
Desktop:
image

Web:
image

chrox:
LGTM

PR #194, opened 2025-01-19
WIP: Users/eg9y/books recursive imports

eg9y:
Add capability to recursively import books.

Flow:
1. On desktop, click plus button at header
2. click Select Directory
3. In the folder selector, Click folder with multiple books, which may have subfolders with books
4. Click select

https://github.com/user-attachments/assets/7e704fdb-a6a4-4d08-bdd1-0dd933aaddee

This is still WIP since it's functional, but very slow. LMK if anyone has any suggestions.


apps/readest-app/src-tauri/src/lib.rs
```diff
@@ -225,3 +248,83 @@ pub fn run() {
...
+        }
+        Err(e) => return Err(format!("Failed to read file metadata: {}", e)),
+    }
+
+    Ok(())
+}
+
+#[command]
+async fn find_book_files(path: String) -> Result<Vec<String>, String> {
+    let book_extensions: Vec<&str> = vec!["epub", "pdf", "mobi", "azw3", "txt"];
```
houqp, line 273: same list has been defined in `readest-app/src/services/constants.ts`, better to consolidate for a single source of truth of supported extensions

eg9y: the issue is that i'm not sure how to set this as a single source between rust and js though

chrox: Can we pass the list from JavaScript to rust as the payload of the invoke?

apps/readest-app/src-tauri/capabilities/default.json
```diff
@@ -23,6 +23,11 @@
       ],
       "deny": []
     },
+    {
+      "identifier": "fs:allow-read",
```
chrox, line 27: This is not preferred and might raise security issues. We can allow specific file like this: https://github.com/readest/readest/blob/main/apps/readest-app/src-tauri/src/lib.rs#L29

apps/readest-app/src/services/webAppService.ts
```diff
@@ -141,25 +141,6 @@ const indexedDBFileSystem: FileSystem = {
   async removeDir() {
     // Directories are virtual in IndexedDB; no-op
   },
-  async readDir(path: string) {
```
chrox, line 144: The web version needs this API to delete books.

chrox (changes requested):

apps/readest-app/src/services/nativeAppService.ts
```diff
@@ -59,8 +59,9 @@ export const nativeFileSystem: FileSystem = {
     return URL.createObjectURL(new Blob([content]));
   },
   async copyFile(srcPath: string, dstPath: string, base: BaseDir) {
-    const { fp, baseDir } = resolvePath(dstPath, base);
-    await copyFile(srcPath, fp, base && { toPathBaseDir: baseDir });
```
chrox, line 63: As I read from the documentation copyFile will avoid file data copying from rust to JavaScript then to rust again. It will just copy the file in rust land, isn't it?

chrox (changes requested):

apps/readest-app/src/services/webAppService.ts
chrox: I double checked this. Currently this API is not used by other functions. But in the future there will be places that utilize readDir in web app for example when book files need pre/post-processing and there are some intermediate caches/documents generated when reading. We need a way to list a directory in order to keep the API consistent across platforms. An optimized implementation is needed of course for the web file system. The indexedDB approach is not performant and [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) might be an option.

apps/readest-app/src/app/library/page.tsx
```diff
@@ -219,6 +220,134 @@ const LibraryPage = () => {
     setIsSelectMode(!isSelectMode);
   };
 
+  const importDirectory = async (dirPath: string) => {
+    if (!appService) return;
+
+    setLoading(true);
+    try {
+      // list files
+      const bookPaths = await invoke<string[]>('find_book_files', { path: dirPath });
```
chrox, line 229: Better to pass `SUPPORTED_FILE_EXTS` here.

apps/readest-app/src/types/system.ts
```diff
@@ -26,6 +26,11 @@ export interface AppService {
   hasWindowBar: boolean;
   isAppDataSandbox: boolean;
 
+  listen?(
```
chrox, line 29: The name “listen” feels very Tauri-specific. Can we consider a different name that is more generic? How about `subscribe`? And even other platforms have no implementation of this function they should have a no-op definition w or w/o a warning log to keep the interface consistent.

chrox (changes requested):

chrox:
@eg9y The general importing flow is fine. But we need to get rid of the full file system file access and allow just necessary files in `validate_book_files` in rust.

@eg9y @houqp We also need to develop tools for performance profiling. Do you have any suggestions?

houqp:
for rust specifically, [criterion](https://github.com/bheisler/criterion.rs) is a pretty good choice.

Issue #166, opened 2025-01-15
Option to delete book isn't clear (website)


matsumurae:
## Bug Description

It took a while to find where the "delete" option was. It's not that obvious, so at first I thought there was no possible to delete them.

## Expected behavior

I have three ideas:

1. place the "delete" (among other options if needed) on the three dot icon of each book, to allow delete individually.
2. Instead of one click to open book, use double click and use the one click to "select". This allows for multiple selection (as Drive does). This ends in two options: add a delete button on the bar (where the select is) disabled and only usable when click, or show the delete button only when click.
3. Place an X to delete the book. Problem is, it will not work on mobile unless you hold or do the idea 2 (use double click to open book).

## Technical inputs

Operating System: any.
Readest Version: website.

chrox:
I think option 1 is more feasible.

PR #204, opened 2025-01-21
Add delete functionality to BookDetailModal with confirmation alert

zangjiucheng:
Handle Issue #166, add delete button to the modal card. 

Match behaviour 
> "place the "delete" (among other options if needed) on the three dot icon of each book, to allow delete individually."

image


chrox:
LGTM

PR #205, opened 2025-01-21
mobile: swipe up to toggle header/footer

houqp:
more nature to toggle header/footer with swipes on mobile

Screenshot 2025-01-20 at 8 53 05 PM

### DIFF
--- a/apps/readest-app/src/app/reader/utils/iframeEventHandlers.ts
+++ b/apps/readest-app/src/app/reader/utils/iframeEventHandlers.ts
@@ -130,3 +130,36 @@ export const handleClick = (bookKey: string, event: MouseEvent) => {
     }
   }, doubleClickThreshold);
 };
+
+const handleTouchEv = (bookKey: string, event: TouchEvent, type: string) => {
+  const touch = event.targetTouches[0];
+  const touches = [];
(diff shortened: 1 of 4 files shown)


apps/readest-app/src/app/reader/utils/iframeEventHandlers.ts
```diff
@@ -130,3 +130,36 @@ export const handleClick = (bookKey: string, event: MouseEvent) => {
...
+  const touches = [];
+  if (touch) {
+    touches.push({
+      clientX: touch.clientX,
+      clientY: touch.clientY,
+      screenX: touch.screenX,
+      screenY: touch.screenY,
+    });
+  }
+  window.postMessage(
```
houqp, line 145: it would be good to figure out a better way to propagate events to make the event handling logic more efficient, right now every handler need to subscribe and filter all messages.

chrox:
Looks great.

chrox:
I also found this gesture conflicts with scrolling in the scrolled mode. We need to distinguish between them.

houqp:
good call, sent a PR to disable it in scrolled mode: https://github.com/readest/readest/pull/224

PR #207, opened 2025-01-21
ci: run rust clippy/fmt

houqp:


chrox:
This is awesome.

chrox:
@houqp Should we delete the Cargo.lock file in the src-tauri directory?

PR #255, opened 2025-01-25
Minor Spelling Check

palashahuja:
Minor spelling check on the contributing guideline. It should be `apps/readest-app` instead of `app/readest-app`


chrox:
Good catch.

Issue #241, opened 2025-01-24
FR: Add Keyboard Shortcuts to Notetaking UI


rougetimelord:
**Does your feature request involve difficulty completing a task? Please describe.**
Having to click the "save" button in the note taking UI interrupts user's flow. Having to click outside of the pane to close it is also disruptive

**Describe the solution you'd like**
A possible solution is to enter a keyboard shortcut to save notes such as control + enter, and a shortcut to close the note pane without saving like escape.

**Describe alternatives you've considered**
Adding a configurable setting for shortcuts

chrox:
Sounds reasonable.

jackplo:
Can you put me on this issue?

chrox:
Sure, no problem.

PR #378, opened 2025-02-15
FR #241: Added Shortcuts to notetaking UI

jackplo:
Added keyboard shortcuts to the note editor component. Users can now use keyboard shortcuts to save and exit the note taking UI.

Saving a note: ctrl + enter
Exiting a note without saving: escape


chrox:
LGTM. As for some code format lines, did you use the prettier to auto-format them as described here https://github.com/readest/readest/blob/main/CONTRIBUTING.md#implement-your-changes. If not we should better to do it otherwise when others modify the files the lines will be reverted back.

jackplo:
Yes I ran the prettier auto formatter using `pnpm build` although nothing was changed so I assumed I follow the code formatting rules. Does what I wrote look correct to you or are you expecting something else?

chrox:
Seems no problem. Merging this now. Thanks.

PR #430, opened 2025-02-22
Add Dockerfile to selfhost

Gigoo25:
Add a Dockerfile to be able to self host the web version of readest.

This can be built locally using:

```bash
git clone https://github.com/readest/readest.git
cd readest
git submodule update --init --recursive
docker build -t readest .
```

And can be run using:

```bash
docker build -t readest .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+FROM node:22-slim
+
+ENV PNPM_HOME="/root/.local/share/pnpm"
+ENV PATH="${PATH}:${PNPM_HOME}"
+
+RUN npm install --global pnpm
+
+# Install necessary packages
+RUN apt update -y && apt install -y --no-install-recommends \
+    libwebkit2gtk-4.1-dev \
+    build-essential \
+    curl \
+    wget \
+    file \
+    libxdo-dev \
+    libssl-dev \
+    libayatana-appindicator3-dev \
+    librsvg2-dev \
+    ca-certificates \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Install Rust and Cargo
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+COPY . /app
+
+WORKDIR /app
+
+RUN pnpm install
+
+RUN pnpm --filter @readest/readest-app setup-pdfjs
+
+WORKDIR /app/apps/readest-app
+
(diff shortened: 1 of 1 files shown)


chrox:
LGTM

maelp:
Is it possible to have a docker-compose setup which includes a local supabase instance so it can be fully self-hosted?

chrox:
I don’t have experience with this and it’s not something I’m currently planning to work on, but a pull request is definitely welcome!

maelp:
@chrox I could help if needed! Can you tell me what Supabase is needed for?
If I self-host the web version, are the books stored in localStorage in the browser, or on disk? What I would need is just a reader hosted on my personal server, with books on disk, and that the web reader would be protected by login, and download books when needed (rather than having everything in localStorage)

chrox:
These tables in supabase are needed for the sync API: https://github.com/readest/readest/wiki/Supabase-Tables-Schema-for-Sync-API

Book files are stored remotely in either S3 or R2 buckets and access information can be set in the .env.local file.

Local book files are stored in IndexedDB in browsers and local disk in Tauri apps.

maelp:
Really cool! So if we use a docker version of Readest, and we don't setup supabase, is it still possible to have ebooks stored on S3? or does that come only with the supabase integration?

chrox:
The sync API is mainly used for cross-device synchronization. Without a central database to store cloud file paths, its usefulness is limited. If you’re using Readest on a single device, there’s no need for the sync API at all.

maelp:
@chrox Thanks! I'm not sure I completely understand though: is there a way for me to just host readest-web on my server with S3-backed bucket for the data, and WITHOUT using sync? (so no supabase?)

in this case how do I upload the books? through the UI? or does drag-and-dropping in the UI force the book in "localStorage" instead of S3?

chrox:
If there is only one device, why the book files need to be uploaded to the cloud. I don't understand the use case.

chrox:
S3 is only used for file storage in Readest, we need a central database to store the list of the uploaded files so that other devices can read the list and download the files.

maelp:
Because if it's the readest-web I can host it somewhere eg readest.myserver.com and access it from my laptop/iphone/etc

maelp:
Here there would be a "single device/service" which would be Readest-web and accessed through it's public URL on my laptop or iphone (through the chrome browser) etc

would that make sense? so that I can access the app anywhere through it's web interface, yet share files

chrox:
> would that make sense? so that I can access the app anywhere through it's web interface, yet share files

Not possible without a database.

PR #432, opened 2025-02-22
Add arabic localization

ahmed-0011:


chrox:
Great. Could you also add `ar` after `vi` [here](https://github.com/readest/readest/blob/main/apps/readest-app/i18next-scanner.config.js#L26) so that it will scan new strings that will be translated in the future? Thanks.

chrox:
Also I might need your help to confirm the RTL layout of the UI is correct when this RTL feature is ready.

ahmed-0011:
Yes, I can help you by double-checking if the RTL support is implemented correctly. Just @me when you want to ask about something.

Also, the following guide is very helpful; if you could take a look at it when you start implementing the RTL support.

https://rtlstyling.com/posts/rtl-styling

chrox:
> https://rtlstyling.com/posts/rtl-styling

This guideline is really helpful. Thanks.

PR #514, opened 2025-03-07
Fixes the scrollbar gutter issue for DaisyUI

xhuajin:
When the Settings are opened, because daisyui has set scrollbar-gullter, the <html> will have an extra virtual scrollbar, causing the page style to be abnormal.

image


chrox:
LGTM

PR #548, opened 2025-03-09
fix: remove seldomly used option to invert colors in dark mode

chrox:
This option is hard to understand and explain. Most hardcoded color in dark mode that affects readability has already been fixed in the reader so there is no necessity to have this option.


pe86ehdl:
"so there is no necessity to have this option"
Screenshot_20250825_034530
Screenshot_20250825_034601
Really?

chrox:
It does have some use cases like this, and we have renamed it to "invert image in dark mode":

image

PR #718, opened 2025-03-26
refactor: support both r2 and s3 storage

XiGou:
since supabase does not support utf-8 file key now
https://github.com/supabase/storage/issues/133

so if we want to support common S3 storage, we can only allow Safe characters in https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

----

and this patch add an abstraction of object storage, then we can support more storage type in the future.


chrox:
Thanks for the patch. There are several issues that need to be addressed:
1. The `@aws-sdk/client-s3` does not work on Cloudflare worker with `opennext`(although it works on vercel edge runtime) so it's better to be refactored with `aws4fetch` which provides a modern and browser/edge friendly API to `s3/r2`. And I planed to delete the `utils/s3.ts` module since we will only use the common APIs of both `s3` and `r2`. So maybe it can be refactored into `utils/object.ts` (sorry that modules in utils should be named with only one word to keep them simple :doge)
2. And there is a tiny styling problem. Indentation should be always with 2 spaces.

apps/readest-app/src/utils/misc.ts
```diff
@@ -7,8 +8,14 @@ export const randomMd5 = () => md5(Math.random().toString());
...
 
 export const makeSafeFilename = (filename: string, replacement = '_') => {
+  var unsafeCharacters
   // Windows restricted characters + control characters and reserved names
-  const unsafeCharacters = /[<>:"\/\\|?*\x00-\x1F]/g;
+  if (getStorageType() === 'r2') {
+    unsafeCharacters = /[<>:"\/\\|?*\x00-\x1F]/g;
+  } else {
+    // S3 storage: https://docs.aws.amazon.com/zh_cn/AmazonS3/latest/userguide/object-keys.html
+    unsafeCharacters = /[^0-9a-zA-Z!\-\/_.*'()]/g;
```
chrox: This is indeed safe but ugly given that the filenames can be revealed in the library with the file manager. How about just setting the book hash + file extension as the object key in s3/r2 and saving with the real name after it is downloaded?

XiGou: that is a goot solution
I'll try it.

apps/readest-app/.env.local.example
```diff
@@ -9,9 +9,18 @@ SUPABASE_ADMIN_KEY=YOUR_SUPABASE_ADMIN_KEY
...
+NEXT_PUBLIC_OBJECT_STORAGE_TYPE=r2
+
 R2_TOKEN_VALUE=YOUR_R2_TOKEN_VALUE
 R2_ACCESS_KEY_ID=YOUR_R2_ACCESS_KEY_ID
 R2_SECRET_ACCESS_KEY=YOUR_R2_SECRET_ACCESS_KEY
 R2_BUCKET_NAME=YOUR_R2_BUCKET_NAME
 R2_ACCOUNT_ID=YOUR_R2_ACCOUNT_ID
 R2_REGION=YOUR_R2_REGION
+
+S3_ENDPOINT=https://uczkovevufgkhrpotpnh.supabase.co/storage/v1/s3
```
chrox: Please use PLACE_HOLDER here.

apps/readest-app/src/pages/api/storage/delete.ts
```diff
@@ -2,7 +2,7 @@ import type { NextApiRequest, NextApiResponse } from 'next';
 import { corsAllMethods, runMiddleware } from '@/utils/cors';
 import { createSupabaseClient } from '@/utils/supabase';
 import { validateUserAndToken } from '@/utils/access';
-import { deleteObject } from '@/utils/r2';
+import { deleteObject } from '@/utils/object_storage';
```
chrox: Please rename `object_storage` to `object` or `storage` or `rs6` or `t6` or whatever reasonable one word module.

apps/readest-app/src/pages/api/storage/upload.ts
```diff
@@ -75,15 +75,14 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
-        objectKey,
+        fileKey,
         objSize,
         1800,
       );
 
       res.status(200).json({
         uploadUrl,
-        fileKey: objectKey,
+        fileKey: fileKey,
```
chrox: Might be just `fileKey` for short.

apps/readest-app/src/services/appService.ts
```diff
@@ -247,34 +246,55 @@ export abstract class BaseAppService implements AppService {
...
+        const content = await this.fs.readFile(lfp, 'Books', 'binary');
+        file = new File([content], cfp);
+      } else {
+        file = await new RemoteFile(this.fs.getURL(`${this.localBooksDir}/${lfp}`), cfp).open();
+      }
+      console.log('Uploading file:', lfp, 'to', cfp);
+      await uploadFile(file, localFullpath, handleProgress, book.hash);
+      uploaded = true;
+      completedFiles.count++;
+    }
```
chrox: Can we reduce the redundant code? Since there might be other intermediate files to download and upload later like the translated file or AI generated contents, we won't like to repeat this for each new file type.

chrox (changes requested):

apps/readest-app/src/services/appService.ts
```diff
@@ -286,52 +306,63 @@ export abstract class BaseAppService implements AppService {
...
-          } else {
-            downloaded = await this.fs.exists(fp, 'Books');
-          }
-        } catch {
-          if (fp === getCoverFilename(book)) {
-            console.log('Failed to download cover image:', fp);
-          } else {
-            throw new Error('Failed to download book file');
-          }
+    var toDownloadFpCount = 0;
```
chrox: `var` is not allowed with our `tsconfig`.

chrox (changes requested):

apps/readest-app/src/services/appService.ts
XiGou: done

apps/readest-app/src/services/appService.ts
XiGou: WIP, when ready for review, I'll let you know

apps/readest-app/src/pages/api/storage/upload.ts
XiGou: done

apps/readest-app/src/pages/api/storage/delete.ts
XiGou: done, renamed to object

apps/readest-app/.env.local.example
XiGou: done

apps/readest-app/src/utils/misc.ts
XiGou: done, filekey changed to bookhash for S3
R2 unchanged, since need a migration process.

apps/readest-app/.env.local.example
```diff
@@ -9,9 +9,18 @@ SUPABASE_ADMIN_KEY=YOUR_SUPABASE_ADMIN_KEY
 DEEPL_PRO_API_KEYS=YOUR_DEEPL_PRO_API_KEYS
 DEEPL_FREE_API_KEYS=YOUR_DEEPL_FREE_API_KEYS
 
+# r2, s3
+NEXT_PUBLIC_OBJECT_STORAGE_TYPE=r2
```
chrox, line 13: I wonder if we can keep the storage type transparent to the frontend so that we can remove the prefix of `NEXT_PUBLIC_`. Probably we should save the fileKey after the file is uploaded and let the API layer decide which storage type it should be so that the frontend won't need to care about the difference between different backend storage types. We can make this a TODO in the future refactoring.

apps/readest-app/src/services/appService.ts
```diff
@@ -218,10 +219,8 @@ export abstract class BaseAppService implements AppService {
   }
 
   async deleteBook(book: Book, includingUploaded = false): Promise<void> {
-    const fps = [getFilename(book), getCoverFilename(book)];
-    const localDeleteFps = (
-      await Promise.all(fps.map(async (fp) => ((await this.fs.exists(fp, 'Books')) ? fp : null)))
-    ).filter(Boolean) as string[];
+    const fps = [getRemoteBookFilename(book), getCoverFilename(book)];
+    const localDeleteFps = [getLocalBookFilename(book), getCoverFilename(book)];
```
chrox, line 222: When deleting we should check the existence of the file otherwise there might be exception thrown.

chrox (changes requested):

apps/readest-app/src/services/appService.ts
```diff
@@ -218,10 +219,8 @@ export abstract class BaseAppService implements AppService {
...
     for (const fp of localDeleteFps) {
       await this.fs.removeFile(fp, 'Books');
```
chrox: Checking existence before removeFile should be fine.

apps/readest-app/.env.local.example
XiGou: nice to have
added TODO comment.
just keep this patch simple now

apps/readest-app/src/services/appService.ts
XiGou: done

apps/readest-app/src/services/appService.ts
XiGou: done

apps/readest-app/src/services/appService.ts
```diff
@@ -243,38 +243,61 @@ export abstract class BaseAppService implements AppService {
...
-      }
-      console.log('Uploading file:', fp);
-      await uploadFile(file, fullpath, handleProgress, book.hash);
+
+    const handleProgress = createProgressHandler(toUploadFpCount, completedFiles, onProgress);
+
+    if (coverExist) {
+      const lfp = getCoverFilename(book);
+      const cfp = `${CLOUD_BOOKS_SUBDIR}/${getCoverFilename(book)}`;
+      await this.uploadFileToCloud( lfp, cfp, handleProgress, book.hash);
```
chrox: extra space here.

chrox (changes requested):

apps/readest-app/src/services/appService.ts
XiGou: hard to identify it, mayeb we should update the linting config sometime.

chrox: Yeah good to have it.

chrox:
Looks great. And I tested locally and didn't notice any problem.

apps/readest-app/src/services/appService.ts
chrox: Btw, it will be handy to install the prettier plugin for vscode to avoid this. See https://github.com/readest/readest/blob/main/CONTRIBUTING.md#implement-your-changes

XiGou:
and I had test R2 and supabase depolyment

didn't notice any problem.

Eyion:
I downloaded Readest_0.9.27_x64-portable, how do I set up R2 storage in cloud sync?

chrox:
It cannot be done in client side because the server API needs to know where the data are stored. And if you self-deploy the server API you can set R2/S3 storage location and access key. Actually the official server API does use the R2 storage for now.

Eyion:
Are there plans to add R2/S3 storage location and access key settings on the client? I signed up for a readest account and found that the free version only has 500Mb of storage capacity. I hope my own r2 storage can work.

chrox:
It's technically possible but the priority is relatively low. We have tons of high priority features to implement. But pull requests are always welcome.

Eyion:
I'd love to help if I can, but unfortunately it's not my area of ​​expertise. Thank you for your hard work and patient responses.

PR #1381, opened 2025-06-11
bump: sync with upstream foliate-js

chrox:


CYJB:
this commit break allowScript options from https://github.com/readest/readest/commit/6ec3a5b33277fe5315259749817cecf70e468432#diff-5b2f0410f43d2bb003f58d2acbaa45e7cb480d857ade10c683257ee14aacc15dL125

is there something wrong?

chrox:
The `allowScript` gate was refactored into a more generic sanitizer defined here: https://github.com/readest/readest/blob/main/apps/readest-app/src/services/transformers/sanitizer.ts which is always enabled.

PR #1384, opened 2025-06-12
add more aggressive overrides for background and text colour

jamessizeland:
Adding more style overrides for Manning epubs that have coloured titles, hopefully this doesn't negatively interact with other books.

no override
image

old override
image

new override
image


chrox:
I think this is pretty safe since overriding color is a very new option and it's disabled by default.

PR #1385, opened 2025-06-12
add vscode settings to include inlayhints

jamessizeland:
This is a suggested edit, no worries if you don't think its appropriate, but it adds ghost text type hints in for typescript, like we're used to in Rust.

image


chrox:
Looks cool.

Issue #1329, opened 2025-06-03
FR: support syntax highlighting in code blocks


jamessizeland:
**Does your feature request involve difficulty for you to complete a task? Please describe.**

> I'd love to have an ebook reader that has syntax highlighting in code blocks, so I can read code in software manuals the way I'd do so in an editor.
>
> i.e. like the nice formatting on https://corrode.dev/blog/prototyping/

![Image](https://github.com/user-attachments/assets/67c6f2c4-0cfb-48a5-b412-c6d9aa69a5a9)

**Describe the solution you'd like**

> Context inference is possible with tools like https://highlightjs.org/ and https://prismjs.com/.  It might be possible to have this be an auto inference across a book (assume the same language, or re-infer each code block and fall back to safe defaults otherwise)  Or could be set in options.

**Describe alternatives you've considered**

> I've not really seen any ebook readers do this (I've tried to find one) and I'm not totally sure why, its relatively standard now in LLM generated code blocks.
>
> I'll be starting to work on this but input/advice would be appreciated.

jamessizeland:
From some further investigating, it looks like there is some syntax highlighting available in Readest already, I'll continue to educate myself about it. https://github.com/readest/readest/issues/1328

chrox:
The code block highlighting is handled through CSS preprocessing, not by Readest itself. You can see a good example of this in O’Reilly ebooks, where code blocks are fully syntax-highlighted.

chrox:
![Image](https://github.com/user-attachments/assets/c62474e2-827b-45bd-b4a5-0401ca900673)

jamessizeland:
Nice, I think there could still be a benefit in trying to add syntax highlighting into existing books that overrides code block style (if it exists) similarly to the general prose themes.

chrox:
`prismjs` seems viable. Would you like to give it a try?

jamessizeland:
Yes I'll be able to take a look next week.

jamessizeland:
Its working! I'll open a draft and we can discuss the way it should be turned on/off and language detection strategies. 

![Image](https://github.com/user-attachments/assets/059c9a77-66d9-4ec2-85dc-a6e0b642bcbf)

![Image](https://github.com/user-attachments/assets/27e7d17f-f478-417b-a0d6-5b1f95e8567c)

PR #1386, opened 2025-06-12
Add syntax highlighting

jamessizeland:
https://github.com/readest/readest/issues/1329

Still need to address 
- [x] language detection.  Autodetection is okay, but not great.
- [x] toggling on/off
- [ ] translations
- [ ] additional themes?
- [ ] side-effects?

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@
 | **Translate with DeepL**                | From a single sentence to the entire book—translate instantly with DeepL.                      | ✅         |
 | **Text-to-Speech (TTS) Support**        | Enjoy smooth, multilingual narration—even within a single book.                                | ✅         |
 | **Library Management**                  | Organize, sort, and manage your entire ebook library.                                          | ✅         |
-| **Code Syntax Highlighting**                | Read software manuals with rich coloring of code examples.                                     | ✅         |
+| **Code Syntax Highlighting**            | Read software manuals with rich coloring of code examples.                                     | ✅         |
 
 ## Planned Features
 


jamessizeland:
Ok ready for your inspection!

output

chrox:
Overall, this looks nice! Just two adjustments we’d like to make:

1. UI consistency: The highlighting toggle and the language selector should be grouped under a “Code Highlighting” section, placed beneath the “Theme Color” section. The language selector should be disabled when code highlighting is turned off—similar to how the “Translation” settings behave.

image

2. Naming convention: The `override` prefix is meant for global flags that apply to an entire settings tab. Since code highlighting isn’t overriding anything, we should rename `overrideCodeHighlighting` and `overrideCodeLanguage` to simply `codeHighlighting` and `codeLanguage`.

apps/readest-app/src/app/reader/components/settings/ColorPanel.tsx
```diff
@@ -52,6 +57,19 @@ const ColorPanel: React.FC<{ bookKey: string }> = ({ bookKey }) => {
...
 
+  useEffect(() => {
+    if (overrideCodeHighlighting === viewSettings.overrideCodeHighlighting) return;
+    saveViewSettings(envConfig, bookKey, 'overrideCodeHighlighting', overrideCodeHighlighting);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [overrideCodeHighlighting]);
+
+  useEffect(() => {
+    if (codeLanguage === viewSettings.overrideCodeLanguage) return;
+    console.log('Setting code language to: ', codeLanguage);
```
chrox: Remove unnecessary console log in production.

apps/readest-app/src/app/reader/components/settings/ColorPanel.tsx
```diff
@@ -154,10 +172,32 @@ const ColorPanel: React.FC<{ bookKey: string }> = ({ bookKey }) => {
...
 
+          <div className='flex items-center justify-between'>
+            <h2 className=''>{_('Override Code Highlighting')}</h2>
+            <input
+              type='checkbox'
+              className='toggle'
+              checked={overrideCodeHighlighting}
+              onChange={() => setOverrideCodeHighlighting(!overrideCodeHighlighting)}
+            />
+            <select
```
chrox: Please reuse the `Select` component in `@/components/Select` for consistent UI.

apps/readest-app/src/app/reader/components/FoliateViewer.tsx
```diff
@@ -54,6 +55,25 @@ const FoliateViewer: React.FC<{
...
+
+    const viewSettings = getViewSettings(bookKey);
+    const docs = viewRef.current.renderer.getContents();
+    if (!viewSettings || !docs) return;
+    // Apply or remove highlighting in all current documents based on the setting
+    for (const { doc } of docs) {
+      manageSyntaxHighlighting(doc, viewSettings, isDarkMode);
+    }
+  }, [
+    getViewSettings(bookKey)?.overrideCodeHighlighting,
```
chrox: To keep the footprint minimal, the enable/disable toggle and language selector callbacks should be handled within the “Color” settings tab. Then, apply the code highlighting logic inside the `docLoadHandler` in `FoliateViewer` as you have already done in this PR, so that when code highlighting is turned off, `manageSyntaxHighlighting` isn’t called at all.

jamessizeland: I'm not sure I'm fully understanding what you're meaning here.  The call in docLoadHandler only triggers on first load and not when a setting is changed, so that's why I added the useEffect callback to trigger on settings change.  Do you mean getting the viewRef in the Color tab to do the change there?

jamessizeland:
image

apps/readest-app/src/app/reader/components/FoliateViewer.tsx
chrox: Yes, you can get the view with `const view = getView(bookKey)` in any hook or component. Settings change can be handled in the useEffect of `codeHighlighting` and `codeLanguage`.

chrox:
Placing the `Code Highlighting` section below the `Theme Color` section helps reduce cognitive load for users who aren’t familiar with coding, making the settings more intuitive and accessible.

apps/readest-app/src/app/reader/components/FoliateViewer.tsx
jamessizeland: awesome, I'll look now

apps/readest-app/src/app/reader/components/settings/ColorPanel.tsx
```diff
@@ -20,6 +20,8 @@ import { useSettingsStore } from '@/store/settingsStore';
 import { useResponsiveSize } from '@/hooks/useResponsiveSize';
 import { saveViewSettings } from '../../utils/viewSettingsHelper';
 import ThemeEditor from './ThemeEditor';
+import { CODE_LANGUAGES, CodeLanguage } from '@/utils/highlightjs';
+import Select from '@/components/Select';
```
chrox, line 23: I’d prefer to place named imports at the top and keep all default imports—for example, `import Select` and `import ThemeEditor` grouped together below, especially for imports from within the project.

apps/readest-app/src/app/reader/components/settings/ColorPanel.tsx
```diff
@@ -67,6 +84,23 @@ const ColorPanel: React.FC<{ bookKey: string }> = ({ bookKey }) => {
...
 
+  // manage syntax highlighting toggle
+  useEffect(() => {
+    const viewSettings = getViewSettings(bookKey);
+    const view = getView(bookKey);
+    if (!viewSettings || !view) return;
+    const docs = view.renderer.getContents();
+    // Apply or remove highlighting in all current documents based on the setting
+    for (const { doc } of docs) {
+      manageSyntaxHighlighting(doc, viewSettings, isDarkMode);
```
chrox: It would be better to apply syntax highlighting directly in the `codeLanguage` and `codeLanguage` useEffect after `saveViewSettings` is called which will update the viewSettings.

chrox (changes requested):

apps/readest-app/src/app/reader/components/settings/ColorPanel.tsx
jamessizeland: it also needs to trigger when isDarkMode toggles too though.

chrox: You can add `isDarkMode` in the dependencies of codeLanguage useEffect as `[codeLanguage, isDarkMode]`.

jamessizeland: Okidoke, yeah just wanted to clarify your preference. I'll combine the two highlight related useEffects and add in the dark mode trigger too 👍.

apps/readest-app/src/utils/highlightjs.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (codeLanguage && codeLanguage !== 'auto-detect') {
+      block.classList.add(`language-${codeLanguage}`);
+    }
+    hljs.highlightElement(block as HTMLElement);
+  });
+};
+
+/** Return either github light or dark theme */
+const getHighlightJsStyles = (isDarkMode: boolean) => {
+  return isDarkMode
```
chrox: Is this possible?
```
/* light mode style */

@media (prefers-color-scheme: dark) {
/* dark mode style */
}
```

jamessizeland: ooh I'll try that yeah

jamessizeland:
happy for me to put a feature line in the Readme table?

```
| **Code Syntax Highlighting**                | Read software manuals with rich coloring of code examples.                                     | ✅         |
```

perhaps also an image?

chrox:
No problem.

jamessizeland:
Amazing! Thanks for being an excellent reviewer!

PR #1444, opened 2025-06-21
fix: correct spelling 'Themeing' to 'Theming' in README

ChuwuYo:
### 1. Fix spelling error in README.md  
- Correct "Themeing Dark Mode" to "Theming Dark Mode"


chrox:
Good catch.

Issue #1453, opened 2025-06-22
Window maximized state leaves gap on right/bottom edges


ChuwuYo:
Description​​:
When Readest window is maximized, a thin gap (unrendered area) appears along the right and bottom edges of the window, exposing the underlying desktop. This behavior differs from standard Windows applications and has been confirmed not to occur in other software or my own Wails-based WebView applications.

I confirm that this happens every time I open it, and I've tried many versions.

​​Reproduction Steps​​:

Launch Readest on Windows 11 22H2 (16:10 resolution 2560×1600)
Maximize the application window
Observe gaps along:
Right edge (visible when moving cursor to screen edge)
Bottom edge (visible above taskbar)
​​Comparison Tests​​:

✅ Other native Windows applications - No gaps
✅ My Wails-based WebView2 application - No gaps
❌ Readest - Gaps present when maximized
​​System Environment​​:

OS: Windows 11 Version 22H2
Display: 16:10 aspect ratio, 2560×1600 resolution
Graphics: NVIDIA RTX 3060 Laptop

​​Video Evidence​​:

https://github.com/user-attachments/assets/dd126b70-8de9-4b9f-ae70-8f4cb07912ac

chrox:
Is it a vertical scroll bar?

ChuwuYo:
> 是垂直滚动条吗？


No, because there are gaps on both the right and bottom sides.

![Image](https://github.com/user-attachments/assets/6141aaff-a033-42bb-a150-c070bcbf1f1d)

ChuwuYo:
![Image](https://github.com/user-attachments/assets/8d2f931c-370a-4417-9138-ac6171272338)

If I use the full-screen display button in the drop-down list, this problem will not occur. The gap only appears when the full screen is enabled by the windowed button in the upper right corner.

chrox:
I suspect this is related to the shadow style (used for the rounded window style) of the window in Windows 11. You might need to disable the style when the window is maximized.

ChuwuYo:
<img width="947" alt="Image" src="https://github.com/user-attachments/assets/e3b0edb7-9bb3-48eb-9ce8-b21f0ee4986b" />


I placed another program in full - screen mode below Readest. As a result, as shown in the figure, the gap at the bottom disappeared, leaving only the one on the right.

ChuwuYo:
![Image](https://github.com/user-attachments/assets/5c716052-fe5e-455a-815b-701ca91d5099)

[...]

PR #1455, opened 2025-06-23
Fix: Standardize error handling in books sync operations

ChuwuYo:
## Problem
Error handling was inconsistent in useBooksSync:
- `processOldBook` downloadBook call had no error handling
- Old books `Promise.all` had no error handling  
- Batch processing `Promise.all` had no error handling
- Only new book downloads had proper error handling

## Changes
- Added missing try-catch blocks to match existing error handling patterns
- Standardized error logging format across all operations
- Added fallback mechanisms for failed batch operations

## Impact
Prevents single operation failures from breaking the entire sync process.

### DIFF
--- a/apps/readest-app/src/app/library/hooks/useBooksSync.ts
+++ b/apps/readest-app/src/app/library/hooks/useBooksSync.ts
@@ -71,7 +71,11 @@ export const useBooksSync = ({ onSyncStart, onSyncEnd }: UseBooksSyncProps) => {
       const matchingBook = syncedBooks.find((newBook) => newBook.hash === oldBook.hash);
       if (matchingBook) {
         if (!matchingBook.deletedAt && matchingBook.uploadedAt && !oldBook.downloadedAt) {
-          await appService?.downloadBook(oldBook, true);
+          try {
+            await appService?.downloadBook(oldBook, true);
+          } catch (error) {
+            console.error('Failed to download old book:', oldBook.title || oldBook.hash, error);
+          }
         }
         const mergedBook =
           matchingBook.updatedAt > oldBook.updatedAt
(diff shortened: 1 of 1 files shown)


apps/readest-app/src/app/library/hooks/useBooksSync.ts
```diff
@@ -82,27 +86,46 @@ export const useBooksSync = ({ onSyncStart, onSyncEnd }: UseBooksSyncProps) => {
...
       if (!updatedLibrary.some((oldBook) => oldBook.hash === newBook.hash)) {
         if (newBook.uploadedAt && !newBook.deletedAt) {
           try {
             await appService?.downloadBook(newBook, true);
-          } catch {
-            console.error('Failed to download book:', newBook);
+          } catch (error) {
+            console.error('Failed to download book:', newBook.title || newBook.hash, error);
           } finally {
-            newBook.coverImageUrl = await appService?.generateCoverImageUrl(newBook);
```
chrox, line 94: `generateCoverImageUrl` is only string manipulation and should never throw exception.

chrox (changes requested):

chrox:
Is there any problem preventing the sync books from working? We might need to solve the problem instead of catching and throwing away the error.

ChuwuYo:
> Is there any problem preventing the sync books from working? We might need to solve the problem instead of catching and throwing away the error.

Sometimes when downloadBook fails, the book is still added to the library in the finally block

chrox:
Note that `useBooksSync` only downloads the book cover and never downloads the epub/pdf file and if there is any error when downloading it's already caught in `appService.downloadBook`, thus it won't raise error any more.

chrox:
We might need steps to reproduce the issue before we can fix it.

chrox:
> > Is there any problem preventing the sync books from working? We might need to solve the problem instead of catching and throwing away the error.
> 
> Sometimes when downloadBook fails, the book is still added to the library in the finally block

That means the library will have books without book cover and the book cover exists in the cloud storage. This is unexpected behavior and I never saw this.

ChuwuYo:
> 我们可能需要采取一些步骤来重现该问题，然后才能修复它。

Sorry.
I synced the books just as I was about to leave, after which the device went offline. When I reconnected to the network and restarted Readest, the problem was resolved.

ChuwuYo:
> > > Is there any problem preventing the sync books from working? We might need to solve the problem instead of catching and throwing away the error.
> > 
> > 
> > Sometimes when downloadBook fails, the book is still added to the library in the finally block
> 
> That means the library will have books without book cover and the book cover exists in the cloud storage. This is unexpected behavior and I never saw this.


I'm sorry to have wasted your time.

chrox:
Never mind. Thank you for your contribution.

ChuwuYo:
> Never mind. Thank you for your contribution.

But this problem still exists. #1453

chrox:
Please discuss in the specific issue ticket.

PR #1546, opened 2025-07-07
feat: support delete the cloud storage of a book

bhscer:
Support user to delete a book's cloud storage without delete its local version.

I also try to fix the BookDetailModal will still show upload/download button when the book is uploaded.

Since the 'cloud delete' icon only have a outline version, I also replace other three button's icon to the outlined version.


apps/readest-app/src/components/BookDetailModal.tsx
```diff
@@ -132,17 +147,22 @@ const BookDetailModal = ({
...
+                      <MdOutlineCloudOff className='fill-red-500' />
+                    </button>
+                  )}
+                  {book.uploadedAt && !book.downloadedAt && handleBookDownload && (
                     <button onClick={handleRedownload}>
-                      <MdCloudDownload className='fill-base-content' />
+                      <MdOutlineCloudDownload className='fill-base-content' />
                     </button>
                   )}
-                  {book.downloadedAt && handleBookUpload && (
```
chrox, line 143: It's intended for re-uploading files.

apps/readest-app/src/components/BookDetailModal.tsx
```diff
@@ -132,17 +147,22 @@ const BookDetailModal = ({
                 <div className='flex flex-wrap items-center gap-x-4'>
                   {handleBookDelete && (
                     <button onClick={handleDelete}>
-                      <MdDelete className='fill-red-500' />
+                      <MdOutlineDelete className='fill-red-500' />
                     </button>
                   )}
-                  {book.uploadedAt && handleBookDownload && (
```
chrox, line 138: This is intended.

apps/readest-app/src/services/appService.ts
```diff
@@ -298,6 +298,20 @@ export abstract class BaseAppService implements AppService {
     }
   }
 
+  async deleteBookCloudStore(book: Book): Promise<void> {
```
chrox: How about adding a param for deleteBook like this?
```
async deleteBook(book: Book, includingUploaded = false, includingLocal = true)
```

chrox:
> I also try to fix the BookDetailModal will still show upload/download button when the book is uploaded.

In rare cases, the local book upload or download status may become out of sync with the remote sync API. Keeping the download and upload buttons always visible allows you to manually re-sync the data.

bhscer:
OK now I have merged two functions into one.

apps/readest-app/src/services/appService.ts
```diff
@@ -291,8 +293,17 @@ export abstract class BaseAppService implements AppService {
...
     }
-    book.deletedAt = Date.now();
-    book.downloadedAt = null;
+
+    if (
+      (!book.uploadedAt && includingLocal) // file only in local
+      || (!book.downloadedAt && !!book.uploadedAt && includingUploaded) // file only in cloud
+      || (!!book.uploadedAt && includingLocal && includingUploaded) // file both in local and cloud
+    ) {
+      book.deletedAt = Date.now();
```
chrox: `deletedAt` is only for the local status. We didn't track the deletedAt of the remote file locally. So
```
if (includingLocal) {
      book.deletedAt = Date.now();
      book.downloadedAt = null;
}
```
should be OK.

chrox (changes requested):

chrox:
LGTM. Thanks.

PR #1606, opened 2025-07-18
Fix dev-web command

vasi:
The readme says to run `pnpm dev-web`, but that wasn't hooked up to anything. A simple script makes it work.


chrox:
Thanks for this.

PR #1652, opened 2025-07-21
feat: Added Yandex Translator

Columpio:
Added [Yandex Translator](https://translate.yandex.ru/) to a list of translators.

Yandex is better than existing translators (Google, Azure, DeepL) for Slavic languages.

I used [this Yandex Translator API](https://translate.toil.cc/v2/docs) as [native Yandex API](https://yandex.cloud/ru/docs/translate/) requires API key (and thus paying money).

I manually tested it and it works well.

`translate.toil.cc` provides 4 Yandex Translate variants: I have chosen the best one by default.


chrox:
LGTM

PR #1770, opened 2025-08-09
feat: Implement KOReader Progress Synchronization

zeedif:
This pull request introduces the initial implementation for synchronizing reading progress with KOReader-compatible sync servers. The goal is to allow users to maintain a reasonably consistent reading position between Readest and KOReader devices.

#### The Compatibility Challenge and Current Approach

A significant technical challenge lies in the fundamental difference between how Readest and KOReader track progress:
- **Readest (foliate-js):** Uses `epubcfi` for reflowable formats (EPUB) and page numbers for paginated formats (PDF).
- **KOReader:** Uses a custom `XPointer` for reflowable formats and a simple page number for paginated formats.

Direct translation between KOReader's XPointer and `epubcfi` is not feasible. Therefore, this initial implementation uses the `percentage` field as a common ground for synchronization.

However, it's important to note that **even the percentage is an approximation**. Due to differences in rendering engines, font rendering, and page layout calculations, the same percentage value in both applications may point to slightly different locations in the text.

#### Implementation Details & Known Limitations

-   **Reflowable Formats (EPUB, MOBI):**
    -   The `progress` field from KOReader (which is a custom XPointer) is checked for `epubcfi` compatibility. **As it is incompatible, the logic falls back to using the `percentage` value for navigation via `view.goToFraction()`.** This provides a "best-effort" approximation of the reading position.

#### Key Features Added:

*   **KOReader Sync Settings UI:** A new modal for configuring the sync server, credentials, and sync strategy.
*   **Conflict Resolution Dialog:** A user-friendly dialog appears when local and remote progress differ, showing contextual previews (e.g., "Page 15 of 300" or "Chapter 3") to help the user decide.
*   **Secure API Proxy:** A Next.js API route (`/api/kosync`) securely proxies requests to the user's sync server.
*   **Automatic & Manual Sync:** Progress is automatically managed, with manual triggers also available.

#### Next Steps

This PR represents a foundational first step. Further work is required to improve the accuracy of the compatibility layer. Long-term, exploring more advanced mapping techniques between progress formats would be beneficial.


zeedif:
Currently, I'm trying to fix an issue where view.goTo in paginated files doesn't change the page (it stays on the current page). I'm aware that this implementation is not yet fully compatible with KOReader for reflowable files due to the difference in the progress format, and that even the percentage is just an approximation.

In the future, I'd like to find a way to align this with a more robust compatibility layer, but I think that will involve a significant amount of work. For now, I'm providing this initial implementation as a functional baseline that we can iterate on.

chrox:
I’m currently working on the epubcfi ↔ xpointer conversion, so we can put this PR on hold until that’s finished. Overall, the implementation looks great, and I don’t expect it will take long to review.

chrox:
Now you can use `xcfi` to convert between epubcfi and xpointer. See https://github.com/readest/readest/pull/1774.

zeedif:
I've done a couple of tests for EPUBs, and the CFI/XPointer conversion works well — it now navigates to the correct page.
However, I'm still running into the issue I mentioned earlier with paginated files (PDFs, CBZs). The call to `view.goTo(pageIndex)` with the progress from `remote.progress` doesn't seem to have any effect. There are no errors in the console; it just silently fails to navigate away from the current page.

zeedif:
I've implemented the change to use view.select(pageIndex) for paginated files, and it seems to be working correctly now—it navigates to the specified page as expected. But I'd appreciate some clarification to ensure I'm on the right track.

My intuition was that `goTo()` would be the general method for navigation (whether by page index or CFI), while `select()` would be for programmatically selecting text on a page, and `deselect()` would clear that selection.

Could you please confirm if using `view.select(pageIndex)` is indeed the correct and intended method for navigating to a specific page in paginated documents? I just want to make sure this fix is robust and won't break with future updates to the view.

chrox:
> I've implemented the change to use view.select(pageIndex) for paginated files, and it seems to be working correctly now—it navigates to the specified page as expected. But I'd appreciate some clarification to ensure I'm on the right track.
> 
> My intuition was that `goTo()` would be the general method for navigation (whether by page index or CFI), while `select()` would be for programmatically selecting text on a page, and `deselect()` would clear that selection.
> 
> Could you please confirm if using `view.select(pageIndex)` is indeed the correct and intended method for navigating to a specific page in paginated documents? I just want to make sure this fix is robust and won't break with future updates to the view.

It should work with goTo(), but there’s probably a bug causing the issue in this case. Let’s stick with select() for now, and we’ll revisit it once we’ve sorted out the fixed-layout rendering stuff.

apps/readest-app/src/pages/api/kosync.ts
```diff
@@ -0,0 +1,49 @@
...
+        'Content-Type': 'application/json',
+      },
+      body: clientBody ? JSON.stringify(clientBody) : null,
+    });
+
+    const data = await response.text();
+    res.status(response.status);
+    try {
+      res.json(JSON.parse(data));
+    } catch (e) {
```
chrox: ./src/pages/api/kosync.ts
41:14  Error: 'e' is defined but never used.  @typescript-eslint/no-unused-vars

chrox (changes requested):

apps/readest-app/src/pages/api/kosync.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  if (!serverUrl || !endpoint) {
+    return res.status(400).json({ error: 'serverUrl and endpoint are required' });
+  }
+
+  const targetUrl = `${serverUrl.replace(/\/$/, '')}${endpoint}`;
+
+  try {
+    const response = await fetch(targetUrl, {
```
chrox: If the `application/vnd.koreader.v1+json` response header is consistently used across all supported koreader-sync implementations, I’d prefer adding this check to prevent potential misuse of the proxy API. Alternatively, we could validate the user authentication token, though that should be considered a last resort.
```
if (response.headers.get('content-type') !== 'application/vnd.koreader.v1+json') {
  throw new Error('Unexpected content type');
}
```

chrox (changes requested):

apps/readest-app/src/app/reader/components/ReaderContent.tsx
```diff
@@ -44,9 +46,22 @@ const ReaderContent: React.FC<{ ids?: string; settings: SystemSettings }> = ({ i
...
+    setActiveBookKey(sideBarBookKey || (bookKeys.length > 0 ? bookKeys[0]! : null));
+  }, [sideBarBookKey, bookKeys]);
+
+  const {
+    syncState,
+    conflictDetails,
+    resolveConflictWithLocal,
+    resolveConflictWithRemote,
+    pushProgress,
+  } = useKOSync(activeBookKey || '');
```
chrox: It's better to use the `useKOSync` hook in each `FoliateViewer` just as `useProgressSync` hook does. In this way multiple documents (in Parallel Read) could be synced in parallel.

chrox (changes requested):

apps/readest-app/src/app/reader/components/ReaderContent.tsx
```diff
@@ -93,20 +112,20 @@ const ReaderContent: React.FC<{ ids?: string; settings: SystemSettings }> = ({ i
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [bookKeys]);
+  }, [bookKeys, pushProgress]);
 
   const saveBookConfig = async (bookKey: string) => {
     const config = getConfig(bookKey);
     const { book } = getBookData(bookKey) || {};
     const { isPrimary } = getViewState(bookKey) || {};
     if (isPrimary && book && config) {
-      eventDispatcher.dispatch('sync-book-progress', { bookKey });
```
chrox, line 109: The `useProgressSync` hook needs this to sync progress when the book is closed.

apps/readest-app/src/app/reader/components/ReaderContent.tsx
```diff
@@ -132,7 +151,8 @@ const ReaderContent: React.FC<{ ids?: string; settings: SystemSettings }> = ({ i
 
   const handleCloseBooks = throttle(async () => {
     const settings = useSettingsStore.getState().settings;
-    await Promise.all(bookKeys.map((key) => saveConfigAndCloseBook(key)));
+    (pushProgress as any).flush?.();
```
chrox: `any` type is not allowed in current lint configurations.

chrox: Is there a global switch for kosync so that when turned off or not login it won't try to sync at all?

apps/readest-app/src/pages/api/kosync.ts
zeedif: I implemented the check for `Content-Type: application/vnd.koreader.v1+json` as you suggested, but it led to an "Unexpected Content-Type" error during my tests.

I dug into the `koreader-sync-server` source code to understand why, and it appears the server's behavior is slightly different from what we assumed based on the `Accept` header.

Here's what I found:

1.  The controllers in `app/controllers/1/syncs_controller.lua` return standard Lua tables (e.g., `return 200, { state = 'OK' }`).
2.  The Gin framework then automatically serializes these tables into a JSON response.
3.  When doing so, Gin sets the standard `Content-Type: application/json` header, not the custom `application/vnd.koreader.v1+json` type.

The server correctly processes the client's `Accept` header for API versioning but responds with the standard JSON MIME type. To align with the server's actual behavior while still achieving the security goal, I've adjusted the check to validate for `application/json` instead. This still effectively prevents the proxy from being used to request non-JSON resources (like HTML, images, etc.):

```typescript
const contentType = response.headers.get('content-type');
if (!contentType || !contentType.startsWith('application/json')) {
  throw new Error('Invalid sync server response: Unexpected Content-Type.');
}
```

This should resolve the issue while keeping the essential security check in place. Let me know what you think

apps/readest-app/src/app/reader/components/ReaderContent.tsx
```diff
@@ -93,6 +96,7 @@ const ReaderContent: React.FC<{ ids?: string; settings: SystemSettings }> = ({ i
     eventDispatcher.on('beforereload', handleCloseBooks);
     eventDispatcher.on('quit-app', handleCloseBooks);
     return () => {
+      handleCloseBooks();
```
chrox: `bookKeys` might be recreated when creating a new parallel read view or closing one. The other books should not be closed in this case.

Issue #1768, opened 2025-08-08
FR: Option to delete local copy of book


arutonee1:
**Does your feature request involve difficulty for you to complete a task? Please describe.**

> Books can take up a lot of space. If #356 is implemented, or if the user is using a paid version, it is very much possible to run out of space on one's device.

**Describe the solution you'd like**

> There already exists an option to remove a book from both local storage and the cloud (deleting the book), as well as an option to remove the book from the cloud only (stop syncing). Logically, there should be an option to remove the book from local storage (resetting the book to its state before being downloaded).

chrox:
In general, this shouldn’t be an issue for most users since storage capacities tend to grow and cost per gigabyte drops significantly every year or two. That said, it's still a good to have option.

PR #1773, opened 2025-08-09
feat: added option to delete only the local copy of a book

arutonee1:
This attempts to implement #1768. I am unsure of how to properly test this functionality, so if tests are needed please point me in the right direction and I'll try my best to implement them ^w^

### DIFF
--- a/apps/readest-app/src/app/library/page.tsx
+++ b/apps/readest-app/src/app/library/page.tsx
@@ -8,7 +8,7 @@ import { OverlayScrollbarsComponent, OverlayScrollbarsComponentRef } from 'overl
 import 'overlayscrollbars/overlayscrollbars.css';
 
 import { Book } from '@/types/book';
-import { AppService, StorageType } from '@/types/system';
+import { AppService, DeleteAction } from '@/types/system';
 import { navigateToLogin, navigateToReader } from '@/utils/nav';
 import {
   formatAuthors,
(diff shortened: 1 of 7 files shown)


arutonee1:
I *have* manually tested it, but of course that's not as rigorous as I'd like :sweat_smile:

chrox:
We should add a parameter for `handleBookDelete` so that it can delete 'local', 'remote' or 'both', and remove the other two variants.

arutonee1:
Sorry, please disregard that commit.

arutonee1:
@chrox I tried a few layouts, and a HOF felt like the easiest option. You could alternatively use a `(b) => deleteBooks(b, '...')`, or just pass deleteBooks and update the signatures, but that gives diminishing returns. Thoughts?

apps/readest-app/src/app/library/page.tsx
```diff
@@ -543,53 +543,40 @@ const LibraryPageContent = ({ searchParams }: { searchParams: ReadonlyURLSearchP
...
-        type: 'error',
-        message: _('Failed to delete cloud backup of the book', {
-          title: book.title,
-        }),
-      });
-      return false;
-    }
+  const handleBookDelete = (storageType: StorageType) => {
+    return async (book: Book) => {
+      const deletionMessages = {
```
chrox, line 548: These string including the following deletionFailMessages  need to be translated by the `_` shorthand of translation:
```
      const deletionMessages = {
        'Both': _('Book deleted: {{title}}'),
        'Cloud': _('Deleted cloud backup of the book: {{title}}'),
        'Local': _('Deleted local copy of the book: {{title}}'),
      };
```

So that `i18next` could find the string and generate translation keys for each string.

chrox (changes requested):

chrox:
Having three separate buttons for deleting a book can feel overwhelming. It would be better to have a single delete button that opens a popup dialog or reuse the confirmation popup with all three options. This approach would be more intuitive and provide a better user experience.

chrox:
> @chrox I tried a few layouts, and a HOF felt like the easiest option. You could alternatively use a `(b) => deleteBooks(b, '...')`, or just pass deleteBooks and update the signatures, but that gives diminishing returns. Thoughts?

I'm good with the HOF.

apps/readest-app/src/app/library/page.tsx
arutonee1: The translation shorthand is used later down the line, at `eventDispatcher.dispatch`, so all strings should be translated. Would it be preferrable to translate in the messages object itself?

chrox: Only the literal strings inside the _ shorthand will be scanned into the translation files as translation keys.

arutonee1: Gotcha- making a commit now

As a side note, after putting the messages together, the cloud deletion fail string is inconsistent with the rest of the messages, as it doesn't mention the title. I don't want to fix all of the translations as well (although it should theoretically be pretty easy with a `:grep` and `:cdo`? Some manual work will be required though), and it really should be in a different PR. Just thought I should mention it.

arutonee1:
> Having three separate buttons for deleting a book can feel overwhelming. It would be better to have a single delete button that opens a popup dialog or reuse the confirmation popup with all three options. This approach would be more intuitive and provide a better user experience.

I'm personally okay with having three buttons, and the confirmation messages make it relatively easy to figure out what the buttons do. Having a dropdown could also work very well, but a new component will have to be made for multi-option confirmation popups. In the end, both of them would be good solutions ^w^

apps/readest-app/src/app/library/page.tsx
chrox: > As a side note, after putting the messages together, the cloud deletion fail string is inconsistent with the rest of the messages, as it doesn't mention the title. I don't want to fix all of the translations as well (although it should theoretically be pretty easy with a `:grep` and `:cdo`? Some manual work will be required though), and it really should be in a different PR. Just thought I should mention it.

Thanks. I will fix it later.

PR #1853, opened 2025-08-20
Change audio object to be continuous between sentences

shaqtiktok:
The audio object currently is being recreated on each sentence. This stops that and allows the playback to be continuous rather than halting.


apps/readest-app/src/services/tts/EdgeTTSClient.ts
```diff
@@ -135,54 +121,53 @@ export class EdgeTTSClient implements TTSClient {
...
           audio.play().catch((err) => {
             cleanUp();
             console.error('Failed to play audio:', err);
             resolve({ code: 'error', message: 'Playback failed: ' + err.message });
           });
         });
+
         yield result;
       } catch (error) {
-        if (error instanceof Error && error.message === 'No audio data received.') {
```
chrox, line 168: This is needed to skip some marks where TTS just cannot generate audio data, e.g. some punctuations.

apps/readest-app/src/services/tts/EdgeTTSClient.ts
```diff
@@ -135,54 +121,53 @@ export class EdgeTTSClient implements TTSClient {
...
-            URL.revokeObjectURL(url);
+            if (audio.src?.startsWith('blob:')) {
+              URL.revokeObjectURL(audio.src);
+            }
+            audio.removeAttribute('src'); // Prevent loading stale blobs
           };
+
           audio.onended = () => {
             cleanUp();
-            if (signal.aborted) {
```
chrox, line 143: Is there a good reason to remove this?

apps/readest-app/src/services/tts/EdgeTTSClient.ts
```diff
@@ -63,64 +63,50 @@ export class EdgeTTSClient implements TTSClient {
     const { marks } = parseSSMLMarks(ssml, this.#primaryLang);
 
     if (preload) {
-      // preload the first 2 marks immediately and the rest in the background
-      const maxImmediate = 2;
-      for (let i = 0; i < Math.min(maxImmediate, marks.length); i++) {
-        const mark = marks[i]!;
+      for (const mark of marks) {
```
chrox: The cold start of TTS can take longer, especially if the first paragraph is lengthy, since it waits for the entire paragraph to preload.

PR #1855, opened 2025-08-20
fix(translator): refresh translation when provider is changed in translator popup

lijiahao66666:
- 在 TranslatorPopup 组件中添加 translate 函数到依赖数组
- 此修改确保在 translate 函数更新时，能够重新获取翻译内容


chrox:
Thanks for the patch.

PR #2269, opened 2025-10-19
fix(css): support at rules of CSS and fix CSS formatter

CreeperIsASpy:
The previous CSS validation was implemented manually using regular expressions and string splitting. This approach was fragile, difficult to maintain, and could not handle modern CSS syntax such as at-rules (@media, @keyframes) or complex selectors.

This PR replaces the manual implementation with `postcss`, a robust and industry-standard CSS parser. If PostCSS can successfully parse the CSS string, it is considered valid; otherwise, it throws an error, which is caught and reported.

This change provides several key benefits:
- Greatly improves the reliability and accuracy of CSS validation.
- Supports a much wider range of modern CSS syntax.
- Simplifies the codebase and improves maintainability.
- Provides more specific and helpful error messages from PostCSS.

> [!IMPORTANT]
> I've added the postcss dependency to package.json, but I'm unable to update the lock file from my current environment. The lock file (package-lock.json or yarn.lock) will need to be regenerated.

### DIFF
--- a/apps/readest-app/package.json
+++ b/apps/readest-app/package.json
@@ -71,7 +71,6 @@
     "aws4fetch": "^1.0.20",
     "clsx": "^2.1.1",
     "cors": "^2.8.5",
-    "cssbeautify": "^0.3.1",
     "dayjs": "^1.11.13",
     "foliate-js": "workspace:*",
     "franc-min": "^6.2.0",
--- a/apps/readest-app/src/components/settings/MiscPanel.tsx
+++ b/apps/readest-app/src/components/settings/MiscPanel.tsx
@@ -1,5 +1,4 @@
 import clsx from 'clsx';
-import cssbeautify from 'cssbeautify';
 import React, { useEffect, useRef, useState } from 'react';
 
 import { useEnv } from '@/context/EnvContext';
@@ -9,7 +8,7 @@ import { useTranslation } from '@/hooks/useTranslation';
 import { useResetViewSettings } from '@/hooks/useResetSettings';
 import { SettingsPanelPanelProp } from './SettingsDialog';
 import { getStyles } from '@/utils/style';
-import cssValidate from '@/utils/css';
+import { validateCSS, formatCSS } from '@/utils/css';
 
 type CSSType = 'book' | 'reader';
 
@@ -47,11 +46,11 @@ const MiscPanel: React.FC<SettingsPanelPanelProp> = ({ bookKey, onRegisterReset 
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const validateCSS = (cssInput: string): { isValid: boolean; error?: string } => {
+  const handleValidateCSS = (cssInput: string): { isValid: boolean; error?: string } => {
     if (!cssInput.trim()) return { isValid: true };
 
     try {
-      const { isValid, error } = cssValidate(cssInput);
+      const { isValid, error } = validateCSS(cssInput);
       if (!isValid) {
         return { isValid: false, error: error || 'Invalid CSS' };
       }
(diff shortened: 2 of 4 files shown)


chrox:
Thanks for the patch! You may need to run pnpm install postcss to properly update both the lockfile and package.json, rather than editing package.json manually. Also, postcss should be listed under dependencies, not devDependencies.

Neclyon:
Now it should work properly. My apologize for using AI!

chrox:
`postcss` is still in `devDependencies` and not in the lockfile thus it will break the CI.

chrox:
Please also check the output of the CI.

chrox:
You will need to actually run the `pnpm install postcss` in you repo and `git add` the lock file and package file that are updated. And ideally you have to lint your code in your local environment and make sure there is no apparent problems before sending a pull request. We won't encourage sending purely AI generated and unverified code to this project. See the [contributing guidelines](https://github.com/readest/readest/blob/main/CONTRIBUTING.md).

chrox:
Another point that still needs justification is why this change is necessary. Are there actual cases where the current CSS validator fails? Keep in mind that the existing validator is only used for user-input custom CSS in ebook layouts, which typically consists of simple, CSS 2-era style snippets. Including the PostCSS library in production would add several hundred kilobytes to the bundle, which doesn’t justify the minimal benefit it would bring over the existing lightweight CSS parser. Moreover, adding another library to the production runtime would reduce, rather than improve, maintainability.

Neclyon:
> Another point that still needs justification is why this change is necessary. Are there actual cases where the current CSS validator fails? Keep in mind that the existing validator is only used for user-input custom CSS in ebook layouts, which typically consists of simple, CSS 2-era style snippets. Including the PostCSS library in production would add several hundred kilobytes to the bundle, which doesn’t justify the minimal benefit it would bring over the existing lightweight CSS parser. Moreover, adding another library to the production runtime would reduce, rather than improve, maintainability.

绝对有，我自己尝试做过一次 CSS，给阅读界面添加一个固定的背景图片，但是却出现各种各样的问题。其中包括图片的尺寸无法自适应、图片会在阅读器层面和应用背景层面各显示一张等等。因此，我尝试使用 @media 来解决这个问题，却发现任何嵌套两个花括号的行为都会导致解析失败。

就算没有这样的问题，直接正则表达式验证 css 的有效性也是非常落后的办法，因为这需要开发者时时刻刻和 CSS 的最新特性保持跟进支持。既然可以用库，为什么还要直接造轮子呢？

___

The motivation for this change came from a practical use case where the current validator falls short. I was trying to add a custom background image to the reader view and wanted to use a @media query to prevent it from scaling improperly on different screen sizes.

However, the current validator fails to parse any CSS that contains nested curly braces, which means modern and essential features like @media, @supports, or @keyframes are unusable. This limits users from applying simple responsive design techniques to improve their reading experience.

Regarding maintainability, while adding a new dependency does introduce management overhead, relying on a dedicated and battle-tested library for a complex task like CSS parsing is generally more sustainable than maintaining a custom regex-based solution. The custom validator would require ongoing effort to keep up with the evolving CSS standard, whereas a library handles this transparently.

That said, your point about the bundle size is completely valid. Including the full postcss library on the client-side is a significant cost. Perhaps we can find a compromise?

Neclyon:
有个小问题……英语不是我的母语，我交的这个 PR …… 也不是很专业。我本身其实不懂这方面开发，您看我的 profile，没有一个是有关 js 系列的，所以我的操作流程 **极其不专业**，会占用您很多的时间；我的所有英语回答……也都是 AI 机翻的，意思难免不准确。
**请您谅解！**

chrox:
No worries at all! It’s perfectly fine.

chrox:
Since our css validator is just a YES or NO hint for users' CSS input it does not need to be full fledged parsing the semantic of CSS. We can enhance it to support media queries in CSS.

chrox:
The new CSS validator now supports nested at-rules. Please let us know if it rejects any valid CSS syntax; note that it is intended to ignore semantic errors.

PR #2376, opened 2025-10-31
feat(ui): Image Gallery Support

manavchaudhary1:
Created new file to handle image zooming features.
Tested on web browser
Was trying to build apk to test on android but taking too much time to build.
So i'll left it for any maintainer to check

And sorry for rearranging order of imports , can anyone revert it back ? it was done by my agent while reordering and removing debug stmt
I would have done it but have to make file for my project submission is on tuesday.
😭

### DIFF
--- a/apps/readest-app/src/app/reader/components/FoliateViewer.tsx
+++ b/apps/readest-app/src/app/reader/components/FoliateViewer.tsx
@@ -1,25 +1,32 @@
-import React, { useEffect, useRef, useState } from 'react';
-import { BookDoc, getDirection } from '@/libs/document';
-import { BookConfig } from '@/types/book';
-import { FoliateView, wrappedFoliateView } from '@/types/view';
-import { Insets } from '@/types/misc';
+import Spinner from '@/components/Spinner';
 import { useEnv } from '@/context/EnvContext';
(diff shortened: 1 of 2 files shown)


apps/readest-app/src/app/reader/components/FoliateViewer.tsx
```diff
@@ -300,12 +429,6 @@ const FoliateViewer: React.FC<{
 
       const { book } = view;
 
-      book.transformTarget?.addEventListener('load', (event: Event) => {
```
chrox: This should not be removed.

apps/readest-app/src/app/reader/components/FoliateViewer.tsx
```diff
@@ -119,6 +122,76 @@ const FoliateViewer: React.FC<{
     );
   };
 
+  const makeImageOverlayScript = (bookKey: string) => {
```
chrox, line 125: There’s already a utility function in the foliate-js library that handles this, and we prefer to reuse existing functionality from foliate-js whenever possible.

chrox:
Image gallery mode is already implemented in #3328.

PR #2430, opened 2025-11-10
fix(annotator): enhance PDF context menu for translation and improve touch handling

ByteFlowing1337:
Now the READEST can translate/wiki/underline/note/search by double-clicking words or selecting the sentences/paragraphs using the mouse cursor.Seriously, why doesn’t the reader support PDF annotations? Most of my ebooks are PDFs, and I can’t interact with the text at all — I’m trying to fix that to make reading PDFs better.


chrox:
Thanks you and this looks good to me. We just need to make sure it works on all platforms so we need a little bit of time to test it.

chrox:
And ideally we should disable the functions that do not work for PDFs.

Issue #2149, opened 2025-10-01
Add Reader Mode for PDFs


achen82:
**Does your feature request involve difficulty for you to complete a task? Please describe.**

Reading PDFs in their default layout is difficult because the text is locked into fixed pages with columns, headers, and other formatting. On smaller screens this requires a lot of zooming and scrolling, which makes it harder to read smoothly like an ebook. It would be great to have something similar to Speechify or ElevenReader app where it extracts the text and formats it like a book. With the current implementation we need to zoom in quite a bit and TTS doesn't work. 

**Describe the solution you'd like**

I’d like a Reader Mode option, similar to ElevenReader or Speechify. This would extract just the text (and optionally inline images) from the PDF and reformat it into a clean, continuous flow — more like an ebook or article view.

This mode would make PDFs much more comfortable to read, especially on phones and smaller screens.

**Describe alternatives you've considered**

Using external apps like ElevenReader, Speechify, or Moon+ Reader, which provide a text-reflow experience but require switching apps.

**Additional context**
Attached is what my pdf looks like in Readest vs ElevenReader.

<img width="642" height="1389" alt="Image" src="https://github.com/user-attachments/assets/775867d3-9723-44f8-a4ef-d46c650ad60b" />

<img width="642" height="1389" alt="Image" src="https://github.com/user-attachments/assets/998116b9-0720-4aae-bbd7-a0d8a72d9cb1" />

hep-cogth:
Due to the sophistication of PDF specification, a temporary solution may be taking each block as unit to request TTS, like 点读机, as opposed to process the whole at once. So we offload the parsing task to the user, to make implementation possible.

ffuentesm:
I would love this if implemented on readest. Till now, I have been using Librera for the TTS function on PDFs, but they lack the word highlight readest adds when tts is enabled.

simonprovost:
+1 ! I was converting my Overleaf LaTeX paper in book based format to read on Readest, but despite being "automated" via a CI it's still no ideal :)

chrox:
It should be possible to convert tex files to EPUB? although I haven’t tried it myself yet 🙂

simonprovost:
> It should be possible to convert tex files to EPUB? although I haven’t tried it myself yet 🙂

I don't wanna spam this thread but in case you would like to point (us) out to materials that could do so, I am happy to try it/them out because in Readest I do not see any option for that unfortunately

chrox:
[...]

Issue #336, opened 2025-02-09
Smart tool menu missed


hsgyc:
## Bug Description

The smart tool menu does not appear when some words are selected.

```
Operating System: Windows 11 Professional 23H2 22631.4602
Readest Version: 0.9.10
```

chrox:
![Image](https://github.com/user-attachments/assets/1889aa8a-a504-4dd9-8648-90cf3c78f21d)

I cannot reproduce this on Windows 11 with Readest 0.9.10. Could you be more specific? In which circumstances does the annotation tool bar not appear when the text is selected?

hsgyc:
> ![Image](https://github.com/user-attachments/assets/1889aa8a-a504-4dd9-8648-90cf3c78f21d)
> 
> I cannot reproduce this on Windows 11 with Readest 0.9.10. Could you be more specific? In which circumstances does the annotation tool bar not appear when the text is selected?

![Image](https://github.com/user-attachments/assets/36274e76-179d-45d3-8b50-cbb20d0bf10c)
This bug does not appear before version 0.9.2. I am not clear what the circumstances do you need. This also exist in the web version.

chrox:
PDF annotation tools are temporarily disabled since 0.9.1 because they are still not working. We will enable them after the functions are fully ready.

chrox:
Yes you have already provided the circumstances: a screenshot, a more detailed description although it's not specific enough. The key point is that PDFs are only experimentally supported for now.

hsgyc:
I check the changelog some times and finally find this infomation in 'various fixes'. It hides deeply for me. It is a quite helpful feature and hope this can be fixed.

chenxin0397:
现在仍然在windows下不能注释PDF吗？我看Web版已经可以了

welearner:
> 现在仍然在windows下不能注释PDF吗？我看Web版已经可以了

为啥我现在在 web 上也还不行，在 PDF 中选中文字后没有任何反应

chrox:
Annotation tools are not supported in PDF yet.

yiersanwu:
> 现在仍然在windows下不能注释PDF吗？我看Web版已经可以了

Hey master, I found the PDF annotation tools in Readest on Mac Chrome still aren't working. Can you share more info about this?

chrox:
PDF annotation is not supported yet.

yiersanwu:
> Annotation tools are not supported in PDF yet.

[...]

PR #2489, opened 2025-11-20
feat(pdf): enhance PDF support with Note and TTS

ByteFlowing1337:
- annotator: Removed Highlight and Search functions for PDF , Restored Notes and Read Aloud (TTS) functions for PDF:

- sidebar: adjust header rendering based on file format

- tts: enable TTS for PDF and remove warning check

This PR removes UI buttons for features that are not supported on PDF documents and restores the ability for users to take notes and use text-to-speech (TTS) on PDFs, ensuring a more consistent annotator experience.


chrox:
I don't think this is necessary as PDF annotation and TTS for PDF will be supported in the upcoming updates.

ByteFlowing1337:
> I don't think this is necessary as PDF annotation and TTS for PDF will be supported in the upcoming updates.

Alright.Looking forward to your update.Thanks.

simonprovost:
Any news on TTS for PDF please ? 👀 @chrox

chrox:
FR of annotation tools and TTS for PDFs are tracked in #336 and #2149 respectively. Closing this PR now.

chrox:
> Any news on TTS for PDF please ? 👀 @chrox

It should be available in the next one or two updates.

PR #2517, opened 2025-11-22
Add Flathub badge in README for Linux installation

archisman-panigrahi:


chrox:
This change disrupts the layout of the Platform-Specific Downloads section, so we’ll need to hold off for now. We can revisit it later if additional download badges (such as F-Droid or Winget) are added, at which point we could move all badges to the end of the section to keep both the text and the badges neatly organized.

PR #2557, opened 2025-11-26
feat(windows): Add explorer thumbnail registration and installer hook…

411A:
**Before (1)**:
{697C4322-E829-4535-B048-3D4E2DD80830}

**After (1)**:
{FC2D592C-E33B-40E7-B2C9-0581D613592C}

**Before (2)**:
{5949446D-0666-4DCE-8DFC-D0867B00FE46}

**After (2)**:
{507C5E20-E559-4B4A-AF51-30BCB780A0FB}

📝 The `The Little Prince (Antoine de Saint-Exupery) (Z-Library).fb2` file shown (the one with the original Readest icon) didn't have a cover photo.
👮 Upon new installation, it now also requires **Administrator** rights to add and use the necessary `.dll` files for loading thumbnails.
⚠️ The thumbnail will work only if Readest is set as the default reader for the corresponding file extension.
💡 Once the thumbnails are loaded, they won't disappear normally! Even by uninstalling the Readest! As they will be cached by Windows. You can remove them by:
1. Manually cleaning the Thumbnails by following steps:
```
Settings > System > Storage > Temporary Files > Thumbnails (Check this one only) > Remove files
```
2. Using the following PowerShell (it will stop the File Explorer):
```powershell
Stop-Process explorer -Force; Remove-Item "$env:LOCALAPPDATA\Microsoft\Windows\Explorer\thumbcache_*.db" -Force -ErrorAction SilentlyContinue; Start-Process explorer
```
😁 The books shown in the screenshot were downloaded specifically for this test!

# TODO
- [ ] Unregister and remove the created `.dll` file from the installation path during uninstallation.
- [ ] Optional – Remove the larger Readest icon when the user clicks on the file and views it in the preview pane.

✅ Tested on:
- Windows 11 Pro (22H2)
- Windows Server 2025 Standard (24H2)


chrox:
Looks really awesome. 👍

chrox:
> ⚠️ This feature updates the app's installation directory on Windows to a more standard location.
> After installing the new version, users may need to delete the previous installation and re-initialize the app.
> **Previous installation path**:
> `C:\Users\USER\AppData\Local\Readest`
> **New installation path**:
> `C:\Program Files\Readest\readest.exe`

The installer already lets users choose a custom installation path, so forcing a move, and requiring everyone to manually delete the old installation, is an unnecessary breaking change and creates a poor user experience.

apps/readest-app/src-tauri/src/windows/thumbnail/extraction.rs
```diff
@@ -0,0 +1,648 @@
...
+
+/// Generate a thumbnail from a file path with caching.
+#[allow(dead_code)]
+pub fn cached_thumbnail_for_file(path: &Path, size: u32) -> Result<Vec<u8>> {
+    let ext = path
+        .extension()
+        .and_then(|e| e.to_str())
+        .ok_or_else(|| anyhow!("No file extension"))?;
+
+    let bytes = std::fs::read(path)?;
```
chrox: Avoid reading the entire file into memory. In Readest, we use a *partial MD5* strategy to quickly generate a stable digest without loading the whole file:
https://github.com/readest/readest/blob/main/apps/readest-app/src/utils/md5.ts#L11-L30

We should adopt a similar approach here:
- hash only small chunks from the beginning/middle/end of the file
- or stream through the file using BufReader

This will significantly reduce memory usage for large EPUB files and make thumbnail generation more efficient. And the cover image for ZIP-based formats or PDB (mobi/azw) files can be retrieved using random access, so there’s no need to load the entire file into memory.

chrox (changes requested):

411A:
I applied your notes and everything works on my system, but I'm not sure why the build fails. I suspect it's because I created the path `.\readest\apps\readest-app\src-tauri\src\windows` and changed how the Windows app builds in Tauri.
Could you please look into it and test it on your side when you have time?

411A:
Ohh-kaaay...!
I think it's ready. Please test it on another system; it may need some review/chores from your side.
Let me know if it needs more work for further PRs.

PR #2569, opened 2025-11-29
feat(annotations): Implement Annotations on Click Feature

### LINKED ISSUE #1845: FR: Highlighted text onclick shows popup modal with related note

> Currently, when I click on a highlighted text that already has a note, it doesn't show me the note directly. Instead, it opens a toolbar, and if I click the "Edit" button, it prompts me to create a new note for the exact same text. This workflow deviates from the conventional user experience of popular reading apps like Kindle, Google Play Books, and WeRead (微信读书). In these applications, users typically expect a single click on a highlighted area to directly reveal or allow them to edit the existing note, not to create a new one. This inconsistency is inconvenient for users who frequently review their notes.

> I propose adding a new feature when clicking a highlighted text, readest directly opens a (big) popup modal displaying its associated note.

> The existing behavior of showing a toolbar can remain the default, but an option could be added to the settings, such as "Click highlighted text to show notes." When this option is enabled, clicking on highlighted text would **bypass the toolbar and immediately open the note modal**.

> I've considered making the "direct note popup" the default behavior. However, to ensure a smooth transition and respect the existing user base, it seems more prudent to introduce this as an optional setting. This approach would cater to users who are accustomed to mainstream reading app conventions while preserving the current design for others.

-- from v2ex with ❤️

thesomeshkarthi:
Closes #1845

New support for viewing Annotations directly on click. A new settings toggle under Behaviour controls this features' enablement.


chrox:
We’re planning a major refactor of this part, so we’ll need to put this PR on hold for a while.

chrox:
Close this in favor of #2798.

chrox:
image

A small icon will now appear to indicate that notes are available.

PR #2640, opened 2025-12-07
Refactor/replacement UI

jarchenn:
## Description

- Fixes the ReplacementOptions popup positioning to ensure it stays within viewport bounds
- Eliminates the "ghost animation" where the popup would briefly appear in the wrong position before jumping to the correct location
- Ensures the popup remains stationary when other UI elements (like the pagination/TTS bar) appear on hover

## Changes

**Modified:**

- `apps/readest-app/src/app/reader/components/annotator/ReplacementOptions.tsx`
  - Added viewport boundary detection to keep popup within visible area
  - Added `isPositioned` state and `hasAdjusted` ref to calculate position only once on mount
  - Used `visibility: hidden` until position is calculated to eliminate ghost animation
  - Used `requestAnimationFrame` for accurate element measurement before positioning
  - Added `max-height: calc(100vh - 40px)` with `overflow-y: auto` for scrollable content when needed
  - Popup now repositions above the selection if it would extend beyond the bottom of the viewport

## Validation

**Manual Testing:**

1. `cd apps/readest-app`
2. Run `pnpm dev-web` (or `npx dotenv-cli -e .env.web -- next dev`)
3. Open a book in the reader
4. Test bottom-of-screen positioning:
   - Select text near the **bottom** of the screen
   - Click the Text Replacement button (The wrench/tool icon)
   - Verify popup appears directly in the correct position (no ghost/jump animation)
   - Verify all buttons are visible and clickable
5. Test stability with other UI elements:
   - With the replacement popup open, hover over the pagination/TTS bar at the bottom
   - Verify the popup does **not** move or teleport when the bottom bar appears
6. Test scrolling:
   - Select a long text and open the replacement popup
   - Verify the popup is scrollable if content exceeds viewport height

### DIFF
--- a/apps/readest-app/src/__tests__/components/ReplacementOptions.test.tsx
+++ b/apps/readest-app/src/__tests__/components/ReplacementOptions.test.tsx
@@ -3,18 +3,15 @@ import { render, screen, fireEvent, cleanup } from '@testing-library/react';
 import ReplacementOptions from '@/app/reader/components/annotator/ReplacementOptions';
 
 describe('ReplacementOptions Component', () => {
-  const mockHandlers = {
-    onFixOnce: vi.fn(),
-    onFixInBook: vi.fn(),
-    onFixInLibrary: vi.fn(),
-    onClose: vi.fn(),
-  };
+  const mockOnConfirm = vi.fn();
+  const mockOnClose = vi.fn();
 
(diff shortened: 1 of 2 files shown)


jarchenn:
Closed because wrong repository

PR #2655, opened 2025-12-09
add rust and node deps to the nix devshell

jpetrucciani:
this should add the remaining deps (a rust toolchain and nodejs 22, plus all required libraries to build with and link for tauri) to the nix development shell!

with these changes, I am able to enter the devshell and build+run the full tauri dev setup on linux and darwin!

bare minimum steps using this after a fresh clone:
```bash
git submodule update --init --recursive
pnpm install
pnpm --filter @readest/readest-app setup-vendors

# run just the web version
pnpm dev-web

# build and run tauri build
pnpm tauri dev
```

this does not fully solve the actual production release builds via the nix env, but that is handled by CI anyways at the moment for all platforms!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,5 @@ fastlane/report.xml
 
 *.koplugin.zip
 
+# nix
+result*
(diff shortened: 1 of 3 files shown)


chrox:
Looks great. Thanks.

Issue #2815, opened 2025-12-31
Font dialog settings are not propagating to all panes when Global Settings is enabled


aangelinsf:
## Bug Description

Open two books in Parallel Read mode.
Click on the font dialog.
Click on the submenu and ensure that Global Settings is enabled.
Change the font size.

## Expected behavior
All opened panes should see their font size change.

## Actual behavior
Only the most recently-opened pane changes font size.

## Technical inputs

Operating System:
macOS 26.2

Readest Version:
0.9.97

## Additional issues
Issue 2: when Global Settings is off, the font dialog always affects the most recently opened pane even if one clicks on the font icon for a different pane.
Issue 3: when the Global Settings is on, change the font size. Then disable Global Settings and the settings in the font dialog should revert to the pane-specific settings but do not. Toggle again and they should display the global settings again.

I can:
• submit a fix for this issue, which I've solved locally. The app should iterate through the multiple property stores and apply the settings.
• create two additional separate issues for the other bugs.

Issue 2 is proving to be trickier. I'm not a developer in this stack.

Edit: I think I have a fix for Issue 2 and that resolve Issue 3 at the same time, it seems.

How would you like me to proceed?

chrox:
Thanks a lot for the detailed report and for digging into this 🙏

That sounds great. Please go ahead and send a PR for Issue 1 with the fix you’ve implemented. Iterating through the multiple property stores and applying the settings is exactly the right approach.

For the other items, please open separate issue tickets for Issue 2 and Issue 3 so we can track and discuss them independently.

Really appreciate you taking the time to investigate and contribute. Thanks again!

aangelinsf:
Will do and my pleasure.

chrox:
Fixed in #2818.

PR #2818, opened 2025-12-31
fix(settings): ensure global settings sync across all open panes

aangelinsf:
Issue: https://github.com/readest/readest/issues/2815

Previously, changing a global setting (like font size) only updated the currently active book pane, causing split-view panes to desynchronize.

This commit updates the settings logic to propagate global changes to all open book instances immediately. It also ensures that settings UI panels correctly re-render when preferences are updated from outside the component.

Build check passed.

BTW, the guidelines say to run `pnpm build` for linting etc. but I had to run `pnpm --filter @readest/readest-app build-check`


aangelinsf:
@chrox ready for your review.

chrox:
Sure.

chrox:
Thanks.

aangelinsf:
Knowing the codebase would have made that a bit neater :-).

PR #2879, opened 2026-01-07
fix(login): Fix login message field

lcd1232:
Fix the invalid field used in login failed screen.
How to test `http https://readest.supabase.co/auth/v1/token email=user@example.com password='123' grant_type==password apikey:'YOURAPIKEY'`

 It returns: 
 ```
{
    "code": 400,
    "error_code": "invalid_credentials",
    "msg": "Invalid login credentials"
}
 ```


chrox:
Good catch. Thanks.

PR #2906, opened 2026-01-10
fix(fonts): update font CDN links to use deno.dev for improved reliability

KonghaYao:
字图 CDN 源站已经被墙 nelify.app，可以使用 deno.dev 提供的 CDN 链接为中国境内提供服务
image


chrox:
Nice! Thank you for the patch.

chrox:
@KonghaYao When accessing the font from `chinese-fonts-cdn.deno.dev` it returns 403 error. And other fonts from Google Fonts works fine.

image

image

chrox:
@KonghaYao We might need to add `tauri://localhost` to the allowed origins for a more permissive CORS policy. For now, I just hotfix this by accessing the files from `ik.imagekit.io` in https://github.com/readest/readest/pull/2944.

PR #2984, opened 2026-01-18
perf(pdf): pre-render next page for PDFs

### LINKED ISSUE #13: Flickering occurs when turning pages in image-based PDF files
## Bug Description

When reading certain image-based PDF files, the screen flashes briefly on every page turn, making the reading experience uncomfortable.

**Example:**

https://github.com/user-attachments/assets/0b8f7f29-f5d4-43af-961a-d8cb3f3b7b1a

## Expected behavior

- Page transitions should be smooth and should use a subtle animation that avoids any harsh flickering.

- The transition animation should adapt appropriately to the selected theme (Light or Dark mode) for a consistent reading experience.

### LINKED ISSUE #1911: FR: preloading next section for better performance

> When reading books with high quality images, there is a noticible stutter/freeze, when you flip to the next page. Even on a really high-end pc.

> It would be beneficial, if the next pages would be pre-rendered. The same way apps like TikTok pre-load the next few videos, so we get a much better UX.

> I would not recommend any other alternatives.

**Additional context**

> Image Screenshot taken while the page is loading (it only takes around 150 ms in reality).

chrox:
This should close #1911 and close #13.

### DIFF
--- a/apps/readest-app/src/app/reader/components/FoliateViewer.tsx
+++ b/apps/readest-app/src/app/reader/components/FoliateViewer.tsx
@@ -188,7 +188,9 @@ const FoliateViewer: React.FC<{
         setViewSettings(bookKey, { ...viewSettings });
       }
 
-      mountAdditionalFonts(detail.doc, isCJKLang(bookData.book?.primaryLanguage));
+      if (!bookData?.isFixedLayout) {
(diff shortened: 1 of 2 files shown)


Kian738:
How would you assess a caching functionality? Possibly even caching some low-quality render of the page to disk?

--- I just read the PR Code and saw the caching impl. looks solid!

Issue #179, opened 2025-01-15
Support importing a directory of books recursively


chrox:
**Does your feature request involve difficulty completing a task? Please describe.**
Support importing a directory of books recursively instead of selecting files manually in an organized book library.

eg9y:
When you say “recursive,” do you mean including books not only from the selected folder but also from all of its subfolders—no matter how deeply nested they are? 

Or do you only require importing books that are directly inside the chosen folder?

eg9y:
It seems there's an web API for selecting directories ([window.showDirectoryPicker](https://developer.mozilla.org/en-US/docs/Web/API/Window/showDirectoryPicker#browser_compatibility)), but it's still considered experimental which not all browsers support. Should we add this feature for desktop only for now?

Essentially, the Add Book button will now become a dropdown providing the option to select a single file or a folder. we could provide a conditional so that this dropdown behavior only applies to desktop, meaning we would still only able to select single files for the web version.

![Image](https://github.com/user-attachments/assets/6ab768ee-8c51-4373-a625-d53b86a566c8)

chrox:
> When you say “recursive,” do you mean including books not only from the selected folder but also from all of its subfolders—no matter how deeply nested they are?
> 
> Or do you only require importing books that are directly inside the chosen folder?

Yes when it is possible, it should import books in all subdirectories if the user select a directory of books to import.

chrox:
> It seems there's n web API for selecting directories ([window.showDirectoryPicker](https://developer.mozilla.org/en-US/docs/Web/API/Window/showDirectoryPicker#browser_compatibility)), but it's still considered experimental that not all browsers support. Should we add this feature for desktop only for now?
> 
> Essentially, the Add Book button will now become a dropdown providing the option to select a single file or a folder. we could provide a conditional so that this dropdown behavior only applies to desktop, meaning we would still only able to select single files for the web version.
> 
> ![Image](https://github.com/user-attachments/assets/6ab768ee-8c51-4373-a625-d53b86a566c8)

Yeah it’s much easier to only support importing directory of books on desktop version. And most use cases should be on desktop.

eg9y:
https://github.com/readest/readest/pull/194

I gave this a shot in this PR, it's functional, but not performant. lmk if anyone has a suggestion

red-co:
+1

charlie-nickerson:
Is this issue still open? I would like to take a crack at it!

chrox:
[...]

PR #2993, opened 2026-01-19
fix(perf): replace JS recursion with native Rust walkdir for imports

Whitestar14:
**Description:**
The current JavaScript-based `readDirRecursively` logic fails on some Windows 11 configurations due to permission/path handling issues with `tauri-plugin-fs`. It also causes UI freezing on large libraries due to excessive IPC calls.

This PR moves the scanning logic to the Rust backend using the `walkdir` crate.

the benefits being:
- fixes the Windows 11 recursion bug (Issue #179 & #2642).
- massive performance boost (1 IPC call vs 1000+ such as in very large libraries).
- prevents UI/Toast crash on mass imports.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,6 +55,7 @@ dependencies = [
  "thiserror 2.0.17",
  "tokio",
  "tokio-util",
+ "walkdir",
 ]
 
 [[package]]
--- a/apps/readest-app/src-tauri/Cargo.toml
+++ b/apps/readest-app/src-tauri/Cargo.toml
@@ -26,6 +26,7 @@ serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 log = "0.4"
 thiserror = "2"
+walkdir = "2"
 tokio = { version = "1", features = ["fs"] }
 tokio-util = { version = "0.7", features = ["codec"] }
 futures-util = "0.3"
@@ -54,7 +55,6 @@ tauri-plugin-native-bridge = { path = "./plugins/tauri-plugin-native-bridge" }
 tauri-plugin-native-tts = { path = "./plugins/tauri-plugin-native-tts" }
 tauri-plugin-websocket = "2"
 tauri-plugin-sharekit = "0.3"
-walkdir = "2"
 
 [target."cfg(target_os = \"macos\")".dependencies]
 rand = "0.8"
--- a/apps/readest-app/src-tauri/src/dir_scanner.rs
+++ b/apps/readest-app/src-tauri/src/dir_scanner.rs
@@ -1,0 +1,96 @@
+use std::path::Path;
+use tauri::AppHandle;
+use tauri_plugin_fs::FsExt;
+use walkdir::WalkDir;
+
+#[derive(serde::Serialize)]
+pub struct ScannedFile {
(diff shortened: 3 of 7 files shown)


chrox:
Thanks for the PR and for the detailed explanation.

We’ve already fixed the recursive directory permission issue in https://github.com/readest/readest/pull/2972, so at this point the app should be able to read all contents of a user-selected directory correctly.

That said, the choice to keep directory traversal on the JavaScript side is deliberate and primarily driven by security considerations rather than implementation convenience.

Readest relies on Tauri’s filesystem scope model to strictly limit file access to whitelisted paths. These paths are defined in the project configuration and can only be extended explicitly through user actions, such as selecting a directory via the system file or directory picker. By using the JavaScript filesystem APIs, all file access is automatically constrained by this scope enforcement. Introducing a Rust-side command like scan_library_recursive that accepts arbitrary paths effectively bypasses this model and exposes the application to unrestricted filesystem access. On platforms without a strong OS-level sandbox, this represents a significant expansion of the attack surface. This is particularly relevant because Readest reads and renders EPUB and PDF files. While we carefully sandbox the rendering engine, these formats can still embed scripts or other crafted content. In a worst-case scenario, if an attacker manages to execute carefully constructed malicious code within the app context, Tauri’s security model ensures that, even with JavaScript API access, the code remains confined to whitelisted filesystem paths. Allowing unrestricted Rust-side directory scanning would weaken this guarantee.

Regarding performance: if you’ve observed measurable performance differences between the current JavaScript-based directory reading and a Rust walkdir implementation, it would be very helpful if you could share concrete benchmark data. With quantitative evidence, we can better evaluate whether targeted optimizations are warranted while still preserving the existing security boundaries.

Whitestar14:
@chrox Thanks for the feedback. I've updated the PR to address the security concern.
1. On the permissions. I added a check using `app.fs_scope().is_allowed(&path) `inside the Rust command. This ensures it respects the exact same whitelist as the JS API and only scans folders the user explicitly opened via the dialog.
```rust
if !scope.is_allowed(&path_buf) {
    return Err("Permission denied: Path not in filesystem scope".to_string());
}
```
2. And on performance benchmarking I tried to capture a clean benchmark but honestly, it was difficult because the current JS implementation hits a bottleneck on my test machine (Windows 11, 8GB RAM, HDD) that causes the UI thread to hang/freeze completely during the scan (150+ books). And for some reason, all of my workarounds to log to console from `nativeAppService` failed.
The Rust implementation doesn't have that overhead because it turns thousands of IPC calls into a single one. It resolves the freezing issue entirely on my end.

Also, one other thing:
While I'm was in the codebase, I noticed some areas in the UI (typography consistency, padding, etc.) that could use some polish. Would you be open to a separate PR for some UI/UX cleanups?

chrox:
Sounds like a good solution 👍
I’ll do some testing on my side before merging this PR.

A separate PR for the UI/UX cleanups is totally fine that would actually be appreciated.

Whitestar14:
ive applied the fixes, but regarding the scope check, I kept it in place as requested to ensure we strictly adhere to the user-selected permission model.

Ready for testing

chrox:
LGTM, thanks.

PR #3009, opened 2026-01-20
feat(ui): switch typography to Inter

Whitestar14:
### Swapping the default application font to Inter (self-hosted).

Currently, Readest relies on the system font stack.
* On macOS/iOS, this defaults to San Francisco, which looks great.
* On Windows/Android, it falls back to Segoe UI or Roboto, which can feel a bit utilitarian and dated (giving off 2016-era Android vibes) for a modern content app.

Using Inter ensures that on:
* Windows/Android: Users get a massive visual upgrade that feels cleaner and more premium immediately.
* macOS: Since Inter is already visually similar to San Francisco, Mac users retain the native feel they expect, but now the UI is consistent across all platforms.

Changes
* Added `Inter-Variable.ttf` to public/fonts
* Updated `tailwind.config.ts` to prioritize Inter.
* Found unused geist font assets and removed them to save space.

Note:
This is ground zero for some larger UI/UX polishing I'm working on. I wanted to land the typography foundation first before proposing some layout/spacing cleanups in a future PR.


apps/readest-app/src/styles/globals.css
```diff
@@ -82,15 +85,17 @@ foliate-fxl {
...
     transform: scale(1.02);
   }
+
   100% {
     transform: scale(1);
   }
 }
 
-.hover-bar-anim:hover ~ .foliate-viewer {
+.hover-bar-anim:hover~.foliate-viewer {
```
chrox: We should rely on the Prettier VS Code plugin to format CSS files. It prefers the spaced style, which is more readable. We probably also need a configuration to enforce Prettier formatting across all CSS, JS, TS, and TSX files.

chrox (changes requested):

apps/readest-app/public/fonts/Inter-Variable.woff2
```diff
@@ -0,0 +1 @@
+Not Found
```
chrox: The font file seems corrupted. Please redownload it.

Whitestar14: I tried using curl to download the font and it appears that it didn't go through. I've manually added it into the public/fonts folder now, so it should resolve without any issues.

chrox:
Looks great.

PR #2740, opened 2025-12-19
feat(ai): AI reading assistant - phase 1

WhiteHades:
closes #2738.

completes phase 1 (of a proposed 3 or more if it's necessary) of the ai assistant integration.

PR #3023, opened 2026-01-21
feat(ai): AI reading assistant phase 2

### LINKED ISSUE #2738: FR: AI reading assistant - phase 1
### `summary`
this issue tracks the _phase 1_ implementation of the ai reading assistant as talked about in discussion #2727, establishing the core infrastructure (service layer, providers, settings ui) to enable both local (ollama) and cloud (openrouter) rag.

### `phase 1 - foundation and provider architecture`
the goal is to build a type safe foundation that abstracts away provider differences, allowing users to switch between local and cloud models.

### `core changes`
- service layer - created `src/services/ai/` directory structure.
- interfaces - defined `AIProvider` and `AISettings` contracts to ensure consistency across implementations.
- factory pattern - implemented `getAIProvider()` factory to instantiate the correct provider based on runtime settings.

### `provider implementations`
1.  ollama
    - uses `ollama/browser` sdk.
    - requires no api keys.
    - implements `chatStream` for token streaming.
    - implements `embed` for vector generation (`nomic-embed-text`).
    - includes `healthCheck()` to verify local server status.

2.  openrouter
    - uses `openai` sdk with custom baseurl.
    - provides access to claude, gpt, llama, gemini via single key.
    - handles auth headers (`HTTP-Referer`, `X-Title`) for openrouter analytics.

### `ui`
- new component - `src/components/settings/AIPanel.tsx`
- detection - automatically detects running ollama models and populates dropdowns.
- integration - matches existing `FontPanel`/`ColorPanel` design system.
- validation - "Test Connection" button verifies api keys/local server reachability before saving.

### `phase 2.1` (will create separate issue later when working on it)
- implement the core "reading aware" logic.
- add `chunker` for text segmentation.
- implement vector store + bm25 index.
- add spoiler filter logic based on `readerStore` progress.

### `phase 2.2` (will create separate issue later when working on it)
- connect this settings foundation to a sidebar chat interface.
- add ai tab to reader sidebar.
- implement `chatService` orchestration.
- enable "explain this" context menu action.

### `phase 3` (will create separate issue later when working on it)
- background indexing via web workers.
- offline error recovery.
- automated tests for rag precision.

WhiteHades:
closes #2738.
closes #2740 (this was phase 1, this current PR supersedes it).

this PR represents the implementation of the AI reading assistant (phase 2). it combines the foundation laid phase 1 with a fully working code.

### `features (non exhaustive)`
- "readest" persona: AI reading companion that "reads" with the reader. it uses "we/us" language, refuses to spoil future events, and has strict anti jailbreak guardrails.
- page based spoiler protection: precise, granular filtering (1500 chars/page) that strictly limits the AI's context to exactly what is have read so far.
- dual provider support: integration for both [ollama](https://ollama.com/) (local) and [vercel AI gateway](https://vercel.com/ai-gateway) (cloud) models.
- full rag pipeline: chunking (with page tracking), hybrid search (both bm25 and vector), context injection.
- ui/ux: [assistant-ui](https://www.assistant-ui.com/) for the chat experience, integrated directly into the sidebar with manual re-indexing (vector embedding) and a delete button.
- default state: AI features disabled, there are global enable / disable buttons.

### `how to use the AI features (i tested only in the browser/web version)`
testing can be done via the UI.
- go to `settings -> ai`
- for vercel ai gateway (cloud): select "AI Gateway" and paste the key directly in the ui. get a key from [vercel.com/ai-gateway](https://vercel.com/ai-gateway) ($5 free credits).
- for ollama (local): select "Ollama" (ensure it's running in the background)
- open a book, open the AI sidebar, index the book and start chatting.

note: the backend infrastructure (`route.ts`) is designed to support a hybrid model. it prioritises user supplied keys from the ui, but automatically falls back to the server side `AI_GATEWAY_API_KEY` system variable if provided. this should enable the goal for future free-tier users while keeping the option open for readers to bring their own keys.

### `screenshots`
ai_sidepanel_1
ai_sidepanel_2
ai_sidepanel_3
ai_sidepanel_4

ai_setting_1
ai_setting_2
ai_setting_3

### `issues & non AI fixes`
getting this to work was quite something. the `@assistant-ui/react` library had some rough edge cases with react 19's strict mode that caused runtime crashes. patching had to be done at multiple levels.

_assistant-ui patches_ (`patches/@assistant-ui__react.patch`)
- fixed `TypeError: Cannot set properties of undefined (setting '_getInitializePromise')` by adding guards in both `src` and `dist` files. react's double-invocation during strict mode was causing `this` to be undefined.
- fixed `tapLookupResources: Resource not found error` on thread deletion by adding a fallback handler for missing resources.

_non ai code changes_
- `utils/file.ts` & `utils/file.web.ts`: added browser safe file utilities and dynamic tauri fs import to prevent web crashes
- `services/nativeAppService.ts`: deferred `osType` call to `init()` for web compatibility
- `services/webAppService.ts`: imported `RemoteFile` from `file.web` for consistency
these changes are non destructive, the existing readest functionality remains fully intact, at least theoretically. didn't test an exhaustive list (to be transparent).

### `setup for testing`
kindly follow these steps exactly to avoid the pitfalls i fell into.

```
# clean slate (stale node_modules can cause patch mismatches, learned it the hard way lol)
rm -rf node_modules apps/readest-app/node_modules

# new install, applies the patches automatically
pnpm install

# run web dev server
pnpm dev-web
```

importantly, when you open the app, do a hard refresh (`cmd/ctrl + shift + r`) and clear site data (`devtools -> application -> storage -> clear site data`). since i bumped the DB to v2 for the AI indices, any stale storage from previous versions will crash the rag pipeline. starting fresh is mandatory here. also, if you see any `TypeError: Cannot set properties of undefined` or `Resource not found` errors, the patch likely didn't apply correctly. run `rm -rf node_modules && pnpm install` again to ensure the patch file (`patches/@assistant-ui__react.patch`) is applied to both `src` and `dist` library files.

### DIFF
--- a/apps/readest-app/src/services/nativeAppService.ts
+++ b/apps/readest-app/src/services/nativeAppService.ts
@@ -38,7 +38,7 @@ import {
 import { getOSPlatform, isContentURI, isFileURI, isValidURL } from '@/utils/misc';
 import { getDirPath, getFilename } from '@/utils/path';
 import { NativeFile, RemoteFile } from '@/utils/file';
-import { copyURIToPath } from '@/utils/bridge';
+import { copyURIToPath, getStorefrontRegionCode } from '@/utils/bridge';
 import { copyFiles } from '@/utils/files';
 
 import { BaseAppService } from './appService';
@@ -421,6 +421,8 @@ export class NativeAppService extends BaseAppService {
   override canCustomizeRootDir = DIST_CHANNEL !== 'appstore';
   override canReadExternalDir = DIST_CHANNEL !== 'appstore' && DIST_CHANNEL !== 'playstore';
   override distChannel = DIST_CHANNEL;
+  override storefrontRegionCode: string | null = null;
+  override isOnlineCatalogsAccessible = true;
 
   private execDir?: string = undefined;
 
@@ -445,6 +447,14 @@ export class NativeAppService extends BaseAppService {
         isPortable: this.isPortableApp,
         execDir,
       });
+    }
+    if (this.isIOSApp) {
+      const res = await getStorefrontRegionCode();
+      if (res.regionCode) {
+        this.storefrontRegionCode = res.regionCode;
+        this.isOnlineCatalogsAccessible =
+          this.storefrontRegionCode.toLowerCase() !== 'chn' || this.distChannel !== 'appstore';
+      }
     }
     await this.prepareBooksDir();
     await this.runMigrations();
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -872,7 +872,6 @@
   "Show Results": "عرض النتائج",
   "Clear search": "مسح البحث",
   "Clear search history": "مسح سجل البحث",
-  "Quick action disabled": "الإجراء السريع معطل",
   "Tap to Toggle Footer": "انقر لتبديل التذييل",
   "Exported successfully": "تم التصدير بنجاح",
   "Book exported successfully.": "تم تصدير الكتاب بنجاح.",
@@ -930,5 +929,8 @@
   "Select Voice": "اختر الصوت",
   "Toggle Sticky Bottom TTS Bar": "تبديل شريط TTS الثابت",
   "Display what I'm reading on Discord": "عرض ما أقرأه على Discord",
-  "Show on Discord": "عرض على Discord"
+  "Show on Discord": "عرض على Discord",
+  "Instant {{action}}": "{{action}} فوري",
+  "Instant {{action}} Disabled": "تم تعطيل {{action}} فوري",
+  "Annotation": "التعليق"
 }
--- a/apps/readest-app/public/locales/bn/translation.json
+++ b/apps/readest-app/public/locales/bn/translation.json
@@ -832,7 +832,6 @@
   "Show Results": "ফলাফল দেখুন",
   "Clear search": "অনুসন্ধান সাফ করুন",
   "Clear search history": "অনুসন্ধান ইতিহাস সাফ করুন",
-  "Quick action disabled": "দ্রুত কার্যক্রম অক্ষম",
   "Tap to Toggle Footer": "ফুটার টগল করতে আলতো চাপুন",
   "Exported successfully": "সফলভাবে রপ্তানি হয়েছে",
   "Book exported successfully.": "বইটি সফলভাবে রপ্তানি হয়েছে।",
@@ -890,5 +889,8 @@
   "Select Voice": "ভয়েস নির্বাচন করুন",
   "Toggle Sticky Bottom TTS Bar": "স্থির TTS বার টগল করুন",
   "Display what I'm reading on Discord": "Discord-এ পড়ছি যা দেখান",
-  "Show on Discord": "Discord-এ দেখান"
+  "Show on Discord": "Discord-এ দেখান",
+  "Instant {{action}}": "তাৎক্ষণিক {{action}}",
+  "Instant {{action}} Disabled": "তাৎক্ষণিক {{action}} নিষ্ক্রিয়",
+  "Annotation": "টীকা"
 }
--- a/apps/readest-app/public/locales/bo/translation.json
+++ b/apps/readest-app/public/locales/bo/translation.json
@@ -822,7 +822,6 @@
   "Show Results": "འབྲས་བུ་སྟོན།",
   "Clear search": "འཚོལ་བཤེར་གཙང་སེལ།",
   "Clear search history": "འཚོལ་བཤེར་ལོ་རྒྱུས་གཙང་སེལ།",
-  "Quick action disabled": "མྱུར་མགྱོགས་བྱ་འགུལ་བཀག་འགོག",
   "Tap to Toggle Footer": "ཞབས་མཇུག་སྒོ་འབྱེད་བྱེད་པར་གནོན།",
   "Exported successfully": "ཕྱིར་འདོན་ལེགས་གྲུབ་བྱུང་།",
   "Book exported successfully.": "དཔེ་དེབ་ཕྱིར་འདོན་ལེགས་གྲུབ་བྱུང་།",
@@ -880,5 +879,8 @@
   "Select Voice": "སྐད་གདངས་འདེམས།",
   "Toggle Sticky Bottom TTS Bar": "TTS སྡོམ་ཐིག་བརྗེ་བ།",
   "Display what I'm reading on Discord": "Discord ཐོག་ཀློག་བཞིན་པའི་དཔེ་ཆ་སྟོན།",
-  "Show on Discord": "Discord ཐོག་སྟོན།"
+  "Show on Discord": "Discord ཐོག་སྟོན།",
+  "Instant {{action}}": "འཕྲལ་མར་{{action}}",
+  "Instant {{action}} Disabled": "འཕྲལ་མར་{{action}}་ལྕོགས་མིན་བཟོས།",
+  "Annotation": "མཆན་འགྲེལ།"
 }
--- a/apps/readest-app/public/locales/de/translation.json
+++ b/apps/readest-app/public/locales/de/translation.json
@@ -832,7 +832,6 @@
   "Show Results": "Ergebnisse anzeigen",
   "Clear search": "Suche löschen",
   "Clear search history": "Suchverlauf löschen",
-  "Quick action disabled": "Schnellaktion deaktiviert",
   "Tap to Toggle Footer": "Tippen, um Fußzeile umzuschalten",
   "Exported successfully": "Erfolgreich exportiert",
   "Book exported successfully.": "Buch erfolgreich exportiert.",
@@ -890,5 +889,8 @@
   "Select Voice": "Stimme auswählen",
   "Toggle Sticky Bottom TTS Bar": "Fixierte TTS-Leiste umschalten",
   "Display what I'm reading on Discord": "Zeige was ich auf Discord lese",
-  "Show on Discord": "Auf Discord zeigen"
+  "Show on Discord": "Auf Discord zeigen",
+  "Instant {{action}}": "Sofort {{action}}",
+  "Instant {{action}} Disabled": "Sofort {{action}} deaktiviert",
+  "Annotation": "Anmerkung"
 }
--- a/apps/readest-app/public/locales/el/translation.json
+++ b/apps/readest-app/public/locales/el/translation.json
@@ -832,7 +832,6 @@
   "Show Results": "Εμφάνιση αποτελεσμάτων",
   "Clear search": "Εκκαθάριση αναζήτησης",
   "Clear search history": "Εκκαθάριση ιστορικού αναζήτησης",
-  "Quick action disabled": "Η γρήγορη ενέργεια απενεργοποιήθηκε",
   "Tap to Toggle Footer": "Πατήστε για εναλλαγή υποσέλιδου",
   "Exported successfully": "Εξαγωγή επιτυχής",
   "Book exported successfully.": "Το βιβλίο εξήχθη επιτυχώς.",
@@ -890,5 +889,8 @@
   "Select Voice": "Επιλογή φωνής",
   "Toggle Sticky Bottom TTS Bar": "Εναλλαγή καρφιτσωμένης μπάρας TTS",
   "Display what I'm reading on Discord": "Εμφάνιση του βιβλίου που διαβάζω στο Discord",
-  "Show on Discord": "Εμφάνιση στο Discord"
+  "Show on Discord": "Εμφάνιση στο Discord",
+  "Instant {{action}}": "Άμεση {{action}}",
+  "Instant {{action}} Disabled": "Άμεση {{action}} απενεργοποιημένη",
+  "Annotation": "Σημείωση"
 }
--- a/apps/readest-app/public/locales/es/translation.json
+++ b/apps/readest-app/public/locales/es/translation.json
@@ -842,7 +842,6 @@
   "Show Results": "Mostrar resultados",
   "Clear search": "Borrar búsqueda",
   "Clear search history": "Borrar historial de búsqueda",
-  "Quick action disabled": "Acción rápida desactivada",
   "Tap to Toggle Footer": "Toca para alternar pie de página",
   "Exported successfully": "Exportado con éxito",
   "Book exported successfully.": "Libro exportado con éxito.",
@@ -900,5 +899,8 @@
   "Select Voice": "Seleccionar voz",
   "Toggle Sticky Bottom TTS Bar": "Alternar barra TTS fija",
   "Display what I'm reading on Discord": "Mostrar lo que estoy leyendo en Discord",
-  "Show on Discord": "Mostrar en Discord"
+  "Show on Discord": "Mostrar en Discord",
+  "Instant {{action}}": "{{action}} instantáneo",
+  "Instant {{action}} Disabled": "{{action}} instantáneo desactivado",
+  "Annotation": "Anotación"
 }
--- a/apps/readest-app/public/locales/fa/translation.json
+++ b/apps/readest-app/public/locales/fa/translation.json
@@ -832,7 +832,6 @@
   "Show Results": "نمایش نتایج",
   "Clear search": "پاک کردن جستجو",
   "Clear search history": "پاک کردن تاریخچه جستجو",
-  "Quick action disabled": "عملکرد سریع غیرفعال است",
   "Tap to Toggle Footer": "برای تغییر پاورقی ضربه بزنید",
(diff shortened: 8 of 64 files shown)


WhiteHades:
this PR establishes the core AI stuffs (phase 2). i plan to expand this for phase 3 if all goes well. i looked into the excellent KOreader AI plugin for some inspirations. below are a couple of features i think is cool and would like to add going forward (not definitive per say but just throwing it out there to see what sticks eventually, and what doesn't).

- `X-ray`: generating individual character history (complete with their names and portraits), the overarching plot so far and how each character plays a part in said plot, and all these information are bound by the reading progress, so, no spoilers
- `summarisation`: generate smart recaps if readers haven't opened the book in 28+ hours
- `character image generation`: generates character portraits, the art style matching the theme and the writings of the book
- `visual artifacts`: generates images for complex fantasy/sci-fi items, vehicles, creatures, scenery, architectures, ambience, etc. described in the book

it is very evident that these features are geared more towards the fiction/fantasy genre and so, i welcome any feedback on whether implementing these would be a good idea and also any other features, for other genres or even the said ones, that anyone would like to see being implemented, please don't hesitate to mention here. thanks.

xrishox:
this feature is way more ambitious than i even imagined when i first saw it mentioned. very excited to see this hopefully get merged!

especially excited that you plan to support local models.

chrox:
This is a huge PR we will need some time to review it. Thanks.

.gitignore
```diff
@@ -28,6 +28,10 @@ yarn-error.log*
 
 # local env files
 .env*.local
+.env
```
chrox: The `.env` is needed to be committed in the code base for self-hosting with the default backend and also for `.env.tauri` and `.env.web`.

apps/readest-app/src/services/nativeAppService.ts
```diff
@@ -24,7 +24,7 @@ import {
   appLogDir,
   tempDir,
 } from '@tauri-apps/api/path';
-import { type as osType } from '@tauri-apps/plugin-os';
+// osType is dynamically imported to prevent crashes in web mode
```
chrox: If `NEXT_PUBLIC_APP_PLATFORM` is properly set in the `.env.tauri` or `.env.web` the `nativeAppService.ts` or the `webAppService.ts` is imported respectively:
See https://github.com/readest/readest/blob/main/apps/readest-app/src/services/environment.ts#L52-L60

And when developing for the web version use `pnpm dev-web`; and when developing for the tauri version use `pnpm tauri dev`. It will not have this problem.

chrox (changes requested):

packages/foliate-js
chrox, line 1: Please rebase with the main branch of https://github.com/readest/readest.

packages/tauri
chrox, line 1: Please rebase with the main branch of https://github.com/readest/readest.

packages/tauri-plugins
chrox, line 1: Please rebase with the main branch of https://github.com/readest/readest.

apps/readest-app/src/services/webAppService.ts
```diff
@@ -1,6 +1,6 @@
 import { FileSystem, BaseDir, AppPlatform, ResolvedPath, FileItem } from '@/types/system';
 import { getOSPlatform, isValidURL } from '@/utils/misc';
-import { RemoteFile } from '@/utils/file';
+import { RemoteFile } from '@/utils/file.web';
```
chrox: No need for a separate file.web module. Please fix this.

apps/readest-app/src/services/nativeAppService.ts
```diff
@@ -57,7 +57,13 @@ declare global {
   }
 }
 
-const OS_TYPE = osType();
+// osType is dynamically fetched in NativeAppService.init() to prevent crashes in web mode
```
chrox: nativeAppService.ts ought not be imported in web mode thus this is not necessary.

chrox (changes requested):

chrox:
> _assistant-ui patches_ (`patches/@assistant-ui__react.patch`)
> 
> * fixed `TypeError: Cannot set properties of undefined (setting '_getInitializePromise')` by adding guards in both `src` and `dist` files. react's double-invocation during strict mode was causing `this` to be undefined.
> * fixed `tapLookupResources: Resource not found error` on thread deletion by adding a fallback handler for missing resources.

Is there any cross reference links of this issue on the @assistant-ui project?

chrox:
> _non ai code changes_
> 
> * `utils/file.ts` & `utils/file.web.ts`: added browser safe file utilities and dynamic tauri fs import to prevent web crashes
> * `services/nativeAppService.ts`: deferred `osType` call to `init()` for web compatibility
> * `services/webAppService.ts`: imported `RemoteFile` from `file.web` for consistency
>   these changes are non destructive, the existing readest functionality remains fully intact, at least theoretically. didn't test an exhaustive list (to be transparent).

`nativeAppService.ts` should never be imported in the web version, please check the environment settings especially if the 
`.env.web` file has proper variable `NEXT_PUBLIC_APP_PLATFORM=web`, see
https://github.com/readest/readest/blob/main/apps/readest-app/package.json#L9

chrox:
It would be great to move the AIAssistant into the Notebook on the right sidebar, where most user input activities, such as note-taking and interactive chat, take place. Meanwhile, the left sidebar could host chat history alongside the existing annotation logs, serving primarily as a read-only reference area.

apps/readest-app/.gitignore
```diff
@@ -25,9 +25,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-# local env files
-.env*.local
```
chrox, line 29: The .env*.local files should be ignored and cannot be committed to the code base as they contain confidential settings like API keys and tokens.

.gitignore
```diff
@@ -27,7 +27,7 @@ yarn-debug.log*
 yarn-error.log*
 
 # local env files
-.env*.local
```
chrox, line 30: We actually have many local envs for different platforms.

image

apps/readest-app/src/utils/file.ts
```diff
@@ -1,6 +1,15 @@
-import { FileHandle, open, BaseDirectory, SeekMode } from '@tauri-apps/plugin-fs';
+import type { FileHandle, BaseDirectory, SeekMode } from '@tauri-apps/plugin-fs';
 import { getOSPlatform } from './misc';
 
+// dynamic import for tauri fs plugin, prevents crashes in web mode
+const isTauri = () =>
+  typeof window !== 'undefined' && (window as unknown as { __TAURI__?: unknown }).__TAURI__ != null;
+
+async function fs() {
```
chrox: No need for this.

apps/readest-app/src/services/nativeAppService.ts
```diff
@@ -425,6 +423,31 @@ export class NativeAppService extends BaseAppService {
   private execDir?: string = undefined;
 
   override async init() {
+    // initialize all OS_TYPE-dependent properties now
```
chrox: No need for this.

chrox:
The project at https://github.com/olivernn/lunr.js hasn’t been updated for about six years, which suggests it may no longer be actively maintained. What advantages does building an inverted index over the entire document provide, compared to performing term searches at runtime with proper caching, as many modern code agents do today?

WhiteHades:
> > _assistant-ui patches_ (`patches/@assistant-ui__react.patch`)
> > 
> > * fixed `TypeError: Cannot set properties of undefined (setting '_getInitializePromise')` by adding guards in both `src` and `dist` files. react's double-invocation during strict mode was causing `this` to be undefined.
> > * fixed `tapLookupResources: Resource not found error` on thread deletion by adding a fallback handler for missing resources.
> 
> Is there any cross reference links of this issue on the @assistant-ui project?

i couldn’t find an existing assistant-ui issue/PR that explicitly references `_getInitializePromise` or the `tapLookupResources` “resource not found” case (at least not with those names).​

​from what I can tell, this looks like a lifecycle edge case (double invocation, mount and unmount timing) where assistant-ui ends up calling an internal setter with an undefined receiver, so defensive guards to prevent the crash and keep the app usable.

​and i’ll be the first to admit that it’s possible some part of runtime code may not be exactly how assistant-ui expects (i haven’t found the mismatch yet), but the patch is currently the only thing that makes the flow stable end-to-end, on my side atleast.

> It would be great to move the AIAssistant into the Notebook on the right sidebar, where most user input activities, such as note-taking and interactive chat, take place. Meanwhile, the left sidebar could host chat history alongside the existing annotation logs, serving primarily as a read-only reference area.

yep, doing it.

> The project at https://github.com/olivernn/lunr.js hasn’t been updated for about six years, which suggests it may no longer be actively maintained. What advantages does building an inverted index over the entire document provide, compared to performing term searches at runtime with proper caching, as many modern code agents do today?

agreed, lunr.js hasn't been updated in years,i realised it after i was halfway done implementing. i chose it because it was the first result that came up when searching for js search libraries (open to change to newer and maintained libraries). and an inverted index typically gets $O(1)$ searches, and avoids the $O(n)$ cost of scanning book length strings every time.

also, when you mention caching term searches at runtime, do you mean doing on demand scans and then caching the matches? code agents often do this because they need substring matching if im not mistaken, which inverted indexes handle poorly. but for natural languages like in books, we need to know how rare/common words are across the whole book upfront to rank results properly. doing that at runtime would seemingly require a full scan per search, making the first hit slow. so it felt safer to pay the indexing cost once.

chrox:
> also, when you mention caching term searches at runtime, do you mean doing on demand scans and then caching the matches? code agents often do this because they need substring matching if im not mistaken, which inverted indexes handle poorly. but for natural languages like in books, we need to know how rare/common words are across the whole book upfront to rank results properly. doing that at runtime would seemingly require a full scan per search, making the first hit slow. so it felt safer to pay the indexing cost once.

That makes sense. We can just keep it for now and later migrate to more actively maintained libraries.

chrox:
> i couldn’t find an existing assistant-ui issue/PR that explicitly references `_getInitializePromise` or the `tapLookupResources` “resource not found” case (at least not with those names).​

Could you also try to check the 'Disable Cache' in the browsers network inspector and see if the problem still exists? There will be some occasional crash if some caches are not invalidated properly in development mode, which I suspect originates from turbopack as this did not appear until we switched to turbopack from webpack.

WhiteHades:
> > i couldn’t find an existing assistant-ui issue/PR that explicitly references `_getInitializePromise` or the `tapLookupResources` “resource not found” case (at least not with those names).​
> 
> Could you also try to check the 'Disable Cache' in the browsers network inspector and see if the problem still exists? There will be some occasional crash if some caches are not invalidated properly in development mode.

sure. yes turbopack is the one of the main culprits, especially that godforsaken HMR issue, i always get it.

chrox:
> sure. yes turbopack is the one of the main culprits, especially that godforsaken HMR issue, i always get it.

We will catch up with the latest `next.js` version which might contain some `turbopack` fixes once it's supported by `opennext`.

WhiteHades:
image

@chrox could you kindly tell me how to solve this `regex` issue stopping the build altogether. thank you.

chrox:
> @chrox could you kindly tell me how to solve this `regex` issue stopping the build altogether. thank you.

This usually indicates that the source code—possibly including imported third-party libraries—uses lookbehind regular expressions, which are only supported in very recent versions of iOS WebKit. Shipping the generated JavaScript bundles to older iOS devices will cause the app to crash immediately.

You may want to run the following command on your machine to locate the patterns:
```
grep -rnoE '\(\?<[!=]' .next/static/chunks/* out/_next/static/chunks/*
```

Then identify where the regex is used and rewrite it without lookbehind, as there is always an alternative way to express the same logic without relying on lookbehind support.

Cargo.lock
chrox, line 1: Please `git checkout` this as it's irrelevant to this PR.

chrox (changes requested):

WhiteHades:
> Could you also try to check the 'Disable Cache' in the browsers network inspector and see if the problem still exists? There will be some occasional crash if some caches are not invalidated properly in development mode, which I suspect originates from turbopack as this did not appear until we switched to turbopack from webpack.

can confirm toggling `Disable Cache` fixes the `HMR` issue, if i knew about this earlier, it would've saved me from depression lol, so, thanks a bunch.

WhiteHades:
> It would be great to move the AIAssistant into the Notebook on the right sidebar, where most user input activities, such as note-taking and interactive chat, take place

i have a question regarding this, do you mean integrate the entire `AIAssistant` side panel within the `Notebook` side panel, like, users have to click the `Notebook` icon in the top right, then the `Notebook` right panel opens and inside it, click another time the `AIAssistant` icon which changes the `Notebook` UI to the `AIAssistant` chat interface UI in the same said panel? isn't that counterintuitive, like one step too many to access the AI? or do you mean having the `AIAssistant` icon on the top right corner, beside the `Font & Layout` and `Notebook` icon, and clicking it would open the right side panel like it currently does for the left side panel? i think the latter is more convenient than the former no?

chrox:
What I mean is the first option, but without the extra friction you’re worried about. Clicking Notebook opens the right-side panel (same as today). Inside that panel, there’s a tab / mode switch at the bottom just like the left sidebar: Notes and AI Assistant. The Notebook panel can be pinned, same as the left sidebar.

So power users can: pin the Notebook panel, leave AI open permanently while reading and switch between Notes ↔ AI instantly

Casual users can open Notebook only when needed and use AI occasionally without UI clutter.

WhiteHades:
and regarding the history you mentioned in the left side panel, am I correct in understanding that said histories of the various chat the user had, is to be stored / contained in the place where the `AIAssitant` chat interface is currently in, so basically instead of the robot icon in the left side panel at the bottom, there is a history icon and clicking it opens up past conversations with the AI and users can edit the title of each conversation, delete them, and click them after which the `AIAssistant` opens up in the right side panel as clarified? if so, are there any other features that needs to be put here other than the ones i mentioned here?

chrox:
> and regarding the history you mentioned in the left side panel, am I correct in understanding that said histories of the various chat the user had, is to be stored / contained in the place where the AIAssitant chat interface is currently in, so basically instead of the robot icon in the left side panel at the bottom, there is a history icon and clicking it opens up past conversations with the AI and users can edit the title of each conversation, delete them, and click them after which the AIAssistant opens up in the right side panel as clarified? if so, are there any other features that needs to be put here other than the ones i mentioned here?

Yes, exactly. The robot icon in the left sidebar becomes a History icon. Clicking it opens a list of past AI conversations, where users can: 1. View chat history; 2. Rename conversation titles; 3. Delete conversations; 4. Select a conversation → which opens the AI Assistant UI in the right-side Notebook panel.

This keeps the UX consistent with the current product and avoids introducing new interaction patterns.

For this PR, that’s sufficient, no additional features are needed there. We can integrate deeper entry points (e.g. annotation toolbar → AI) in a follow-up PR.

chrox:
image

Can we revert the layout and style changes so we can focus on reviewing and testing functionality first? The background and sizing of some buttons have changed while others haven’t, which makes the UI feel inconsistent. I suspect this was introduced by the tooltip widget, and I’d suggest sticking with the native title tooltip to keep behavior and appearance consistent across all platforms.

This AI assistant feature will not be available to all users immediately before we refine the UI and UX in later PRs.

WhiteHades:
yes, i was just about to say this. reverting the tooltip.

chrox:
The `_getInitializePromise` error might be related to this: https://github.com/assistant-ui/assistant-ui/issues/3103

WhiteHades:
> The `_getInitializePromise` error might be related to this: [assistant-ui/assistant-ui#3103](https://github.com/assistant-ui/assistant-ui/issues/3103)

interesting. so atleast it's not an isolated incident. do you suggest to use a lower version for assistant-ui (as well as its markdown dependency) or keep using the current hack-y fix?

chrox:
Sure, manual patching for third party libraries is hard to maintain, we should be able to remove the `assistant-ui` patch with this, and the lookbehind assert patches will be removed after we target the iOS version to 16.4+ which I think will be soon:

```
-    "@assistant-ui/react": "^0.11.58",
-    "@assistant-ui/react-ai-sdk": "^1.2.0",
-    "@assistant-ui/react-markdown": "^0.11.10",
+    "@assistant-ui/react": "0.11.56",
+    "@assistant-ui/react-ai-sdk": "1.1.21",
+    "@assistant-ui/react-markdown": "0.11.9",
```

chrox:
We will merge this for now as it's already huge. Before we polish and release it, the features are only available in the development mode.

WhiteHades:
great, thanks for the merge, and also, i was working on some UX improvements though, what should i do with it?

chrox:
It's better to checkout a clean branch from the origin main and start your new PR from there, as your local branch has already drift away from the origin and there will be more merge conflicts in the next PR if you do not do that.

PR #3075, opened 2026-01-25
feat: auto position reader ruler to top of page and some fixes

WhiteHades:
this pr includes:
- auto position to the first visible line of text when changing pages
- improved persistence position saving
- fixed the issues of the reader ruler appearing before all book reading contents when any book is opened and during close, it stays for a bit longer compared to the other components.

https://github.com/user-attachments/assets/57e552c5-4a16-4850-a78b-89cd8860bd37

### DIFF
--- a/apps/readest-app/src/app/reader/components/ReadingRuler.tsx
+++ b/apps/readest-app/src/app/reader/components/ReadingRuler.tsx
@@ -5,8 +5,8 @@ import { BookFormat, FIXED_LAYOUT_FORMATS, ViewSettings } from '@/types/book';
 import { useEnv } from '@/context/EnvContext';
 import { useReaderStore } from '@/store/readerStore';
 import { saveViewSettings } from '@/helpers/settings';
-import { eventDispatcher } from '@/utils/event';
 import { READING_RULER_COLORS } from '@/services/constants';
+import { throttle } from '@/utils/throttle';
 
 interface ReadingRulerProps {
   bookKey: string;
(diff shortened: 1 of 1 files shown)


chrox:
I’ve reimplemented the ruler area using a backdrop filter in https://github.com/readest/readest/pull/3076. This approach preserves the original luminance while reducing background contrast, while maintaining sufficient contrast within the ruler area so readability isn’t affected. This should be beneficial for accessibility, especially for users with ADHD who may rely more on this feature.

Could you please rebase this PR onto #3076? Sorry about that.

WhiteHades:
sure, can do that.

WhiteHades:
@chrox 

https://github.com/user-attachments/assets/c0f8b7b6-2ea2-4477-b54d-7837ef7eae6a

apps/readest-app/src/app/reader/components/ReadingRuler.tsx
```diff
@@ -45,19 +49,27 @@ const ReadingRuler: React.FC<ReadingRulerProps> = ({
...
+  // States for visibility animation (fade in/out)
+  const [isVisible, setIsVisible] = useState(false);
+  const [isClosing, setIsClosing] = useState(false);
+
+  // State for smooth auto-position animation
+  const [shouldAnimate, setShouldAnimate] = useState(false);
+
   const isDragging = useRef(false);
+  const lastPageRef = useRef<number | null>(null);
+  const animationTimeoutRef = useRef<NodeJS.Timeout | null>(null);
```
chrox: Please use `useRef<ReturnType<typeof setTimeout> | null>(null)`

apps/readest-app/src/app/reader/components/ReadingRuler.tsx
```diff
@@ -78,10 +90,166 @@ const ReadingRuler: React.FC<ReadingRulerProps> = ({
...
+        envConfig,
+        bookKey,
+        'readingRulerPosition',
+        currentPositionRef.current,
+        false,
+        false,
+      );
+      setIsClosing(true);
+    };
+    eventDispatcher.on('reader-closing', handleClosing);
```
chrox: We don't need another custom event for this. Please refer to the `saveBookConfig` in `apps/readest-app/src/app/reader/hooks/useProgressAutoSave.ts` to throttle the `saveViewSettings` and trigger it in the `handleRelocate`.

apps/readest-app/src/app/reader/components/ReadingRuler.tsx
```diff
@@ -78,10 +90,166 @@ const ReadingRuler: React.FC<ReadingRulerProps> = ({
...
+      const currentPage = view.renderer.page;
+
+      // Only auto-move if page actually changed (not on initial load)
+      if (lastPageRef.current !== null && lastPageRef.current !== currentPage) {
+        requestAnimationFrame(() => performAutoMove(range as Range | null));
+      }
+      lastPageRef.current = currentPage;
+    };
+
+    view.renderer.addEventListener('relocate', handleRelocate);
```
chrox: You can use `const progress = getProgress(bookKey);` and react to changes in progress with a `useEffect`. This encapsulates the details of the `relocate` event and should be treated as the single source of truth.

apps/readest-app/src/helpers/settings.ts
```diff
@@ -29,9 +29,9 @@ export const saveViewSettings = async <K extends keyof ViewSettings>(
         const view = getView(bookKey);
         view?.renderer.setStyles?.(getStyles(viewSettings));
       }
-      const config = getConfig(bookKey);
-      if (viewState?.isPrimary && config) {
-        await saveConfig(envConfig, bookKey, config, settings);
+      const updatedConfig = useBookDataStore.getState().getConfig(bookKey);
+      if (viewState?.isPrimary && updatedConfig) {
+        await saveConfig(envConfig, bookKey, updatedConfig, settings);
```
chrox: Changes in this file should be identical to the previous version. Did you encounter any unexpected behavior?

WhiteHades: there was a race condition while dragging the ruler where stale configuration was being saved (in my testing at least), which is why the `isPrimary` check and the latest `config`

PR #3096, opened 2026-01-28
feat:  paragraph by paragraph reading mode

### LINKED ISSUE #1844: FR: Paragraph-by-paragraph reading mode (similar to Readwok)

> Sometimes, reading takes a lot of brain power and concentration, so you tend to get tired. You zone out and get lost in your mind, inevitably losing track of where you were reading.

> With paragraph-by-paragraph reading, it's easy to focus your attention and track last reading position. It's a more intuitive approach where instead of being bombarded with a wall of text, you consume it in chunks, fostering anticipation and enjoyment.

**Additional context**

Source: Readwok.com

Image

https://github.com/user-attachments/assets/6d134a16-b045-42a3-8d9f-b5c87236b44b

WhiteHades:
closes #1844.

this pr adds:
- paragraph by paragraph reading mode
- can control using the arrow keys or the mouse
- preserves sync progress and auto goes to the current position in the normal reading view when exiting para mode.

https://github.com/user-attachments/assets/7bfc75cc-80f2-44d8-99d8-4978bd3f6577

### DIFF
--- a/apps/readest-app/src/app/reader/components/paragraph/ParagraphOverlay.tsx
+++ b/apps/readest-app/src/app/reader/components/paragraph/ParagraphOverlay.tsx
@@ -34,7 +34,6 @@ const AnimatedParagraph: React.FC<{
   const showContent = state === 'active' && isReady;
 
   return (
-    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
     <div
       ref={contentRef}
       className={clsx(
@@ -359,6 +358,7 @@ const ParagraphOverlay: React.FC<ParagraphOverlayProps> = ({
   const exitingParagraph = paragraphs.find((p) => p.state === 'exiting');
 
   return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
     <div
       ref={containerRef}
       role='dialog'


chrox:
I noticed an issue with paragraph mode where the reading position doesn’t sync with the normal page mode. The progress in paragraph mode seems to be locked to the initial value from the first time I enabled it. No matter how much progress is made in paragraph, page, or scroll mode, switching back to paragraph mode always jumps back to 7%.

Testing EPUB file: 
[The Age of Reason Begins.epub.zip](https://github.com/user-attachments/files/24927825/The.Age.of.Reason.Begins.epub.zip)

chrox:
It should be possible to keep only one progress with the CFI location for paragraph, page, and scroll mode, so that switching between these modes could be seamless.

WhiteHades:
> I noticed an issue with paragraph mode where the reading position doesn’t sync with the normal page mode. The progress in paragraph mode seems to be locked to the initial value from the first time I enabled it. No matter how much progress is made in paragraph, page, or scroll mode, switching back to paragraph mode always jumps back to 7%.

yes, i am aware of this. so, the current setup is that, it **does** save progress, in the sense that when you move through the paragraph, in the background you can see the page change if the new paragraph is in a different page. so this confirms that the progress is saved in _this_ specific regard.

however, the issue is, as you rightly pointed out, that is not persisted when you reopen the para mode. But, there is an important caveat here as explained below.

there are two types of user stories here:
1. the user opens para mode, reads through a couple of paragraphs and when they exit to be in the normal reading mode, see the page which contains the paragraph they last read in (said para mode) to be present in the current page (be it single or double coloumn). Afterwards, they start reading normally in the normal reading mode, and make progress. now, when they open the para mode again, they would expect it to start from the paragraph which they last read _from the current page_, but currently, it is not doing so because it is impossible to determine what paragraph they exactly read in the current page (be it single or double coloumn) for obvious reasons. so, i was thinking for the fix, we can have the para mode start from the first para in the current page, no matter if it's single or double coloumn (given they close the para mode, read a couple of pages normally and then reopen para mode).
2. the user opens para mode, reads through a couple of paragraphs and when they exit to be in the normal reading mode, see the page which contains the paragraph they last read in (said para mode) to be present in the current page (be it single or double coloumn). Afterwards, they open the para mode again, without reading any more pages in the normal mode, meaning they're in the exact same location as they were when they last closed the para mode. in this instance, they would expect it to start from the last paragraph they closed, currently this is not being done.

these are the two user stores i can think of that covers all aspects and needs a fix, i will do them ofcourse. are they exhaustive or did i miss any other scenarios?

@chrox

chrox:
It’s reasonable to rewind to the first paragraph on the page when switching from page mode to paragraph mode. It’s also acceptable that when switching from paragraph mode (paragraph 1) to page mode and then back to paragraph mode (paragraph 2), the two paragraphs are not identical, as long as a given page in page mode maps to exactly one paragraph in paragraph mode. This paragraph could be either the first paragraph on the page or the paragraph corresponding to the CFI location; I would prefer the CFI-based paragraph, as it may be cleaner to implement.

This behavior is similar to switching between paginated mode and scroll mode. When switching from scroll mode (location 1) to paginated mode and then back to scroll mode (location 2), location 1 and location 2 are usually not the same. However, because each page in paginated mode corresponds to a single location in scroll mode, repeated switching between scroll and paginated modes remains stable and the resulting location changes are expected.

Issue #2637, opened 2025-12-07
0.9.94 isn't building


svenstaro:
Hello, I'm the maintainer for the readest package in Arch Linux. I can't build version 0.9.94. 0.9.93 builds just fine, though.

I build it like this:
```sh
git submodule update --init --recursive
pnpm install
pnpm --filter @readest/readest-app setup-pdfjs
# Disable updater artifacts in tauri.conf.json
# Otherwise it will try to sign the built package with developer's GPG key which is unnecessary,
# and then it causes build failure because the GPG key is not available in the build environment.
# Also the updater artifacts are not needed for the AUR package.
# For more details, see https://github.com/tauri-apps/tauri/issues/13259
sed -i 's/"createUpdaterArtifacts": true/"createUpdaterArtifacts": false/' \
  apps/$pkgname-app/src-tauri/tauri.conf.json
pnpm tauri build -b deb
```

```
Submodule 'packages/foliate-js' (https://github.com/readest/foliate-js.git) registered for path 'packages/foliate-js'
Submodule 'packages/simplecc-wasm' (https://github.com/readest/simplecc-wasm.git) registered for path 'packages/simplecc-wasm'
Submodule 'packages/tauri' (https://github.com/readest/tauri.git) registered for path 'packages/tauri'
Submodule 'packages/tauri-plugins' (https://github.com/readest/tauri-plugins-workspace.git) registered for path 'packages/tauri-plugins'
Cloning into '/build/readest/src/readest/packages/foliate-js'...
Cloning into '/build/readest/src/readest/packages/simplecc-wasm'...
Cloning into '/build/readest/src/readest/packages/tauri'...
Cloning into '/build/readest/src/readest/packages/tauri-plugins'...
Submodule path 'packages/foliate-js': checked out 'a6b6252d7c1fa2c30556f1ba2c69338ab1972b62'
Submodule path 'packages/simplecc-wasm': checked out '3e4689325f769da999a407b07bc705b1f91c5622'
Submodule 'OpenCC' (https://github.com/BYVoid/OpenCC.git) registered for path 'packages/simplecc-wasm/OpenCC'
Cloning into '/build/readest/src/readest/packages/simplecc-wasm/OpenCC'...
Submodule path 'packages/simplecc-wasm/OpenCC': checked out 'e0d41d7f5e7c62f9be1cfdd9e6cb1d03847c06e7'
[...]

PR #3139, opened 2026-02-02
Change setup command for vendors in CONTRIBUTING.md to reflect additional vendor files

charrondev:
I ran into the same issue as https://github.com/readest/readest/issues/2637

It seems the CONTRIBUTING.md wasn't updated to note additional dependencies besides `pdf-js` needed to be setup.


chrox:
Good catch. Thank you.

PR #3150, opened 2026-02-03
fix(rsvp): fix position restoration and keyboard handling

boludo00:
## Summary
- Fix bug where exiting RSVP after changing speed navigates to wrong position (several pages backwards)
- Fix arrow keys inadvertently changing chapter dropdown instead of controlling RSVP speed

### Position restoration fix
When RSVP starts, it creates DOM `Range` objects tied to that moment's document state. If the document re-renders during the session, these Range objects become stale. The fix:
- Added `docWordIndex` to track word position within each document
- Validate Range before using, recreate from current DOM if stale
- Use per-document index instead of global index for accurate position recovery

### Keyboard handling fix
Arrow keys during RSVP mode were being captured by the chapter dropdown. Fixed by:
- Using capture phase for keyboard events
- Stopping event propagation to prevent native elements from receiving events

## Test plan
- [ ] Start RSVP, read for a bit, change speed, pause, exit - should return to correct position
- [ ] During RSVP, press arrow keys - should change speed, not navigate chapter dropdown

### DIFF
--- a/apps/readest-app/src/services/rsvp/types.ts
+++ b/apps/readest-app/src/services/rsvp/types.ts
@@ -21,7 +21,6 @@ export interface RsvpState {
 export interface RsvpPosition {
   cfi: string;
   wordText: string;
-  wordIndex?: number; // Legacy field for backward compatibility
 }
 
 export interface RsvpStopPosition {


apps/readest-app/src/app/reader/components/rsvp/RSVPControl.tsx
```diff
@@ -16,6 +16,109 @@ interface RSVPControlProps {
...
 }
 
+// Helper to find a word in document and create a fresh Range
+const findWordInDocument = (
+  doc: Document,
+  wordText: string,
+  docWordIndex: number | undefined,
+  docTotalWords: number | undefined,
+): Range | null => {
+  // Walk through text nodes to find the word (same logic as RSVPController.extractWordsFromElement)
```
chrox: Please use CFI to locate the word which can be generated from document range and document index and can be persistent to configuration or localstorage.

Try to reuse the code if possible and support CJK words.

apps/readest-app/src/app/reader/components/rsvp/RSVPOverlay.tsx
```diff
@@ -299,34 +304,33 @@ const RSVPOverlay: React.FC<RSVPOverlayProps> = ({
       onTouchEnd={handleTouchEnd}
     >
       {/* Header */}
-      <div className='rsvp-header flex shrink-0 items-center justify-between gap-2 p-3 md:gap-4 md:p-4'>
+      <div className='rsvp-header flex shrink-0 items-center justify-between p-4'>
         <button
-          aria-label={_('Close RSVP')}
-          className='flex h-10 w-10 shrink-0 cursor-pointer items-center justify-center rounded-full border-none bg-transparent transition-colors hover:bg-gray-500/20 md:h-11 md:w-11'
+          aria-label='Close RSVP'
```
chrox: Labels and titles should be translated with `_`, please revert this. It's better to rebase with the upstream main branch before you make any new PRs.

apps/readest-app/src/services/rsvp/types.ts
```diff
@@ -29,6 +30,8 @@ export interface RsvpStopPosition {
   text: string;
   range?: Range;
   docIndex?: number;
+  docWordIndex?: number; // Index within the specific document (for position recovery)
+  docTotalWords?: number; // Total words in the specific document
 }
```
chrox, line 33: Use `CFI` instead of word index for better protability.

apps/readest-app/src/app/reader/components/rsvp/RSVPOverlay.tsx
boludo00: Sorry about this! I will keep that in mind for future PRs.

chrox:
RSVP mode was temporarily disabled in #3164 in case a new version is released before it gets fixed. Please re-enable it in this PR.

boludo00:
Thanks for the review feedback, I've done my best to address all the concerns
                                                                                                                                                                                                                                                    
  1. ✅ CFI-based positioning - Switched from word index to CFI for position tracking (commit 610dac9c)                                                                                                                                             
  2. ✅ i18n translations - Wrapped all RSVP overlay strings with _() (commit 4b577b4d)                                                                                                                                                             
  3. ✅ Re-enabled RSVP mode - Speed Reading Mode is now re-enabled in production (commit 48408b8e)

  Ready for another look when you have time.

apps/readest-app/src/services/rsvp/types.ts
```diff
@@ -19,8 +20,8 @@ export interface RsvpState {
 
 export interface RsvpPosition {
   cfi: string;
-  wordIndex: number;
   wordText: string;
+  wordIndex?: number; // Legacy field for backward compatibility
```
chrox: Please remove this. There is no need to keep the backward compatibility as it's never in the production.

apps/readest-app/src/services/rsvp/RSVPController.ts
```diff
@@ -132,37 +137,176 @@ export class RSVPController extends EventTarget {
...
+    }
+
+    return { path: basePath, charOffset };
+  }
+
+  /**
+   * Compare two CFI document paths to determine order.
+   * Returns: negative if path1 < path2, positive if path1 > path2, 0 if equal
+   */
+  private compareCfiPaths(path1: string, path2: string): number {
```
chrox: Use `CFI.compare` in `foliate-js/epubcfi.js`. See `apps/readest-app/src/utils/cfi.ts`.

apps/readest-app/src/services/rsvp/RSVPController.ts
```diff
@@ -132,37 +137,176 @@ export class RSVPController extends EventTarget {
...
+
+    // If all compared steps are equal, shorter path comes first
+    return steps1.length - steps2.length;
+  }
+
+  /**
+   * Compare two CFI positions including character offsets.
+   * Returns: negative if pos1 < pos2, positive if pos1 > pos2, 0 if equal
+   */
+  private compareCfiPositions(
```
chrox: Use `CFI.compare`.

apps/readest-app/src/services/rsvp/RSVPController.ts
```diff
@@ -132,37 +137,176 @@ export class RSVPController extends EventTarget {
...
+    localStorage.removeItem(`${POSITION_KEY_PREFIX}${this.bookId}`);
   }
 
-  private extractBaseCfi(cfi: string): string {
+  /**
+   * Extract the spine/section reference from a CFI.
+   * For `epubcfi(/6/6!/4/2/2/6,/1:363,/1:366)`, returns `/6/6!`
+   * This identifies which section/chapter the CFI points to.
+   */
+  private extractSpineRef(cfi: string): string {
```
chrox: Use `fake.toIndex` to parse spine index. See `readest/apps/readest-app/src/utils/xcfi.ts`.

boludo00: using `XCFI.extractSpineIndex` now.

chrox:
Thanks for the fix.

PR #3320, opened 2026-02-19
fix(txt): stabilize iOS large TXT import and decode picker paths

royisme:
## Summary
- add missing TXT worker files used by large import flow
- fix iOS file picker path decoding for imported files with encoded names
- stabilize large TXT conversion path by avoiding fragile stream/handle lifecycle on iOS
- ## Validation
- pnpm -C apps/readest-app build


chrox:
Please verify if https://github.com/readest/readest/pull/3302 has already fixed the problem of importing txt files on iOS. If there are still problems when importing large txt files please also share a sample file and the system information for others to reproduce the issue.

chrox:
Looks great. Thanks.

royisme:
thanks! this is my first pr in this year ! lol

chrox:
Welcome to the Readest community!

PR #3361, opened 2026-02-24
fix: empty paragraphs not skipped in paragraph mode

### LINKED ISSUE #3329: In Paragraph Mode, empty paragraphs aren't skipped
## Bug Description

Empty paragraphs aren't skipped in paragraph mode. Probably due to weird formatting, readest sees an empty paragraph after every real paragraph in my book, thus I always have to double tab the next paragraph button.

Image

## Expected behavior

In my opinion it would be reasonable to exclude empty paragraphs (or such that only include whitespaces)

## Technical inputs

Operating System:
Arch Linux 

Readest Version:
Version 0.9.100 (WebView 605.1.15)

WhiteHades:
closes #3329.

this pr adds:
- skip whitespace only paragraphs in paragraph mode
- add tests for empty and media only paragraphs

https://github.com/user-attachments/assets/c7e2e5d6-9662-4c45-b89d-90b56ad339ca

### DIFF
--- a/apps/readest-app/src/utils/style.ts
+++ b/apps/readest-app/src/utils/style.ts
@@ -867,9 +867,15 @@ export const applyTableStyle = (document: Document) => {
         table.style.width = `calc(min(${computedWidth}, var(--available-width)))`;
       }
     }
+    const parentWidth = window.getComputedStyle(parent as Element).width;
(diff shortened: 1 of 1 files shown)


dryBoneMarrow:
Thank you very much :)

Issue #3343, opened 2026-02-22
FR: Rename `pages left in chapter` to `screens left in chapter` following the #3231 commit


Blyrium:
Since #3231, the "remaining pages" displayed at the bottom of the screen are not "pages" but "physical left pages", meaning the number of screens a user must scroll through on their device before reaching the end of a chapter.

### The problem

It is confusing to read "25 pages left in chapter" when the actual chapter is only 15 pages long.

### Proposed change

All occurrences of `{{count}} pages left in chapter_one`, `{{count}} pages left in chapter_many`, and `{{count}} pages left in chapter_other` should be renamed to `{{count}} screens left in chapter_one`, `{{count}} screens left in chapter_many`, and `{{count}} screens left in chapter_other.`

### Why

The displayed count reflects how many screens remain to be scrolled on the current device, not the number of actual pages in the chapter. Replacing "pages" with "screens" avoids confusion.

chrox:
Thank you for bringing this up.

It may be clearer to explain the distinction in the wiki or official documentation instead of introducing a new term like “screens.” Adding new terminology could create more confusion, since “pages” is already a familiar concept for most readers.

Blyrium:
> It may be clearer to explain the distinction in the wiki or official documentation instead of introducing a new term like "screens."

Documentation would be a start, but I wouldn't expect users to look it up before wondering why their chapter has more "pages" than it actually does.

> since "pages" is already a familiar concept for most readers

It is, which is precisely why displaying two different page counts is confusing. If we're calling them pages, users will naturally compare that number to the chapter's actual page count.

> instead of introducing a new term like "screens."

I'm not a native English speaker, so I'm open to better wording, but "screens" seems fairly self-explanatory: it's what you see in front of you. Alternatively, if the terminology is a concern, the cleaner fix might be to display the actual remaining page count instead, so no new term is needed and the label stays accurate.

Either way I think the current state, where the label says "pages" but the number doesn't reflect pages, is the most confusing outcome.

As a workaround for users like me, I'd appreciate it if the label text inside `.text-start` was wrapped in its own element, so `<span class="text-start">8 pages left in chapter</span>` becomes `<span class="text-start">8<span class="pages-left-label"> pages left in chapter</span></span>`. This would let me target and replace the label text with CSS, without touching any app logic.

chrox:
[...]

PR #3368, opened 2026-02-24
Allow CSS targeting of the "NUMBER pages left in chapter" label via `<Trans>` component

Blyrium:
Splits the "NUMBER pages left in chapter" string into two independently targetable elements by replacing the `_()` call with a `<Trans>` component in `ProgressInfo.tsx`:

- `<span class="pages-left-number">` wraps the dynamic page count
- `<span class="pages-left-label">` wraps the label text

All locale files have been updated accordingly to use `<Trans>` component index placeholders (`<0>` for the number, `<1>` for the label), preserving correct word order across all supported languages including Japanese, where the number appears mid-sentence.

This allows users to override the label text via custom CSS, for example to replace "pages" with "screens" as discussed in #3343, without touching any app logic.

### Number comes first (e.g. "8 pages left in chapter")
```css
.pages-left-label { display: none; }
.pages-left-number::after { content: ' whatever you want after the number'; }
```

### Label wraps around the number (e.g. Japanese "章に8ページ残り")
```css
.pages-left-label { display: none; }
.pages-left-number::before { content: 'whatever you want before the number '; }
.pages-left-number::after { content: ' whatever you want after the number'; }
```

### DIFF
--- a/apps/readest-app/src/app/reader/components/ProgressInfo.tsx
+++ b/apps/readest-app/src/app/reader/components/ProgressInfo.tsx
@@ -171,33 +171,30 @@ const ProgressInfoView: React.FC<PageInfoProps> = ({
           isVertical ? 'h-full' : 'h-[52px] w-full',
         )}
       >
-      {(progressInfoMode === 'all' || progressInfoMode === 'remaining') && (
-        <>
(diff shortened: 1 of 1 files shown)


chrox:
Looks great. Thanks.

Issue #3334, opened 2026-02-21
Reader font not applied when epub defines `font-family: serif` on body


Blyrium:
## Bug Description

When an epub's stylesheet defines `body { font-family: serif; }`, Readest resolves `serif` to "Ascender Serif WGL Regular" (apparently), bypassing the app's configured font. In Readest settings, I have Bitter as the default serif font. Despite this, "Ascender Serif WGL Regular" is used instead of Bitter whenever the epub declares `body { font-family: serif; }`.

This can be confirmed in DevTools (using web.readest.com): removing the `body { font-family: serif; }` rule from the epub's stylesheet causes Readest to correctly apply Bitter. Enabling "Overwrite Book Font" also restores Bitter, but should not be necessary in this case.

<img width="1271" height="264" alt="Image" src="https://github.com/user-attachments/assets/c2f252e9-56a5-4a02-a4a3-96e8f9ff26f0" />

_With `body { font-family: serif; }`_

<img width="1271" height="264" alt="Image" src="https://github.com/user-attachments/assets/2f64a5b8-bf54-4b93-9fb1-572fdbabf967" />

_Without `body { font-family: serif; }`_

## Expected Behavior

CSS generic font family keywords (`serif`, `sans-serif`, `monospace`) are not specific font requests; they are fallback hints. When an epub uses `body { font-family: serif; }` without embedding or naming a specific font, Readest should apply the user's configured reader font for that family (in this case, Bitter) by default. The "Overwrite Book Font" toggle should be reserved for epubs that explicitly name a specific font.

## Technical Inputs

Operating System: Kubuntu 24.04  
Readest Version: 0.9.100 (AppImage)

PR #3394, opened 2026-02-26
Fix generic font family keywords bypassing user font settings

Blyrium:
## Summary

When any element other than `body` declares a generic font family keyword, whether as a sole value (`font-family: serif`) or as a trailing fallback in a stack (`font-family: "Times New Roman", "Utopia", serif`), Readest was leaving the keyword untouched. The browser or WebView would then resolve it to a system font, bypassing the user's configured font entirely.

Commit 79ba9b3 partially addressed this for `body`: when an epub's stylesheet sets `font-family: serif` or `font-family: sans-serif` on `body`, it unsets the property so Readest can honour the user's "Default Font" setting, which lets the user choose between their configured serif and sans-serif font. This is a good behaviour and this PR preserves it. However, it does not fully solve #3334, as the fix only applies to `body` (but that's on me for not being accurate enough in my bug report).

This already worked correctly for `monospace` (via an existing replacement in `transformStylesheet`), but only when `monospace` was the sole value. A stack like `font-family: "Courier", "Cousine", monospace` was not handled.

## The fix

In `transformStylesheet` (`style.ts`), we replace the existing `monospace`-only entry with four replacements covering all three generic keywords:

```ts
.replace(/(font-family\s*:[^;]*?)\bsans-serif\b/gi, '$1READEST_SS_PLACEHOLDER')
.replace(/(font-family\s*:[^;]*?)\bserif\b(?!-)/gi, '$1var(--serif)')
.replace(/READEST_SS_PLACEHOLDER/g, 'var(--sans-serif)')
.replace(/(font-family\s*:[^;]*?)\bmonospace\b/gi, '$1var(--monospace)')
```

`sans-serif` is temporarily replaced with a placeholder before processing `serif`, preventing the `serif` pattern from matching inside `var(--sans-serif)`. The placeholder is then converted to `var(--sans-serif)`. The lookahead `(?!-)` on the `serif` pattern guards against matching `serif` in hypothetical font names like `serif-something`.

The CSS variables `--serif`, `--sans-serif`, and `--monospace` are already defined by Readest and always list the user's configured font first, so this ensures the user's font preferences are respected in all cases.

## Bonus fix

The `body`-specific block from commit 79ba9b3 used `[;$]` intending to match a semicolon or end-of-string, but inside a character class `$` is a literal dollar sign, not an anchor. This meant a bare `font-family: sans-serif` on `body` without a trailing semicolon, which is valid CSS, was silently skipped and not unset. The fix replaces `[;$]` with `(?:;|\}|$)`.

## Test epub

A test epub is attached right here [Test_epub.zip](https://github.com/user-attachments/files/25585399/Test_epub.zip). It contains several `div` elements, each testing a specific `font-family` declaration, with inline descriptions of the expected behavior. It covers all six cases: bare and stacked `serif`, `sans-serif`, and `monospace`.


Blyrium:
I'll fix this in a minute.

PR #3402, opened 2026-02-28
feat(reader): adding current Time and Battery to Footer (#3306)

### LINKED ISSUE #3306: FR: Displaying time and maybe battery in footer
I always liked the footer displaying useful information (time, battery) with other apps such as moon reader, apple books,... and miss it on here and I think its a nice optional (toggle via settings) extra to have

I think a toggle in settings to have such information would be nice, it should display the current time but current battery on phones or current chapter progress in percentage would be nice

an example would be from moon reader:

Image

scinac:
close #3306 

#### what this pr adds:
- Current Time and Battery in Footer
- Settings to toggle current Time and Battery being shown
- Settings to toggle a 24 Hour Clock instead of 12 Hour

#### showcase 
image

#### notes
The battery format might be confusing considering the whole book progress uses the same % format


ybbluesky:
good

PR #3485, opened 2026-03-06
feat(database): add platform-agnostic schema migration system

chrox:
Add a unified migration runner that works across all three database backends (native/node/wasm) via the DatabaseService interface. Uses PRAGMA user_version as an O(1) fast-path to skip checks on already- current databases, making it safe to call on every open even with hundreds of db files. Schema types allow different migration sets for different database kinds.

Also adds test:all and test:pr:tauri scripts with automated Tauri app lifecycle management, and silences noisy tracing/tantivy logs.


Hermotimus:
Since adding the “turso” brick to the project (commit 5273ef7), I have been unable to perform any builds for Android.

There is a systematic Rust compiler error on two files in `apps/readest-app/src-tauri/plugins/tauri-plugin-turso/src`

Attached is a complete log of the build on the latest commit (54bc151).

Could you, please, help me resolve this issue?

[20260306-build-log.html](https://github.com/user-attachments/files/25797941/20260306-build-log.html)

chrox:
We will fix it shortly. Thanks.

chrox:
Please check out the latest main branch, it should fix the android build. You also need to update the `tauri-plugin-turso` submodule.

Hermotimus:
> Please check out the latest main branch, it should fix the android build. You also need to update the `tauri-plugin-turso` submodule.

Great fix! Thank you.

Sidenote: _The binary takes considerably longer to build…_

chrox:
Partly because we bumped tauri version so it invalidated some rust caches.

PR #3527, opened 2026-03-13
Update docker db schema to match FileRecords

azeemba:
In #2636, FileRecord was defined to have updated_at field which is expected to exist in the database. But the local dev setup is missing this field.

This diff adds an updated_at column to match the expectation


chrox:
Good catch. Thanks.

PR #3546, opened 2026-03-16
feat(epub): support continuous scroll and spread layout for EPUBs, closes #3419 and closes #1745

### LINKED ISSUE #1745: FR: Continuous Scrolling in EPUBs and PDFs
Hi, thanks for the great work on this project!

I know there’s already a scrolling mode available, but currently it seems to briefly freeze when switching chapters, rather than allowing a smooth continuous scroll into the next chapter (like in WeChat Reading). This becomes a problem when reading near the end of a chapter — if I haven’t finished the last few lines, it jumps to the next chapter too soon, disrupting the reading flow.

I tried modifying the code to implement a true continuous scrolling experience. From what I understand, foliate.js currently only supports fixed-layout and paginator views, and doesn't provide a native continuous scrolling view. So I created a basic demo myself to test the concept, and recorded a short mp4 video to show how it currently works.

[continuous scrolling](https://github.com/user-attachments/assets/6c68a41f-9c91-4b33-995b-403766f8ef6a)

There are still some issues — for example, I haven't implemented syncing the scroll position with the table of contents yet. I also removed the use of iframes and instead injected all chapters using divs, which causes some style problems and breaks things like the select menu. You can also see in the video that chapter order glitches slightly at the beginning.

The code is here: (txt -> js)

[continuous-paginator.txt](https://github.com/user-attachments/files/21580191/continuous-paginator.txt)

I'm not sure if any of this will be helpful, but I'd really love to see a proper continuous scrolling mode like in WeChat Reading. Let me know what you think!

### LINKED ISSUE #3419: FR: dual-page mode for EPUB files for two sections
Hi Xin,

I’ve already mentioned this in a larger feature request thread (issue #996), but I wanted to make a dedicated request for this.

### The Problem

When using Book View for PDF files (like in SumatraPDF), the layout looks correct. However, in Readest, EPUB files don't handle two-page spreads properly.

* **Sumatra PDF (How it should look):** 

https://github.com/user-attachments/assets/124f0cc7-9909-4455-9c83-2f4a1eeb5280

* **Readest (Current behavior):** 

https://github.com/user-attachments/assets/9b13236c-7a9e-41da-a052-2164b02af1ad

Right now, if an image is meant to be a two-page spread, Readest often forces it onto a single page, or separates the halves so that the image feels incomplete. In some files, Readest shows two images (each taking up a page) in dual-page mode, but in many others, it only shows one half of an image in the entire dual-page view.

### The Solution I'm Looking For

I’m looking for a way to override how images are placed so the dual-page view works more like a physical book. Specifically:

1. **Spread Override:** If an image is intended to be a two-page spread, I want an option to override the positioning so that the two parts of the image can be placed side-by-side to complete the picture, rather than being split by a page break.
2. **Cover Page Logic:** In book mode, the first page should be a single page (the cover), and the rest of the pages should be in dual-page mode, just like a real book.

Basically, I want the ability to "force" the layout so that the dual-page mode doesn't break images that were meant to be viewed together.

chrox:


Hermotimus:
Hello,

Since this exact commit, TTS playback has not been working properly.
In Scrolled mode, the `Back to TTS location` button keeps reappearing every time playback advances. If you click this button, the issue returns as soon as the text advances.

chrox:
Thanks, I will take a look.

PR #3579, opened 2026-03-21
feat(koplugin): add support for annotation sync

chrox:
Add bidirectional annotation/highlight sync between KOReader and Readest:

- Add xpointer0/xpointer1 fields to BookNote and DBBookNote types for KOReader XPointer positions alongside Readest's CFI format
- Extend transform layer to pass through xpointer fields to/from DB
- Convert CFI→XPointer on push and XPointer→CFI on pull in useNotesSync, discarding notes that fail conversion
- Support KOReader's text()[K].N indexed text node format in xcfi.ts for paragraphs with inline elements (e.g. <a> page anchors)
- Generate KOReader-compatible XPointers: text().N for single text nodes, text()[K].N only when multiple direct text nodes exist
- Skip cfi-inert elements (injected by Readest at runtime) in XPointer path building and resolution
- Map highlight colors between KOReader and Readest color systems
- Implement KOReader plugin annotation push/pull with deterministic IDs, auto-sync on document open/close, and UIManager refresh on pull
- Refactor koplugin into focused modules: syncauth, syncconfig, syncannotations, selfupdate


mallomar:
I'm very happy to see this feature being added! This is another step in being able to use KOReader or Readest interchangeably, which makes me more likely to pick up Readest. You may also consider posting about this in the KOReader subbredit, as I imagine it'd bring this project more readers, though I am happy to do so if you prefer.

@chrox hoping vocabularly builder sync and reading statistics syncs are also in the pipeline for the not too distant future, though I recognize both are dependent on Readest rolling them out first.

chrox:
We'd love to bring them into the sync pipeline too. I also hope it won’t be long.

PR #3668, opened 2026-03-28
feat(scroll): add single section scroll option, closes #3663

### LINKED ISSUE #3663: Scrolled Mode
Is the Scrolled Mode no longer locked by chapters? I liked how it was before; now it's kind of like all the chapters are on the same page. You can see the end of the last chapter and the start of the next one, which I didn't like. 

Image

chrox:
## Summary
- Add "Single Section Scroll" toggle in Settings > Scroll, enabled only in scrolled mode
- When enabled, the paginator renders only the current section — no preloading or adjacent section loading
- Guards added at all multi-section entry points in `paginator.js` (scroll handlers, `#preloadNext`, `#fillVisibleArea`, `#display`, `#goTo`)
- `attributeChangedCallback` trims non-primary views immediately when the attribute is set
- Translated "Single Section Scroll" across all 29 locales

Closes #3663

## Test plan
- [x] Unit tests pass (945 passed)
- [x] Lint clean
- [x] Toggle "Single Section Scroll" in scrolled mode — verify only one section renders
- [x] Navigate between sections — verify previous section is destroyed
- [x] Disable the toggle — verify adjacent sections load again
- [x] Verify toggle is disabled when not in scrolled mode

### DIFF
--- a/apps/readest-app/src/services/commandRegistry.ts
+++ b/apps/readest-app/src/services/commandRegistry.ts
@@ -393,6 +393,12 @@ const controlPanelItems = [
     section: 'Scroll',
   },
   {
+    id: 'settings.control.scroll.noContinuousScroll',
+    labelKey: _('Single Section Scroll'),
+    keywords: ['single', 'section', 'scroll', 'continuous', 'one', 'chapter'],
+    section: 'Scroll',
+  },
+  {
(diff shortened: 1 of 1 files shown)


sevillaarvin:
Appreciate this. Thanks!

PR #3699, opened 2026-03-31
Claude/integrate shadow libraries 20kxf

kirihitogaryu:


chrox:
We won't provide access to shadow libraries directly in Readest.

PR #3723, opened 2026-04-01
feat(rsvp): progress bar per chapter, speed selector dropdown, and UX improvements

lexm2:
all changes are based on what i expected as a user when using rsvp for the first time.

 **chapter progress bar scoped to one chapter**
the bar previously represented all loaded content at once. it now tracks one spine item at a time, filling 0–100% per chapter and resetting on advance

**progress bar is draggable**
tap and drag the progress bar to seek to any word in the current chapter

**wpm badge is now a speed selector dropdown**
lets you choose your reading speed from presets directly in the header

**pausing during countdown**
clicking pause while the 3-2-1 countdown is running now cancels it cleanly

**chapter header updates during playback**
the chapter title now updates as playback crosses toc section boundaries. previously it was stuck on "select chapter" on entry and after advancing

**chapter dropdown navigation works correctly**
selecting a chapter now actually navigates to it and restarts playback from there

**resume from saved position restores correctly**
it was starting from the beginning of the loaded chunk instead of the actual saved word, including when the saved position was in a different section

**stop at end of book**
playback was restarting the last chapter on reaching the end, it now stops cleanly

**wpm and punctuation delay settings persist across sessions**
speed and punctuation delay preferences are now saved and restored between sessions

**active chapter scrolls into view**
opening the chapter dropdown scrolls the current chapter to center

**rename "pause" to "punctuation delay"**
clearer label for what the setting actually controls

**header buttons no longer shrink on click**
removed the active scale animation from close, chapter, and speed selector buttons

**snappier chapter transitions**
chapter advance now uses the renderer's relocate event instead of a fixed 500ms timeout

**internal: one spine section at a time**
rsvp now processes one spine section at a time instead of extracting all loaded documents simultaneously, removing a lot of old multi-section boundary tracking

### DIFF
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -1030,7 +1030,6 @@
   "words": "كلمات",
   "{{time}} left": "متبقي {{time}}",
   "Reading progress": "تقدم القراءة",
-  "Click to seek": "انقر للتمرير",
   "Skip back 15 words": "تخطي للخلف 15 كلمة",
   "Back 15 words (Shift+Left)": "للخلف 15 كلمة (Shift+اليسار)",
   "Pause (Space)": "إيقاف مؤقت (المسافة)",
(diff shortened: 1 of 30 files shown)


.vscode/settings.json
```diff
@@ -27,5 +27,8 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "prettier.requireConfig": true,
-  "prettier.documentSelectors": ["**/*.{js,jsx,ts,tsx,css,json,md,html,yml}"]
+  "prettier.documentSelectors": [
+    "**/*.{js,jsx,ts,tsx,css,json,md,html,yml}"
+  ],
+  "cmake.sourceDirectory": "/home/lex/Documents/Github/readest/packages/simplecc-wasm/OpenCC"
```
chrox: Please revert this.

lexm2: reverted

chrox:
LGTM. Thank you.

PR #3724, opened 2026-04-02
feat: Hardcover.app Sync

AnandakrishnanV:
## Summary

This PR adds one-way Hardcover sync to Readest and improves ISBN handling so imported books can be matched to Hardcover more reliably. (Apologies for the big PR, Can break it up into two if needed)

Users can now:

- connect a Hardcover account with an API token
- enable or disable Hardcover sync globally
- enable Hardcover sync per book
- push reading progress to Hardcover
- push annotations and excerpts to Hardcover

This PR also promotes valid ISBNs into a dedicated `metadata.isbn` field.

### Hardcover integration

- Added a dedicated Hardcover client and GraphQL layer for:
  - token validation
  - book lookup by ISBN, with title/author fallback
  - read-progress sync
  - reading-journal sync
  - request pacing and retry handling
- Added a server-side GraphQL proxy for web requests to avoid CORS issues.
- Added a persistent note-to-journal mapping store:
  - `localStorage` on web
  - SQLite on native
- Added a database migration for native note-sync mappings.

### Reader UI and settings

- Added a new Hardcover settings dialog.
- Added global Hardcover settings to app settings.
- Added per-book Hardcover sync enablement.
- Added reader menu actions for:
  - opening Hardcover settings
  - enabling sync for the current book
  - pushing progress
  - pushing notes
- Added sync status toasts and last-synced timestamp updates.

### Progress and note sync behavior

- Progress sync creates or updates the active Hardcover read entry.
- Progress sync promotes a book to `Currently Reading` when progress is pushed.
- Progress sync marks a book as `Read` when progress reaches 100%.
- Note sync exports annotations and excerpts as Hardcover reading journals.
- Note sync deduplicates overlapping Readest annotation/excerpt pairs by text plus CFI base node.
- Note sync skips unchanged entries using stored payload hashes.
- Note-only sync no longer changes an existing Hardcover book’s reading status.

### Metadata and ISBN handling

- Added `isbn` as a first-class metadata field and type.
- Added ISBN editing and validation in the metadata editor.
- Updated source selection UI to show ISBN.
- Added import-time normalization that extracts valid ISBNs from `identifier` and `altIdentifier` and stores them in `metadata.isbn`.
- Added default Hardcover settings and the new per-book config flag to the relevant settings and book types.

## Implementation Areas

This PR updates:

- Hardcover service/client code
- Hardcover GraphQL definitions
- Hardcover sync persistence and migration code
- the web proxy route
- reader settings UI
- reader sidebar/book menu UI
- the reader sync hook
- metadata editor and metadata source selection
- document, metadata, settings, and book type definitions
- import-time book metadata normalization
- unit tests for Hardcover client behavior, sync mapping persistence, and ISBN promotion during import

## Testing

Validated with focused tests for:

- token normalization
- ISBN extraction
- progress date formatting
- note deduplication
- rate-limit retry behavior
- status promotion during progress sync
- preventing status changes during note-only sync
- sync mapping persistence on web and native
- import-time ISBN promotion

## Notes

- This is a one-way sync from Readest to Hardcover.
- Hardcover sync is opt-in globally and per book.
- On web, Hardcover requests go through a local proxy route to avoid CORS issues.
- On native, sync mappings are stored in SQLite; on web, they are stored in `localStorage`.

### DIFF
--- a/apps/readest-app/src/services/hardcover/HardcoverClient.ts
+++ b/apps/readest-app/src/services/hardcover/HardcoverClient.ts
@@ -397,10 +397,13 @@ export class HardcoverClient {
     const percent = totalPages > 0 ? (boundedPage / totalPages) * 100 : 0;
 
     let entry = '';
-    if (note.text) {
-      entry += `> "${note.text}"\n\n`;
+    if (note.text?.trim()) {
+      entry += note.text.trim();
     }
     if (note.note) {
+      if (entry) {
+        entry += '\n\n━━━\n\n';
+      }
       entry += note.note;
     }
 
--- a/apps/readest-app/src/__tests__/services/hardcover/HardcoverClient.test.ts
+++ b/apps/readest-app/src/__tests__/services/hardcover/HardcoverClient.test.ts
@@ -32,7 +32,7 @@ type HardcoverClientTestApi = {
     note: BookNote,
     config: BookConfig,
     context: TestBookContext,
-  ) => { action_at: string };
+  ) => { action_at: string; entry: string; event: string };
   ensureBookInLibrary: (book: Book, isReading?: boolean) => Promise<TestBookContext | null>;
   pushProgress: (book: Book, config: BookConfig) => Promise<void>;
 };
@@ -251,6 +251,30 @@ describe('HardcoverClient', () => {
     expect(progressCall).toBeDefined();
     const variables = progressCall?.[1] as { started_at?: string } | undefined;
     expect(variables?.started_at).toMatch(/^\d{4}-\d{2}-\d{2}$/);
+  });
+
+  test('should format multiline quote text with a short divider before the note', () => {
+    const note = {
+      id: '1',
+      type: 'annotation',
+      text: "She smiled. 'Are you, Overseer? Still?'\n\n'What do you mean?'",
+      note: 'Follow-up note',
+    } as BookNote;
+    const config = { progress: [5, 100] } as BookConfig;
+    const context: TestBookContext = {
+      editionId: 2,
+      pages: 100,
+      bookId: 1,
+      bookPages: 100,
+      userBook: null,
+    };
+
+    const payload = clientApi.buildJournalPayload(note, config, context);
+
+    expect(payload.event).toBe('note');
+    expect(payload.entry).toBe(
+      "She smiled. 'Are you, Overseer? Still?'\n\n'What do you mean?'\n\n━━━\n\nFollow-up note",
+    );
   });
 
   test('should promote an existing user book to currently reading before syncing progress', async () => {


apps/readest-app/src/services/hardcover/HardcoverClient.ts
```diff
@@ -0,0 +1,543 @@
...
+
+  private formatDate(date: Date): string {
+    return date.toISOString().replace(/\.\d+/, '').replace('Z', '+00:00');
+  }
+
+  private formatDay(date: Date): string {
+    return date.toISOString().slice(0, 10);
+  }
+
+  private async paceRequest() {
```
chrox: Should be better to use throttled requests here.

apps/readest-app/src/services/hardcover/HardcoverClient.ts
```diff
@@ -0,0 +1,543 @@
...
+    const rawNotes = (config.booknotes ?? []).filter(
+      (note) => (note.type === 'annotation' || note.type === 'excerpt') && !note.deletedAt,
+    );
+
+    // Readest can keep both an excerpt (quote) and an annotation (quote + note)
+    // for the same highlight, sometimes with slight CFI trailing offset differences.
+    // We group by the CFI base node and text to safely dedupe.
+    const getDedupKey = (n: BookNote) => {
+      const text = n.text?.trim() || '';
+      const cfiNode = n.cfi ? n.cfi.split(',')[0] : '';
```
chrox: The first part of `n.cfi.split(',')` is the beginning of the range of the text selection. This will dedupe all annotations with the same start point. Please reconsider if it's safe or desired.

apps/readest-app/src/services/hardcover/HardcoverSyncMapStore.ts
```diff
@@ -0,0 +1,157 @@
...
+      await db.close();
+    }
+  }
+
+  async loadForBook(bookHash: string): Promise<void> {
+    this.loadedBookHash = bookHash;
+    this.mappings.clear();
+    this.modified = false;
+
+    if (this.isWebStorageAvailable()) {
```
chrox, line 47: Should we stick with the wasm turso database for the web version? Or is it not working yet?

apps/readest-app/src/services/bookService.ts
```diff
@@ -28,9 +28,66 @@ import { deserializeConfig, serializeConfig } from '@/utils/serializer';
 import { ClosableFile } from '@/utils/file';
 import { TxtToEpubConverter } from '@/utils/txt';
 import { svg2png } from '@/utils/svg';
+import { validateISBN } from '@/utils/validation';
 import { BookFileNotFoundError } from './errors';
 import { fetch as tauriFetch } from '@tauri-apps/plugin-http';
 
+const extractIsbnCandidates = (value: unknown): string[] => {
```
chrox: Should we move this to a `utils/isbn.ts` file?

apps/readest-app/src/services/bookService.ts
```diff
@@ -28,9 +28,66 @@ import { deserializeConfig, serializeConfig } from '@/utils/serializer';
...
+  if (prefixedMatches) {
+    for (const match of prefixedMatches) {
+      candidates.push(match.replace(/^[\s,;]+/, '').replace(/^(?:urn:)?isbn:/i, ''));
+    }
+  }
+
+  return candidates;
+};
+
+const normalizeMetadataIsbn = (metadata: BookDoc['metadata']) => {
```
chrox: Should we move this to a `utils/isbn.ts` file?

apps/readest-app/src/services/hardcover/HardcoverClient.ts
AnandakrishnanV: Good point. Changed to a req Queue, moved the interval to a var

chrox: Please also check if `utils/throttle` works for this.

apps/readest-app/src/services/hardcover/HardcoverClient.ts
AnandakrishnanV: Good catch, I did have quite a bit of trouble with this part initially. I’ve updated the dedupe logic to now compare a normalized CFI instead of just the starting segment, so the match should be narrower now and be more reliable for highlight vs highlight-with-note pairs

chrox:
Also, you can run `pnpm format` to format the code before pushing the commit. We will add this in the hook later.
It's already setuped:
```
"lint-staged": {
    "**/*.{js,jsx,ts,tsx,css,json,md,html,yml}": [
      "prettier --write"
    ]
  },
```
please check if your `prettier` tool is working.

apps/readest-app/src/services/hardcover/HardcoverSyncMapStore.ts
AnandakrishnanV: I tried switching the web sync map over to the WASM Turso path, but it failed in the actual browser runtime on my self-hosted setup with The WebAssembly.Memory object cannot be serialized. Copilot suggested that it may be because Cross-Origin-Opener-Policy and Cross-Origin-Embedder-Policy are not being provided by the runtime yet. I’m not fully familiar with that part of the setup, with WASM, so I’m not sure yet whether that’s specific to my deployment or a more general issue with the current web runtime.

chrox: For the wasm version of turso it does have some issues which have been fixed in my local working branch. Please go ahead with this and after this is merged I will refactor this with turso db storage.

chrox: Yes we need to setup the Cross-Origin-Opener-Policy and Cross-Origin-Embedder-Policy headers and also fix the file path of the database.

apps/readest-app/src/services/hardcover/HardcoverClient.ts
AnandakrishnanV: I had taken a lot at it, and don't think its a good fit. The Hardcover reqs need to be awaitable, and the util one looks more geared toward UI-based throttling where calls can be coalesced. Since this flow is stateful, the function above feels a lot safer for this use case. That said, I could move it into a separate util if that makes sense, since it does seem like something that could be reused for other integrations.

apps/readest-app/src/services/bookService.ts
AnandakrishnanV: Yep. Moved to utils.

apps/readest-app/src/services/bookService.ts
AnandakrishnanV: Yep. Moved to utils.

PR #3740, opened 2026-04-04
feat(opds): support custom catalog headers with web proxy consent

### LINKED ISSUE #3739: FR: Support custom OPDS headers, with safe handling for hosted web
## Summary

It would be useful to support custom headers for OPDS catalogs, especially for deployments protected by services like Cloudflare Access.

Examples:
- `CF-Access-Client-Id`
- `CF-Access-Client-Secret`

## Why this matters

Some OPDS servers are not accessible with only:
- basic auth
- digest auth
- URL credentials

They require custom request headers to reach the feed or download content.

## Important design constraint for hosted web

On the hosted web app, OPDS requests currently go through the existing OPDS proxy path for browser compatibility / CORS reasons.

That means custom secret-bearing headers need special handling on web. In particular, they should **not** be transported in a way that exposes them in request URLs or other easy-to-log surfaces.

This is especially important for users trying to access OPDS servers behind Cloudflare Access or similar header-based auth.

## Expected behavior

- Native / local app:
  - allow per-catalog custom headers
  - send them directly as request headers

- Hosted web app:
  - explicitly do not support secret-bearing custom headers on hosted web, and document that limitation clearly in the UI

## Suggested UX

Per-catalog fields such as:
- Custom headers
- One header per line: `Header-Name: value`

Or more guided fields for common cases like:
- Cloudflare Access Client ID
- Cloudflare Access Client Secret

## Notes

This request is specifically about enabling OPDS catalogs behind header-based auth, while making sure the hosted web OPDS flow handles that safely.

ShunnyBunny:
## Summary

This PR adds support for per-catalog custom OPDS headers and wires them through the OPDS request flow.

### What changed

- Added per-catalog `customHeaders` support for OPDS catalogs
- Added custom header parsing/formatting helpers for the OPDS catalog form
- Passed custom headers through:
  - OPDS catalog validation
  - feed loading
  - auth probing
  - downloads
  - cached image fetching
  - the web OPDS proxy route
- Persisted custom headers in saved catalog settings
- Displayed an indicator in the catalog manager when a catalog has custom headers configured
- Added a required hosted-web consent checkbox when sensitive OPDS inputs are present (username, password, or custom headers)

### Why

Some OPDS servers require header-based authentication rather than only basic auth, digest auth, or URL credentials. This is especially relevant for setups protected by services like Cloudflare Access.

On hosted web, OPDS requests are proxied, so this PR also adds an explicit consent step to make that behavior clear before saving a sensitive OPDS connection.

Closes #3739.

## Testing

- `pnpm -C apps/readest-app test -- src/__tests__/utils/opds-custom-headers.test.ts src/__tests__/utils/opds-req.test.ts`
- `pnpm -C apps/readest-app build-web`
- Built the Apple Silicon macOS app bundle locally via Tauri for manual verification
- Manually verified the OPDS catalog form flow in the web app, including the web proxy consent checkbox behavior

### DIFF
--- a/apps/readest-app/src/__tests__/app/opds/opds-utils.test.ts
+++ b/apps/readest-app/src/__tests__/app/opds/opds-utils.test.ts
@@ -546,8 +546,12 @@ describe('opdsUtils', () => {
       expect(result.isValid).toBe(false);
     });
 
-    it('should pass username and password to fetchWithAuth', async () => {
+    it('should pass auth and custom headers to fetchWithAuth', async () => {
       const xmlFeed = `<feed xmlns="http://www.w3.org/2005/Atom"><title>Auth Feed</title></feed>`;
+      const customHeaders = {
+        'CF-Access-Client-Id': 'client-id',
+        'CF-Access-Client-Secret': 'client-secret',
+      };
 
       mockFetchWithAuth.mockResolvedValue({
         ok: true,
@@ -556,7 +560,7 @@ describe('opdsUtils', () => {
         headers: new Headers(),
       } as Response);
 
-      await validateOPDSURL('https://example.com/opds', 'user', 'pass', true);
+      await validateOPDSURL('https://example.com/opds', 'user', 'pass', true, customHeaders);
 
       expect(mockFetchWithAuth).toHaveBeenCalledWith(
         'https://example.com/opds',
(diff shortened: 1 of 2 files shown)


chrox:
LGTM. Thanks.

PR #3755, opened 2026-04-05
feat(rsvp): split words option, faster countdown, and skip pages RSVP cant open

lexm2:
- Split words option -- adds a toggle to split words in RSVP display,        
  improving readability when reading fast. Includes unit tests. (i.e "split-word" will get displayed in two parts
as "split-" "word")
 - Faster countdown -- reduces the 3-2-1 start countdown from 2.4s to 1.5s    
  (500ms per tick instead of 800ms). The previous delay felt too long before    
  reading even started.
 - Auto-skip image-only pages -- when RSVP encounters a page with no readable 
  text (e.g. a cover image or title slide), it now automatically advances to the
   next section instead of silently failing.
- Fix RSVP focus moving -- when a new chapter loaded, the context     
  window (the sliding display buffer) wasn't reset, causing the RSVP word       
  position to visually jump as the DOM expanded. Now the context window resets  
  whenever the chapter's word list changes, so the entire chapter is loaded and it wont jump. Also you can scroll through the context to see the entire chapter not just what you read.

 - Fix validation tests -- two date validation tests failed due to a UTC      
  timezone boundary issue when parsing years at the edges of the valid range    
  (1000, currentYear + 10). Fixed by reading the year directly from the input 
  string instead of via Date.getFullYear(). Not part of RSVP code but made my prepush hook fail.

### DIFF
--- a/apps/readest-app/src/services/rsvp/utils.ts
+++ b/apps/readest-app/src/services/rsvp/utils.ts
@@ -181,7 +181,7 @@ export function segmentCJKText(text: string): string[] {
  */
 export function getHyphenParts(word: string): string[] {
   if (!/[a-zA-Z]-[a-zA-Z]/.test(word)) return [word];
-  const parts = word.split(/(?<=[a-zA-Z])-(?=[a-zA-Z])/);
+  const parts = word.split(/-(?=[a-zA-Z])/);
   return parts.map((part, i) => (i < parts.length - 1 ? part + '-' : part));
 }
 


apps/readest-app/src/utils/validation.ts
lexm2, line 1: `new Date('1000-01-01')` is parsed as UTC midnight 1000-01-01T00:00:00Z

date.getFullYear() returns the year in local time, for me this is behind UTC so midnight becomes 999-12-31T19:00:00 so getFullYear() returns 999 instead of 1000.

Only a problem for westerners though.

chrox:
LGTM. Thanks.

PR #3774, opened 2026-04-06
feat(library): align grid hover highlight corner radius

WhiteHades:
this pr adds:
- adjustment of the library grid item hover highlight radius to better match the book cover corner treatment

image


chrox:
Looks good. Thanks.

PR #3784, opened 2026-04-07
fix(dictionary): add Chinese dictionary lookup with pinyin support

### LINKED ISSUE #3783: Dictionary lookup fails for Chinese characters. No pinyin or definitions shown
## Bug Description

The dictionary lookup feature does not work for Chinese characters. When selecting a Chinese word in a book and tapping "Dictionary", the popup shows "Error -- Unable to load the word" for most entries.

This is caused by the Wiktionary REST API (`/api/rest_v1/page/definition/{word}`) not supporting Chinese entries:

- **Simplified characters** (e.g. 书, 龙) return HTTP 404 -- their Wiktionary pages are stub entries that redirect to traditional forms via `{{zh-see}}` templates, which the REST API cannot follow.
- **Traditional characters** (e.g. 書, 龍) return a response, but with no Chinese (`"zh"`) key -- only Japanese or Vietnamese entries appear, so the app finds no results.
- **Multi-character words** (e.g. 你好) sometimes load definitions, but **never include pinyin/pronunciation data** since the REST API definition endpoint strips pronunciation sections entirely.

Additionally, the fetch URL does not use `encodeURIComponent()`, which can cause issues with CJK characters on some platforms (notably Android WebView).

## Expected behavior

Selecting a Chinese word and tapping "Dictionary" should display:

- The word's **pinyin romanization** (e.g. "shu" for 書/书)
- **English definitions** from Wiktionary
- Automatic resolution of simplified to traditional character redirects (e.g. 书 -> 書)

## Technical inputs

Operating System: Android (reproducible on all platforms)

Readest Version: 0.10.4

The Wiktionary Action API (`action=parse&prop=wikitext`) returns complete wikitext including pinyin (via `{{zh-pron|m=...}}` templates) and definitions, and can be used as an alternative data source for Chinese lookups while keeping the existing REST API for all other languages.

pythontyphon:
## Summary

Fixes #3783

- Add a Chinese-specific dictionary service using the Wiktionary Action API (`action=parse`), which returns full wikitext with pinyin and definitions
- Handle simplified-to-traditional character redirects (e.g. 书 → 書) by detecting `{{zh-see}}` templates
- Display pinyin romanization prominently in the dictionary popup for Chinese words
- Add `encodeURIComponent()` to the Wiktionary REST API URL for non-Chinese languages
- Keep existing Wiktionary REST API behavior unchanged for all other languages

## Test plan

- [x] Unit tests for wikitext parsing (pinyin extraction, redirect detection, definition parsing, markup cleanup) — 25 tests passing
- [x] Biome lint passes on all changed files
- [x] Manual testing on Android debug APK:
  - 书 (simplified single char → redirects to 書, shows "shū")
  - 你好 (multi-char → shows "nǐ hǎo")
  - English words → unchanged behavior via REST API

### DIFF
--- a/apps/readest-app/src/__tests__/services/dictionaries/chineseDict.test.ts
+++ b/apps/readest-app/src/__tests__/services/dictionaries/chineseDict.test.ts
@@ -1,0 +1,260 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import {
+  parseRedirect,
+  parsePinyin,
+  cleanWikiMarkup,
+  parseDefinitions,
+  fetchChineseDefinition,
+} from '@/services/dictionaries/chineseDict';
+
+// Sample wikitext for 書 (traditional, full entry)
+const WIKITEXT_SHU_TRADITIONAL = `
+==Chinese==
+
+===Glyph origin===
+Some origin text.
+
+===Pronunciation===
+{{zh-pron
+|m=shū
+|c=syu1
+|cat=n,v,pn
+}}
+
+===Definitions===
+{{head|zh|hanzi}}
+
+# [[book]]; [[codex]]
+# [[letter]]; [[document]]
+# form of a written or printed [[Chinese character]]; [[style]]
+# {{lb|zh|literary}} [[Chinese character]]; [[writing]]; [[script]]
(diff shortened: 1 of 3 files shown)


chrox:
Looks great. Thanks.

Issue #1711, opened 2025-07-30
FR: 书籍打开非常卡顿


m2cc:
**Does your feature request involve difficulty for you to complete a task? Please describe.**

> 环境win11操作系统，Readest Version: 0.9.67、0.9.68（包括之前的很多版本都是如此），书籍打开后，会卡住好几秒，无法滚动，文字渲染很慢

**Describe the solution you'd like**

> 书籍正常打开，无卡顿，内容滚动流畅，文字渲染无延迟，左侧目录栏正常渲染

**Describe alternatives you've considered**

> 无

**Additional context**

> 文件地址：https://drive.google.com/file/d/1OsHo1Un0xbuoM4TwCcMibJVzyL_qdjUV/view?usp=drive_link

chrox:
@m2cc I have requested access of the file, please check.

m2cc:
> [@m2cc](https://github.com/m2cc) I have requested access of the file, please check.

你再看看，我已经共享给你了

chrox:
This epub file has a single html file that includes all the chapters of the book which is not a good design. We also need to improve the performance of Readest for these cases in the future. I'll label this issue as an enhancement.

For now you can convert the epub to epub format with calibre, with the default parameters this will automatically generate an epub with multiple chapters parsed from the table of contents and readest renders the book without lag on the testing devices.

m2cc:
> This epub file has a single html file that includes all the chapters of the book which is not a good design. We also need to improve the performance of Readest for these cases in the future. I'll label this issue as an enhancement.这个 epub 文件只有
> 
> For now you can convert the epub to epub format with calibre, with the default parameters this will automatically generate an epub with multiple chapters parsed from the table of contents and readest renders the book without lag on the testing devices.

好的，感谢您，我将使用calibre对书籍进行转换，再进行阅读。

KyleCaiW:
I've encountered the same problem. I look forward to the team enhancing the support for this as soon as possible

ybbluesky:
> This epub file has a single html file that includes all the chapters of the book which is not a good design. We also need to improve the performance of Readest for these cases in the future. I'll label this issue as an enhancement.
> 
> For now you can convert the epub to epub format with calibre, with the default parameters this will automatically generate an epub with multiple chapters parsed from the table of contents and readest renders the book without lag on the testing devices.

程序有无办法处理一下？用户有的不会calibre

ybbluesky:
[...]

PR #3839, opened 2026-04-12
perf(reader): optimize book open for EPUBs

lexm2:
- Defer TOC location indexing and annotation imports off the critical         
    path via requestIdleCallback, unblocking first paint                        
  - Prefetch the target section after view.open so decompression overlaps       
    with style/attribute setup                                                  
  - Preload foliate-js view module during initViewState so it is cached         
    by the time FoliateViewer mounts                                            
  - Cache ZIP central directory entry index at import time and use              
    Blob.slice + native DecompressionStream on re-open, skipping the            
    expensive zip.js getEntries() scan entirely                                 
  - Enable native DecompressionStream for all zip.js reads                      
  - Cache metaHash across opens to avoid redundant recomputation
  
  Requires https://github.com/readest/foliate-js/pull/9 to be           
  merged first this PR's deferred `updateToc` calls                           
  `bookDoc.ensureSubItemsResolved()` which is added in that foliate-js          
  patch. Without it, TOC subitems won't resolve.


chrox:
Thanks for the work here, these optimizations look promising.

That said, it would be great to first profile the performance and identify the actual hotspots before applying these changes, to avoid premature optimization and ensure we’re addressing the real bottlenecks.

chrox:
> Cache ZIP central directory entry index at import time and use
Blob.slice + native DecompressionStream on re-open, skipping the
expensive zip.js getEntries() scan entirely

I’m not entirely convinced that getEntries() is a real bottleneck for typical EPUB files. For archives with only a few dozen entries, zip.js should handle parsing efficiently, it can access file headers directly and optimize entry structures without needing to read the entire ZIP content.

What has actually been identified as a bottleneck, in some cases, is the rendering process, for example in https://github.com/readest/readest/issues/1711. Currently, the renderer needs to process at least one section before anything is painted in the reader window on the first open. While browsers on some platforms cache rendering effectively, making subsequent opens much faster, the initial load still incurs this cost.

Overall, it would be beneficial to establish a quantitative performance profiling system to better measure and validate these kinds of optimizations.

lexm2:
I did profile these changes before submitting and  measured roughly a 14% time reduction when opening books with all changes excluding the folate.js change (the test book was 2000+ chapters). Every individual change contributing at least some improvement. The zip.js getEntries() cache and native DecompressionStream path did show measurable gains.

That said, your point about the rendering bottleneck is well taken.  That's a more fundamental issue that micro optimizations here won't fully address. I think the best way to optimize is an architectural approach (e.g. painting something before the first section finishes rendering) rather than squeezing time off the current init path. Going to close this PR for now in favor of exploring that direction.

PR #3845, opened 2026-04-12
Update package installation for Linux-x64

Drsheppard01:
This will fix the white screen error on Fedora 44


chrox:
Interesting, and how did you verify this?

Drsheppard01:
> Interesting, and how did you verify this?

Most applications which used Tauri and worked for me use this deps: https://github.com/cjpais/Handy/blob/main/.github/workflows/build.yml#L108 https://github.com/0xle0ne/wireguard-gui/blob/master/.github/workflows/build.yml#L51

chrox:
Great. We will see as I cannot test and verify this. Thanks.

PR #3892, opened 2026-04-18
feat(kosync): defer push after resume

### LINKED ISSUE #3888: Easy to lose kosync progress when reopening
## Bug Description

When reopening/resuming the app, it can be easy to unintentionally overwrite KOSync progress if the reader components didn't get unmounted. This is particularly noticeable on mobile (I tested on Android).

**Example:**
If you read for a bit, then press the power button, wait a few minutes, and unlock the screen, it will open back up to the page you were previously looking at. At the same time, it thinks `progress` has changed, so it immediately schedules a (debounced) `pushProgress`. If you advanced progress on another device during the downtime, it may get overwritten (depending on the behavior of your sync server).

## Expected behavior

Currently, after the component is mounted, it waits for the first `pullProgress` before attempting any progress pushes. I suggest this should apply when the app is hidden and resumed as well.  I'm happy to submit a PR that implements this if you'd like.

## Technical inputs

Operating System: I tested Windows, Android, and Web. It's most noticeable on Android because just hitting the power button without closing the app is a common behavior. On the other platforms it's more likely that you will close the app/tab, which sidesteps this issue.

Readest Version: 0.10.6

zb140:
After "resuming" the app, defer `pushProgress` until after the first `pullProgress`. This mirrors what happens on first opening a book.

Fixes #3888 .

Implementation details:
Introduces a new hook, `useWindowActiveChanged`, to abstract out the decision about what makes a window "active" or not. On most platforms, the standard DOM `visibilitychange` event is used, but this event is unreliable on desktop platforms. (On macOS it only fires when the window is actually maximized or minimized, and on Windows it just never fires at all). On those platforms we use the Tauri focus/blur events instead.

Uses a ref to maintain the latest `pushProgress` and `pullProgress` functions. This allows the `useEffect` calls to drop their dependencies on those functions, which means less subscribe/unsubscribe churn.

### DIFF
--- a/apps/readest-app/src/app/reader/hooks/useWindowActiveChanged.ts
+++ b/apps/readest-app/src/app/reader/hooks/useWindowActiveChanged.ts
@@ -1,0 +1,62 @@
+// used to execute a callback when the "active" state of the current window changes.
+// On web abd mobile, "active" means "is visible". On desktop, the 'visibilitychange'
+// event is unrelaible, so "active" means "has focus".
+
+import { useEffect, useRef } from 'react';
+import environment from '@/services/environment';
+
+export type ActiveCallback = (isActive: boolean) => void;
+
+type Cleanup = () => void;
+async function activeChangedDesktop(onChange: ActiveCallback): Promise<Cleanup> {
+  const { getCurrentWindow } = await import('@tauri-apps/api/window');
+  const appWindow = getCurrentWindow();
+
+  const unFocus = await appWindow.listen('tauri://focus', () => onChange(true));
+  const unBlur = await appWindow.listen('tauri://blur', () => onChange(false));
+  return () => {
+    unFocus();
+    unBlur();
+  };
+}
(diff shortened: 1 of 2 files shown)


apps/readest-app/src/app/reader/hooks/useWindowActiveChanged.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  useEffect(() => {
+    let isAlive = true;
+    let unsub: Cleanup | undefined;
+    const onChange = (isActive: boolean) => {
+      onActiveChanged.current?.(isActive);
+    };
+
+    environment
+      .getAppService()
```
chrox: Please refer to `apps/readest-app/src/hooks/useTransferQueue.ts` on how to use `appService` in a hook.

chrox (changes requested):

chrox (approved):

PR #3951, opened 2026-04-24
feat(opds): add OPDS-PSE streaming support and custom OPDS 2.0 parser

### LINKED ISSUE #2970: FR: OPDS-PS support
OPDS-PS is an extension of OPDS that allows for clients to stream individual pages from the server instead of requiring the full book to be sent. It also keeps the server's reading position in sync with the client's current reading position. I'd like to be able to stream books from my server and keep progress synced rather than downloading whole books individually. Kavita and Komga both support the spec but there are hardly any clients supporting it (only KOReader as far as I am aware).

zeedif:
Resolves #2970

Added a custom `opdsParser.ts` to replace the `foliate-js` dependency for parsing feeds. I previously opened a PR to `foliate-js` (johnfactotum/foliate-js#125) to fix OPDS 1.2 to 2.0 mapping and add PSE support, but since it has been pending for weeks, I brought the parser directly into Readest to avoid blocking this feature.

The OPDS-PSE integration logic in the reader (handling `pse://` virtual files and streaming them as CBZ) was mostly written using AI. I debugged it and successfully tested it against a Suwayomi instance serving OPDS 1.2 feeds with OPDS-PSE links, and it behaves as expected. Please review the implementation carefully to identify any potential issues.


chrox:
Would it be possible to refactor this for integration with https://github.com/readest/foliate-js? In Readest, we maintain our own fork of foliate-js, stay aligned with the upstream project with contributing changes back or regularly rebasing.

zeedif:
Of course! In fact, my original plan—as I mentioned—was to get my PR accepted in foliate-js. I can open the same PR against your fork so it can be integrated there instead. Then I would just remove the custom parser from this branch and import it again from foliate-js, which should work the same way. I’ll take a closer look at it tomorrow.

zeedif:
@chrox, following up on your suggestion, I've submitted the updated parser as a PR to the `readest/foliate-js` fork: [readest/foliate-js#14](https://github.com/readest/foliate-js/pull/14). Once it’s accepted, I’ll push a commit to remove the local opdsParser.ts from this branch and switch back to using the foliate-js dependency with these changes.

zeedif:
@chrox, I've finished swapping out the custom parser and switched over to the updated `foliate-js` dependency. Apologies for the noisy commit history; I had to `force-push` a few times to fix the test and TypeScript errors that popped up from the updated interfaces. Most of the adjustments were needed in `feedChecker.ts`, which was added in the `38d7ba8` auto-download commit.

One thing to note in `feedChecker.ts`: I added a fallback for the title, like `pub.metadata.title || acqLink.title || 'Untitled'`. Even though the OPDS spec says the title attribute should be there, it can still come back as `undefined` from the parser. This fallback just makes sure we handle it gracefully.

It should all be sorted now. Ready for look when you have a chance.

chrox (changes requested):
Thanks for tackling this — the `pse://` virtual-file approach reuses `comic-book.js` cleanly and the type refinement is a real improvement. Two issues I'd like to see addressed before merge:

### 1. Desktop streaming with auth is broken

`src/app/opds/page.tsx:507` only embeds `authHeader` into `streamUrl` when `useProxy` is true. On Tauri/desktop, `needsProxy(url)` returns `false`, so the bare URL is stored in `streamData.template`. Then `src/libs/document.ts:226` calls `fetch(url)` with no `Authorization` header and no `customHeaders`. Compare to `handleDownload` (`page.tsx:441-443`), which explicitly sets `headers['Authorization'] = authHeader` for the non-proxy path.

The result: any OPDS-PSE server requiring Basic/Digest auth (Komga, Kavita, Suwayomi behind a reverse proxy, etc.) will work on web but 401 on desktop.


### 2. Credentials get persisted and synced

When `useProxy` is true, `getProxiedURL(url, authHeader, …)` (`src/app/opds/utils/opdsReq.ts:65-88`) bakes the auth header into the proxy URL as the `auth` query parameter. That URL ends up in `streamData.template`, gets JSON-stringified, percent-encoded, and stored as `book.url = file.name` in `src/services/bookService.ts:460-462`.

Although the book is marked `deletedAt: Date.now()` (transient), `src/app/library/hooks/useBooksSync.ts:25-30` still syncs it on first push because `!book.syncedAt` is true:

```ts
const newBooks = library.filter(
  (book) =>
    !book.syncedAt ||
    lastSyncedAtBooks < book.updatedAt ||
    lastSyncedAtBooks < (book.deletedAt ?? 0),
);
```

So the user's Basic-auth credential gets shipped to the sync server in plaintext as part of the book record.

Suggested fix: either (a) store only the upstream OPDS template URL plus the OPDS catalog id in `streamData`, and rebuild the proxy URL with auth at fetch time from the stored catalog credentials, or (b) explicitly exclude `pse://` books from sync.

---

I'll leave the smaller items (title fallback for missing metadata title, retry on transient page-fetch failures, hard-coded `maxWidth=2000`, missing `pse:lastRead`/`pse:lastReadDate` write-back, magic-string `pse://`) as follow-ups — happy to file separate issues if useful.

chrox (approved):

chrox:
@zeedif Please verify it now works as expected after the refactoring in your PSE setup.

chrox:
It works with Komga OPDS-PSE extension for OPDS v1 feeds. Ref: https://komga.org/docs/guides/opds#opds-v1

PR #3960, opened 2026-04-26
fix(rsvp): resume at stop word, prevent section replay, restore full context

lexm2:
Three RSVP bugs and one web auth fix.

  ---

  **Advancing to the next chapter in RSVP was replaying the same chapter**      
   
  ### RSVP: prevent section replay after `primaryIndex` reverts                 
  After `nextSection()`, foliate-js's `#detectPrimaryView` reverts
  `renderer.primaryIndex` back to the old section. The advance handler was      
  reading `indexBefore` from `primaryIndex` on every call, so a second advance
  compared the current section against the already-reverted index and replayed  
  the same section. The fix tracks the actual RSVP section in stable refs
  (`rsvpSectionRef` / `rsvpChapterHrefRef`) and filters out revert relocate     
  events (`newIndex === indexBefore`) so they don't resolve the advance promise.
   
  ---                                                                           
                  
  **When you exited RSVP, the book reopened at the wrong position**

  ### RSVP: resume at the exact word on exit                                    
  When leaving RSVP mode, `location` in BookConfig was left at whatever CFI a
  mid-session section transition had written, so reopening the book dropped you 
  at the wrong paragraph. `buildRsvpExitConfigUpdate` now pins `location` to the
  RSVP word's own CFI so normal-mode resumes at the right word.                 
   
  ---                                                                           
                  
  **The iOS perf fix trimmed the RSVP context window, scrollIntoView gives the same 
  performance gain while keeping the full chapter visible**

  ### RSVP: restore full-chapter context with `scrollIntoView`                  
  The old auto-scroll used `getBoundingClientRect` offsets plus a
  `contextWindow` state that sliced the word list, causing part of the chapter's text to be
  hidden. Replaced with `scrollIntoView({ block: 'nearest', behavior: 'instant'
  })`                                                                           
  on the highlighted word and removed the slicing entirely so the full chapter
  is always rendered.                                                           
                  
  ---

**On web, there was a narrow race where a sync could fire with an expired 
  token after a long idle**                                                 
                                                                                
  ### Auth: refresh expired token before web sync requests
  `AuthProvider` calls `refreshSession()` on mount, which covers the vast majority                                                               
  of cases. The edge case: a tab sits idle past Supabase's ~1 h token lifetime, 
  the user returns, and a sync fires before the background refresh completes
  sending a stale token and getting a 403. `getAccessToken` now decodes the JWT 
  and, if it expires within 30 s, awaits `supabase.auth.refreshSession()` before
  re-reading localStorage. Falls back to the cached token on refresh failure so 
  the server's auth error stays the source of truth. Native (Tauri) path is     
  unchanged `supabase.auth.getSession()` already auto-refreshes there.


chrox:
Please keep each PR minimal and avoid including unrelated changes, such as `apps/readest-app/.claude/skills/gstack` and `‎apps/readest-app/src/utils/access.ts`. If those changes are necessary, consider submitting them in a separate PR. Thanks!

chrox (approved):

PR #3972, opened 2026-04-27
chore(deps): bump react-window to the latest version

FabrisNick:
I updated react-window from 1.8.11 to 2.2.7


chrox:
image

There is a regression of the list layout:
1. The padding of the system fonts item is not the same with other items;
2. The system fonts popup window is not aligned with the main fonts popup window.

Please make sure the layout is exactly the same with the current web version at https://web.readest.com except that there is a good reason.

PR #4012, opened 2026-04-30
feat(reader): custom dictionaries (StarDict + MDict)

### LINKED ISSUE #170: Implement different providers for dictionary
**Does your feature request involve difficulty completing a task? Please describe.**
High possibility.

Easiest option would be to allow different providers.
Hardest option is to implement Yomichan, natively on the app. Seems hard to do on website.

Easiest would be to call for other online dictionaries, like [Jisho](https://jisho.org/) or [Kotobank](https://kotobank.jp). You can see a list of dictionaries I've collected [here](https://gist.github.com/matsumurae/ebf417928af48166dea2fc36a7bf29a6).

### LINKED ISSUE #2215: FR: Kindle style Dictionary, Wikipedia and Translation Cards
Hi Readest developers. 
Thank you for creating and maintaining this project. 

It would be great if you could add a Kindle style Dictionary, Wikipedia and Translation cards that float over the book when a word is long pressed. 

This Kindle feature is very comfortable, because it immediately gives the reader the opportunity to understand what a word means and then return to reading with minimum clicks. 

The current dictionary function in Readest requires two clicks, which might seem like small difference, but the Kindle system is much easier, faster and really helps maintain the reading flow. 

Here are two screenshots that shows how this works on iPad and iPhone.

Image

Image

Hope you can make this happen 🙏

### LINKED ISSUE #4003: Selection toolbar clips the last icon
## Bug Description

Just a minor thing:

Image

## Expected behavior

The box should be slightly larger so that all icons are visible.

## Technical inputs

Operating System: macOS Tahoe

Readest Version: 0.10.6 (Chrome 147.0.0.0)

### LINKED ISSUE #4006: Link to Wikipedia

Usually the provided description from Wikipedia is very brief or even a disambiguation page. Unfortunately I cannot easily navigate to the Wikipedia site/app from there.

Simply link to the Wikipedia page from the popup. This also follows the principle of least surprise, since IMO navigating to Wikipedia from there is the expectation.

### LINKED ISSUE #4013: FR: better dictionary support for non-English books

For a book in Spanish (with right metadata) instead of the definition of a word in https://es.wiktionary.org/ I get the translation of the word in https://en.wiktionary.org/, which is quite useless for a native speaker. And for that there already is the translator.

A three folded strategy to improve the dictionary feature, in order of importance:

- Use `<LANG>.wiktionary.org`
- Provide a link (similar to https://github.com/readest/readest/issues/4006)
- Add a share intent to the selection menu to be able to open external dictionaries or browsers.

chrox:
## Summary

Adds a pluggable dictionary provider system with built-in Wiktionary +
Wikipedia (relocated from the legacy single-popup buttons into a tabbed
shell) plus user-importable **StarDict** (`.ifo`/`.idx`/`.dict.dz`/`.syn`)
and **MDict** (`.mdx`/`.mdd`) bundles.

Closes #4013.
Closes #4006.
Closes #4003.
Closes #2215.
Closes #170.

- **Reader popup**: tabbed UI, per-tab lookup history, scroll-aware back
  button, last-active tab persists. Tabs grow to natural width up to a
  cap and truncate with ellipsis when crowded; phantom-bold layer prevents
  layout shift on focus.
- **Settings → Language → Dictionaries**: import / enable / drag-reorder
  (`@dnd-kit`) / delete (delete-mode toggle mirrors `CustomFonts`).
- **StarDict reader is self-contained** (replaces unused
  `foliate-js/dict.js`): lazy random-access binary search on `.idx` /
  `.syn` (`Int32Array` of byte offsets, no parsed-entry array on heap),
  lazy DictZip chunk decompression via fflate's streaming `Inflate`
  (`inflateSync` rejects per-chunk DictZip data; the streaming class
  doesn't), and an optional `.idx.offsets` sidecar generated at import to
  skip the init scan. **cmudict 105K-entry init drops from ~10 MB heap +
  2 MB IO to ~1.7 MB heap + ~500 KB IO.**
- **MDict** uses [`readest/js-mdict`](https://github.com/readest/js-mdict)
  (added as a submodule, consumed via tsconfig paths so its devDeps stay
  out of `pnpm-lock.yaml`). The fork adds a browser-friendly
  `BlobScanner` reading via `blob.slice(...).arrayBuffer()` — slices are
  lazy when the Blob is Readest's `NativeFile` / `RemoteFile`. `encrypt=2`
  (key-info-only) is fully supported via ripemd128-based `mdxDecrypt`;
  `encrypt=1` (record-block, requires user passcode) surfaces as
  `unsupported`.
- **Wikipedia annotation tool removed** — Wikipedia is now a tab inside
  the unified popup. Stale `annotationQuickAction === 'wikipedia'` is
  coerced to `'dictionary'` on settings load. Legacy `WiktionaryPopup` /
  `WikipediaPopup` deleted.
- **iOS-friendly external links**: skip `target="_blank"` on Tauri to
  avoid the WebView's "open externally" path triggering the shell scope
  error; the popup's container click handler routes through `openUrl`
  (matches `Link.tsx` semantics).
- **i18n**: 939 strings across 31 locales (30 base keys + CLDR plural
  forms for ar/he/sl/pl/ru/uk/ro/it/pt/fr/es).

3396 unit tests pass; lint clean. Test fixtures bundled (cmudict, eng-nld,
Longman Phrasal Verbs MDX) so the suite exercises the production paths
against real-world data.

## Test plan

- [x] `pnpm dev-web` — open a book, select a word, click the Dictionary
      tool. Verify Wiktionary tab loads (network), Wikipedia tab loads only
      when activated. Verify Wikipedia tool button is gone from the
      annotation toolbar.
- [x] Settings → Language → Dictionaries: import a real StarDict bundle
      (e.g. WordNet from kdr2/Wordnet-StarDict). Verify it appears, lookup
      a word, the StarDict tab shows results.
- [x] Import a real MDict (`.mdx` + `.mdd`); lookup a word with image
      resources; verify rendered HTML and inline images.
- [x] Drag-reorder providers; close and reopen the popup — order
      persists; default-tab persists.
- [x] Disable Wiktionary; verify it disappears from the tab strip.
- [x] Delete an imported dictionary; verify files removed from
      `<AppData>/Dictionaries/<id>/` and the provider is gone from the
      popup.
- [x] `pnpm tauri dev` — repeat 2-6 to confirm the native file dialog
      accepts the new extensions and the Tauri 'Dictionaries' base dir is
      created.
- [x] iOS Tauri build — click "Read on Wikipedia →" / "Search for X on
      the web" links; verify they open in the system browser without the
      "Operation not permitted" error.
- [x] Sync: log in on a second device after metadata syncs; verify
      `unavailable: true` providers show a "Re-import" affordance and are
      hidden from the popup.
- [x] eink: with `?eink=true` URL flag, verify tabs and back button
      render without `not-eink:` styling artifacts.
- [x] `pnpm test --run` and `pnpm lint` — both clean.


mjeshurun:
Hi @chrox, thank you for making this feature happen. 
Will it be possible to add support for Babylon dictionary .BGL files for the custom dictionaries?

PR #4101, opened 2026-05-09
fix(library): parse author from filename and use edited metadata on fallback cover

### LINKED ISSUE #4095: Example: Imported novels do not display author and cover info correctly
Whether the imported novel displays the author is not determined by the novel name, but depends on whether the author information is present in the novel content.
Even if the author name is manually added in the app, it still won't show up on the novel cover.

Image
Image
Image
Image

chrox:
Closes #4095.

## Summary
- Add `parseFilenameMetadata` (`src/utils/path.ts`) and use it in `bookService.importBook` so common filename conventions are mined for title/author when the file's embedded metadata is missing them. Recognized patterns:
  - `《Title》`, `《Title》Author`, `《Title》[Author]`, `《Title》(Author)`, `《Title》【Author】`
  - `[Author] Title`, `(Author) Title`, `【Author】Title`
  - `Title [Author]`, `Title (Author)`, `Title 【Author】`
  - Plain `[Title]` / `(Title)` wrappers are unwrapped as the title
  - Ambiguous shapes like `A - B` are intentionally left as a single title
- `BookCover` fallback now reads `book.metadata?.author` when `book.author` is empty, so authors typed into the metadata edit dialog appear on auto-generated covers.
- `handleUpdateMetadata` no longer mutates the existing `Book` reference; it builds a new object before calling `updateBook`. Without this, `BookCover`'s `React.memo` comparator sees identical prev/next refs and skips the re-render right after a metadata edit.

## Test plan
- [x] `pnpm test` (4095 passed / 8 skipped)
- [x] `pnpm lint`
- [x] New unit tests:
  - `src/__tests__/utils/path.test.ts` covers all recognized filename patterns and the ambiguous-dash carve-out.
  - `src/__tests__/components/BookCover.test.tsx` covers the metadata-author fallback path.
- [ ] Manual: import a book whose file is named `《三体》刘慈欣.epub` with no embedded author and confirm both title and author show in the library.
- [ ] Manual: import a book with no author info, edit metadata to add an author, confirm the fallback cover updates immediately (no need to navigate away).

### DIFF
--- a/apps/readest-app/src/__tests__/components/BookCover.test.tsx
+++ b/apps/readest-app/src/__tests__/components/BookCover.test.tsx
@@ -38,4 +38,15 @@ describe('BookCover', () => {
     expect(img).toBeTruthy();
     expect(img?.getAttribute('loading')).toBe('lazy');
   });
+
+  it('falls back to metadata.author on the fallback cover when book.author is empty', () => {
+    const book = makeBook({
+      author: '',
+      coverImageUrl: undefined,
+      metadata: { author: 'Edited Author' } as Book['metadata'],
+    });
+    const { container } = render(<BookCover book={book} coverFit='crop' />);
+    const fallback = container.querySelector('.fallback-cover');
+    expect(fallback?.textContent).toContain('Edited Author');
+  });
 });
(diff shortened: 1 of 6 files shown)


chrox:
Approach was wrong — re-doing from scratch with the correct understanding (txt-conversion path, not generic epub filename parsing).

PR #4137, opened 2026-05-12
feat(reader): improve Japanese selection UX by disabling furigana selection

souzanshaquille:
When selecting and copying Japanese text with furigana (e.g., 門前払), the <rt> annotation text leaks into the clipboard: 門もん前ぜん払ばら. This creates visual clutter and can interfere with the selection "hitbox" for users trying to look up specific words, furthermore it can break translation to the furigana cutting in after the kanji.

Validation:

* Confirmed with the Japanese reading community (including native speakers) that this is the preferred behavior.
* Verified that this does not break the internal translation/copy logic
* Community feedback thread: https://www.reddit.com/r/japanese/comments/1ta1peb/question_for_japanese_readers_how_do_you_prefer/

Added `user-select: none` to the `rt` tag. The furigana remains visible but is ignored by the selection cursor, resulting in a cleaner highlight of the core content. The ruby transformer moves furigana text from DOM text nodes into CSS pseudo-elements (`::before` with content: `attr(data-text)`). Browsers only serialize text nodes during copy, so CSS-generated content is excluded from the clipboard while remaining fully visible on screen.


chrox:
Just adding the css here should be OK: 
https://github.com/readest/readest/blob/main/apps/readest-app/src/utils/style.ts#L297

No need for the ruby transformer.

chrox:
The annotator has already stripped the ruby when copying selected text.

chrox:
See here https://github.com/readest/readest/blob/main/apps/readest-app/src/app/reader/components/annotator/Annotator.tsx#L210

souzanshaquille:
The annotator strips rt for in-app actions, but native clipboard copy (Ctrl+C) does not. When I copy a word (like 先生) with rt annotations besides it, it stills copy what's inside it (先せん生せい). I think the default should not include the text in ruby.

chrox:
The copied text of `Ctrl+C` should also goes through `getAnnotationText` and gets the `rt` stripped, if not in your system we might need to fix that directly. It works as expected on macOS, iOS and Android, no `rt` text is copied with the system copy shortcuts.

PR #4153, opened 2026-05-13
Add namespace to koreader plugin modules to avoid collisions with other plugins

j-v:
The KOReader plugin was not showing when installed on my device, logging an error:
`Error when loading plugins/readest.koplugin/main.lua plugins/readest.koplugin/main.lua:19: attempt to call local '_' (a table value) `
The issue is there was another plugin that had a module called `i18n`, and when readest.koplugin loaded its own, it received the cached version from the other plugin, due to lua's `require` behavior.
The convention to avoid this kind of collision seems to be to add a namespace prefix to the module name - e.g https://github.com/AndyHazz/bookends.koplugin/commit/b8792e75383658be019cbfd224934668651b459a 
This change adds `readest_` prefix to the modules in root plugin folder.
The modules in the `library` directory are already kind of prefixed since they are imported using `require("library.*")` syntax, so I didn't change those, though it could be hardened by changing folder name to "readest_library"

Testing:
- Works fine in KOReader including syncing progress, annotations and library
- Unit tests pass


chrox (approved):

chrox:
Good catch. Thanks.

PR #4161, opened 2026-05-14
fix(koplugin): omit empty note field when syncing annotations

jenningsb2:
When syncing highlights between Readest and KOReader, the note field
was forced to an empty string ("") for annotations and bookmarks
without notes. KOReader's native annotations omit the field entirely
when no note exists, so the empty string caused KOReader to treat
every synced highlight as having a (blank) note. Apply the same
omission in both push and pull directions.


chrox:
@jenningsb2 Please resolve the conflicts.

PR #4174, opened 2026-05-15
feat(reader): import annotations from Moon+ Reader (.mrexpt)

loveheaven:
Add a new menu entry under the reader sidebar 'More' menu that lets users import highlights and notes exported from the Moon+ Reader Android app.

Implementation:

- utils/mrexpt.ts: parser for the .mrexpt plaintext format (entry id, NCX navPoint index b4, character offset b6, type marker, word and note).

- services/annotation/providers/mrexpt.ts: convert mrexpt entries to BookNote[] using bookDoc. Locate the chapter via b4 -> toc -> spine, then TreeWalker-search the section DOM for the highlighted word with English suffix tolerance (ing/ed/s/...). Falls back to a section-level CFI when the exact word can't be located. Re-imports are deduplicated by a stable id derived from entryId.

- BookMenu: add 'Import from Moon+ Reader' menu item dispatching the 'import-mrexpt' event.

- Annotator: handle 'import-mrexpt' — pick the file (Web File / Tauri path), parse, convert against the live bookDoc, merge into booknotes (latest updatedAt wins), persist via saveConfig, and apply to all live views so highlights appear immediately. User feedback via toasts (importing / imported N / N unmatched / nothing new).

image
image

### DIFF
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -1548,5 +1548,19 @@
   "Are you sure to clear all {{count}} highlights and notes?_other": "هل أنت متأكد من مسح التمييزات والملاحظات الـ{{count}}؟",
   "Discord": "Discord",
   "Leave blank to use the official Readwise API. Set a custom URL only to target a self-hosted, Readwise-compatible service.": "اتركه فارغًا لاستخدام واجهة Readwise الرسمية. عيّن عنوان URL مخصصًا فقط للاتصال بخدمة مستضافة ذاتيًا ومتوافقة مع Readwise.",
-  "Custom URL": "عنوان URL مخصص"
+  "Custom URL": "عنوان URL مخصص",
+  "Book is not ready yet, please try again.": "الكتاب غير جاهز بعد، يرجى المحاولة مرة أخرى.",
+  "Select Moon+ Reader Export File": "اختر ملف تصدير Moon+ Reader",
+  "Failed to read the selected file.": "فشل قراءة الملف المحدد.",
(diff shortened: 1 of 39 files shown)


chrox:
@loveheaven Please resolve the conflicts.

loveheaven:
Done. Please review it.

chrox:
Pushed two follow-up commits reworking the import UX:

**`refactor(reader): simplify Moon+ Reader import notifications`**
- Dropped the intermediate "Importing N annotations…" toast — the toast system shows one toast at a time, so it only flashed before the result toast replaced it.
- Removed the duplicate "Failed to read the selected file." toast in the read `catch` block; it now falls through to the existing empty-content check.
- Collapsed the three-way result toast (already imported / N unmatched / N imported) into one: `Imported {{count}} annotations` or `No new annotations to import`.
- Fixed a result-message bug: when every converted note was already imported and nothing was unmatched, the toast read "Imported 0 annotations." It now reports "No new annotations to import".
- Switched the success message to i18n `{{count}}` so it pluralizes (the previous `{{n}}` never did, e.g. "Imported 1 annotations").
- Extracted the dedupe/merge logic into a pure, unit-tested `mergeImportedBookNotes` helper (`src/__tests__/services/mrexpt-import.test.ts`).

Net: the import flow now shows exactly one toast per run instead of up to two.

**`chore(i18n): translate Moon+ Reader import strings`**
- Ran i18next extraction (the new strings hadn't been extracted) and translated all 9 new strings across the 33 locales — 340 keys total, with correct per-locale plural forms.

`pnpm test` (4261 passed) and `pnpm lint` are green.

PR #4175, opened 2026-05-15
feat(reader): add Clear All button to annotation/bookmark sidebar tabs

loveheaven:
Add a small 'Clear All' button at the top of each booknote list (annotations, bookmarks, excerpts) in the reader sidebar. The button shows a confirmation dialog (rendered via ModalPortal so it appears above the reader area), then soft-deletes every note of the active type by setting deletedAt, removes the corresponding overlays from all live views, persists the change via saveConfig, and resets the booknote navigation state.

image
image

### DIFF
--- a/apps/readest-app/src/app/reader/components/annotator/Annotator.tsx
+++ b/apps/readest-app/src/app/reader/components/annotator/Annotator.tsx
@@ -13,6 +13,7 @@ import { useBookDataStore } from '@/store/bookDataStore';
 import { useSettingsStore } from '@/store/settingsStore';
 import { useReaderStore } from '@/store/readerStore';
 import { useNotebookStore } from '@/store/notebookStore';
+import { useSidebarStore } from '@/store/sidebarStore';
 import { useCustomDictionaryStore } from '@/store/customDictionaryStore';
 import { useTranslation } from '@/hooks/useTranslation';
 import { useResponsiveSize } from '@/hooks/useResponsiveSize';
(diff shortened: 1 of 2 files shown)


chrox:
I’d prefer moving this action into the book menu since it’s not something users need frequently, and it doesn’t really justify having a persistent button in the annotations list. It also breaks the minimalist feel of the sidebar view.

loveheaven:
> I’d prefer moving this action into the book menu since it’s not something users need frequently, and it doesn’t really justify having a persistent button in the annotations list. It also breaks the minimalist feel of the sidebar view.

okay. I will move it to the book menu .
I need this function because I need to import annotations from MoonReader. If you want to support importing, you have to support clearing annotations sometimes.

apps/readest-app/src/app/reader/components/annotator/Annotator.tsx
```diff
@@ -927,6 +937,56 @@ const Annotator: React.FC<{ bookKey: string }> = ({ bookKey }) => {
...
+    if (updatedConfig) {
+      saveConfig(envConfig, bookKey, updatedConfig, settings);
+    }
+    // Reset any browse-mode state in the annotations sidebar tab so it
+    // doesn't keep paging through stale (now soft-deleted) entries.
+    clearBooknotesNav(bookKey);
+
+    eventDispatcher.dispatch('toast', {
+      type: 'info',
+      message: _('Cleared {{n}} highlights and notes.', { n: cleared }),
```
chrox: Use {{count}} so that i18n could extract plural format of the strings.

apps/readest-app/src/app/reader/components/annotator/Annotator.tsx
```diff
@@ -1082,6 +1142,21 @@ const Annotator: React.FC<{ bookKey: string }> = ({ bookKey }) => {
           onExport={handleConfirmExport}
         />
       )}
+      {clearAnnotationsCount > 0 && (
+        <ModalPortal>
+          <Alert
+            title={_('Clear Annotations')}
+            message={_('Are you sure to clear all {{n}} highlights and notes?', {
+              n: clearAnnotationsCount,
```
chrox: Same as above.

chrox (changes requested):

loveheaven:
image
image

now I moved it to the book menu.

chrox (approved):

PR #4177, opened 2026-05-15
feat(reader): custom hardware-button page turning

### LINKED ISSUE #4139: FR: Allow custom hardware buttons to faciltate use of page turners (accessability feature)

Readest currently does not support remote page turners which is an accessibility issue. 

I am trying to transition fully to Readest from Moon Reader. The largest impediment currently is the lack of support for page turners. I have a disability that makes it extremely difficult at times to read without the use of a page turner so this is extremely important accessibility feature and I know I am not alone in this. 

Currently, Readest only supports the use of volume up/down buttons to turn pages. Many page turners ([this](https://www.amazon.com/dp/B0B6RBHJFY?ref_=ppx_hzsearch_conn_dt_b_fed_asin_title_1) is the one I use) send other hardware commands such as previous song/next song or play/pause. 

Currently, the only way I am able to use Readest is by using a secondary utility (Key Mapper) to send volume up/down commands when the page turner is used. This is hacky and shouldn't be needed. 

Please add better support for page turners. Id like to see an additional section in the behavior menu that would allow the user to configure their page turner. 

This should be separate option from the existing "volume keys for page flip" toggle. It should allow users to turn on support for page turners then choose which hardware key command should trigger page up/down. 

**Additional context**
Here is a screenshot of how Moon Reader accomplishes this, as an example. 

Image

chrox:
## Summary

Adds a configurable **Page Turner** section to reader settings, letting users bind hardware remote keys to page/section navigation — an accessibility feature requested in #4139 for users of Bluetooth page-turner remotes.

- **Learn-mode capture UI**: tap "Set key", press a remote button, the key is captured and bound. Four bindable actions: Previous/Next Page and Previous/Next Section.
- **Key sources**: media keys (next/previous/play-pause) are captured natively — Android via `dispatchKeyEvent`, iOS via `MPRemoteCommandCenter`; D-pad/keyboard keys are captured via DOM `keydown` and book-iframe `iframe-keydown` messages (so bindings work as soon as a book is open).
- **All platforms**: the section shows on web, desktop, and mobile. Web/desktop bind keyboard keys only; native media-key interception stays mobile-only.
- The existing "Volume Keys for Page Flip" toggle is preserved, relocated into the same section.
- New strings translated across all 33 locales.

Pure key-normalization/matching logic lives in `src/utils/keybinding.ts` with unit tests.

Closes #4139

## Test plan

- [x] `pnpm test` — unit tests pass
- [x] `pnpm lint` — Biome + tsgo clean
- [x] `pnpm fmt:check` / `pnpm clippy:check` — Rust clean
- [x] Manual: Android device with a Bluetooth page turner — capture + page/section turn (media-key and D-pad remotes)
- [x] Manual: iOS device — D-pad/keyboard capture; media keys best-effort via `MPRemoteCommandCenter`
- [x] Manual: regression check that the volume-keys toggle and D-pad toolbar navigation still work
- [x] Manual: E-ink mode legibility of the Page Turner section


titandrive:
Youre amazing. That was so quick.

PR #4196, opened 2026-05-16
feat(readwise): allow overriding the Readwise sync base URL

### LINKED ISSUE #4114: FR: Allow advanced users to override the Readwise sync base URL

I’d like to use Readest’s existing Readwise sync/export support with a custom Readwise-compatible API endpoint. At the moment, the integration hardcodes the official Readwise API base URL, which makes it hard to self-host or build custom receivers for highlights and notes.

Add an advanced option to override the Readwise sync base URL in the readwise settings interface.

When unset, Readest should continue behaving exactly as it does today. When set, Readest would send Readwise sync/export requests to the custom URL instead, while keeping the same Readwise-compatible API contract and auth flow where possible.

This could be hidden under advanced settings so it does not affect or confuse normal users.

**Additional context**

This would enable advanced/self-hosted workflows where users run their own Readwise-compatible service to receive highlights, notes, and annotations from Readest, then forward or transform them into systems like Obsidian, a database etc.

The default behaviour would remain unchanged for everyone who does not configure a custom URL.

chrox:
## Summary

Adds an advanced option to point Readwise sync/export at a custom, Readwise-compatible endpoint instead of the hardcoded official API. When the override is unset or blank, behavior is unchanged for everyone.

Closes #4114

## Changes

- **`ReadwiseClient`** resolves a custom `baseUrl` over `READWISE_API_BASE_URL`, trimming whitespace and trailing slashes so `${baseUrl}${endpoint}` joins cleanly. Falls back to the official endpoint when unset or blank.
- **`ReadwiseSettings`** gains an optional `baseUrl` field. It syncs across devices as plaintext via the settings sync whitelist (it's a URL, not a credential, so it's not in `encryptedFields`).
- **`ReadwiseForm`** exposes the URL under a collapsed **Advanced** disclosure on the connect screen, with a `Tips` note. Once connected, a custom URL is surfaced read-only in the card. Disconnect preserves the custom URL so reconnecting is friction-free.
- i18n: extracted 2 new strings and translated them across all 33 locales.

## Testing

- New `ReadwiseClient.test.ts` (test-first): default URL, custom URL, trailing-slash normalization, blank fallback, and POST auth header — 5 tests.
- `pnpm test` — 4231 passed, 8 skipped.
- `pnpm lint` — Biome + tsgo clean.

### DIFF
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -1546,5 +1546,7 @@
   "Are you sure to clear all {{count}} highlights and notes?_few": "هل أنت متأكد من مسح التمييزات والملاحظات الـ{{count}}؟",
   "Are you sure to clear all {{count}} highlights and notes?_many": "هل أنت متأكد من مسح التمييزات والملاحظات الـ{{count}}؟",
   "Are you sure to clear all {{count}} highlights and notes?_other": "هل أنت متأكد من مسح التمييزات والملاحظات الـ{{count}}؟",
-  "Discord": "Discord"
+  "Discord": "Discord",
+  "Sync Base URL": "عنوان URL الأساسي للمزامنة",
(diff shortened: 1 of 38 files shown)


momadacoding:
新版本还没有这个功能是吧？

chrox:
It will be available in the next update. Please checkout the release note.

Issue #4108, opened 2026-05-09
SonarQube Findings


iam-mj:
Hello!
I've been looking into static analysis tools lately, and so happened to run [SonarQube](https://www.sonarsource.com/products/sonarqube/) on the Readest repo. It has surfaced some minor improvements which could be brought to the codebase, mostly regarding logic and readability. I thought I could just as well let you know about them.

 ### Identified Issues

It has highlighted multiple issues, but the more relevant are:
* unused variable: `duration` in TTS's [onStartCommand](https://github.com/readest/readest/blame/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src-tauri/plugins/tauri-plugin-native-tts/android/src/main/java/MediaPlaybackService.kt#L245)
* conditional expression always returns the same value: 
   * in [txt-converter's test file](https://github.com/readest/readest/blob/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src/__tests__/utils/txt-converter.test.ts#L59)
   * in [NotebookToggler](https://github.com/readest/readest/blob/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src/app/reader/components/NotebookToggler.tsx#L38)
* using return value of void function: in [AnnotationNotes](https://github.com/readest/readest/blob/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src/app/reader/components/annotator/AnnotationNotes.tsx#L100)
* unexpected `await` on non-Promise value: 
   * in the [useFileSelector hook](https://github.com/readest/readest/blob/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src/hooks/useFileSelector.ts#L92)
   * in the [style transformer](https://github.com/readest/readest/blob/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src/services/transformers/style.ts#L15)
   * in the [EdgeTTSClient](https://github.com/readest/readest/blob/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src/services/tts/EdgeTTSClient.ts#L219)
   * in the [TTSController](https://github.com/readest/readest/blob/295a588988af21a2db0759fb923bedb1e6c37923/apps/readest-app/src/services/tts/TTSController.ts#L257)
[...]

PR #4201, opened 2026-05-17
Fix sonarqube findings

iam-mj:
## Changes
Started by addressing the matters discussed in issue #4108:
* remove a few unnecessary `await`s
* add one `await` which was missing on a Promise
* simplify conditionals that return the same value in both cases
* remove a duplicated type declaration

When running `pnpm test` afterwards, 4 tests in `node-app-service.test.ts` were failing due to a mismatch in some expected vs received paths. The mismatch was coming from the expected path using Windows' `/` path separator, while the received path used POSIX's `\` path separator. Therefore, I added a second commit to:
* refactor NodeAppService and AppService methods to not have path concatenation hardcoded, but resolved using node's `path.join`; this achieves platform-independent path resolution and enables unit tests to pass on Windows as well

## Testing
* `pnpm build`
* `pnpm test`

### DIFF
--- a/apps/readest-app/src/services/appService.ts
+++ b/apps/readest-app/src/services/appService.ts
@@ -23,6 +23,7 @@ import { CustomTextureInfo } from '@/styles/textures';
 import { CustomFont, CustomFontInfo } from '@/styles/fonts';
 import type { ImportedDictionary } from './dictionaries/types';
 import type { SelectedFile } from '@/hooks/useFileSelector';
+import * as nodePath from 'node:path';
 
 import * as BookSvc from './bookService';
 import * as CloudSvc from './cloudService';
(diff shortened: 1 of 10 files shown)


apps/readest-app/src/services/appService.ts
```diff
@@ -23,6 +23,7 @@ import { CustomTextureInfo } from '@/styles/textures';
 import { CustomFont, CustomFontInfo } from '@/styles/fonts';
 import type { ImportedDictionary } from './dictionaries/types';
 import type { SelectedFile } from '@/hooks/useFileSelector';
+import * as nodePath from 'node:path';
```
chrox, line 26: We cannot introduce the `node:path` here, since `appService` will be imported in browser environment.

chrox (changes requested):

chrox:
This PR breaks some fundamental CI requirements. Please run the unit tests locally before submitting further changes.

Also, GPT-5.4/5.5 and Claude Opus 4.6/4.7 should be able to identify and handle these issues properly, so please make sure the changes are reviewed and refined with those models before updating the PR.

PR #4203, opened 2026-05-17
fix(tests): materialize zip.js blob before wrapping in File for case-mismatch fixture

loveheaven:
The case-mismatch EPUB fixture builds an archive with @zip.js/zip.js' BlobWriter and then wraps the resulting Blob into a File:

  const blob = await writer.close();
  new File([blob], 'case-mismatch.epub', ...);

Under vitest's happy-dom/jsdom, the File/Blob polyfill does not correctly pull bytes out of a nested Blob part produced by zip.js. The outer File reports a non-zero size, but the bytes BlobReader sees in DocumentLoader (libs/document.ts: 'new BlobReader(this.file)') are not a valid ZIP — getEntries() yields nothing, open() falls through with book = null, and the test crashes at:

  TypeError: Cannot read properties of null (reading 'sections')

Materialize the zip bytes into a plain ArrayBuffer first, then construct the File from that. ArrayBuffer parts go through the polyfill cleanly because they don't require recursive Blob unwrapping, so zip.js reads a real archive and the test passes:

  const arrayBuffer = await blob.arrayBuffer();
  new File([arrayBuffer], 'case-mismatch.epub', ...);

This brings the fixture in line with the rest of the test suite (paginator-expand, page-progress-epub, toc-cfi-mapping, ...) which already use ArrayBuffer-based File construction. No production code is affected: real browsers handle nested-Blob File construction correctly.

### DIFF
--- a/apps/readest-app/src/__tests__/document/document-loader-epub-case.test.ts
+++ b/apps/readest-app/src/__tests__/document/document-loader-epub-case.test.ts
@@ -47,7 +47,8 @@ const createCaseMismatchEpub = async () => {
   );
 
   const blob = await writer.close();
-  return new File([blob], 'case-mismatch.epub', { type: 'application/epub+zip' });
+  const arrayBuffer = await blob.arrayBuffer();
+  return new File([arrayBuffer], 'case-mismatch.epub', { type: 'application/epub+zip' });
 };
 
 describe('DocumentLoader EPUB zip lookup', () => {


loveheaven:
pnpm test always fails on my macbook pro. so I tried to fix it.

chrox:
It works on my dev machine for all macOS, Linux and Windows 10 and on GitHub CI. I'm curious why it failed on your machine. Did you use a node version other than 24?

chrox:
Reviewed and merging — the fix is good. One note for the record.

**The fix is correct and worth taking.** `await blob.arrayBuffer()` reads the zip bytes via the blob's *own* implementation, then `new File([arrayBuffer], ...)` hands the `File` constructor a plain primitive every implementation reads identically. No cross-implementation Blob unwrapping. It also matches how every other EPUB fixture in the suite builds its `File`.

**That said, the "before" crash isn't reproducible on `main` with the committed config.** The test passes with the old `new File([blob], ...)` code on both GitHub CI (`node-version: 24`) and local macOS (Node 24.14), under the repo's actual `vitest.config.mts` (`environment: 'jsdom'`, jsdom 28.1).

Root cause of the discrepancy: the crash needs an **implementation mismatch** between the Blob and the File constructor. A `Blob`'s bytes live in an internal slot only its own implementation can read. If zip.js's `BlobWriter` produces a Blob from implementation A while `File` is implementation B, `new File([blobA])` yields a File of the correct `.size` but **zero readable bytes** — exactly the described symptom (`getEntries()` empty → `book = null` → `Cannot read properties of null (reading 'sections')`).

In the repo's jsdom config there's no mismatch: jsdom installs one `Blob`/`File` pair as globals and zip.js picks up that same jsdom `Blob` (`blob instanceof globalThis.Blob === true`), so jsdom's `File` unwraps it cleanly. The mismatch only appears under a different setup — e.g. **happy-dom** (ships its own Blob/File), or an older jsdom that didn't register `Blob` globally so zip.js fell back to Node's native `buffer.Blob`. The PR description mentions "happy-dom/jsdom"; the committed config is jsdom-only.

So this lands as **defensive hardening** rather than a fix for a bug reproducible on `main` today — it makes the fixture immune to which DOM env / Node version a future contributor runs. Good change, merging. Consider tweaking the PR description so it doesn't imply the old code crashes universally.

PR #4204, opened 2026-05-17
feat(integrations): add WebDAV sync to Reading Sync settings

loveheaven:
Adds a WebDAV entry under Settings -> Integrations -> Reading Sync with configure/browse UI, library-wide Sync now, and per-book sync of progress, annotations and (opt-in) book files + covers.

Reading progress and annotations are always synced when WebDAV is enabled; only Sync Book Files stays as a toggle since it's bandwidth-heavy.
image
image
image
image

### DIFF
--- a/apps/readest-app/src/__tests__/services/webdav-connect-settings.test.ts
+++ b/apps/readest-app/src/__tests__/services/webdav-connect-settings.test.ts
@@ -1,0 +1,107 @@
+import { describe, test, expect } from 'vitest';
+import { buildWebDAVConnectSettings } from '@/services/webdav/webdavConnectSettings';
+import type { WebDAVSettings, WebDAVSyncLogEntry } from '@/types/settings';
+
+describe('buildWebDAVConnectSettings', () => {
+  test('applies form fields onto a blank previous state', () => {
+    const result = buildWebDAVConnectSettings(undefined, {
+      serverUrl: '  https://dav.example.com  ',
+      username: 'alice',
+      password: 'hunter2',
+      rootPath: '/Readest',
+    });
+    expect(result).toEqual({
+      enabled: true,
+      serverUrl: 'https://dav.example.com',
+      username: 'alice',
+      password: 'hunter2',
+      rootPath: '/Readest',
+    });
+  });
+
+  test('preserves prior bookkeeping fields across reconnect', () => {
+    // Simulates the disconnect → reconnect flow: the user previously
+    // synced (deviceId minted, syncBooks toggled on, history populated),
+    // disabled WebDAV, and is now reconnecting with the same credentials.
+    const log: WebDAVSyncLogEntry[] = [
+      {
+        id: 'log-1',
+        startedAt: 1_700_000_000_000,
+        finishedAt: 1_700_000_001_500,
+        status: 'success',
+        trigger: 'manual',
+        totalBooks: 3,
+        booksDownloaded: 0,
+        filesUploaded: 1,
+        filesAlreadyInSync: 2,
+        configsUploaded: 3,
+        configsDownloaded: 0,
+        coversUploaded: 0,
+        failures: 0,
+        summary: 'Sync complete',
+      },
+    ];
+    const previous: WebDAVSettings = {
+      enabled: false,
+      serverUrl: 'https://dav.example.com',
+      username: 'alice',
+      password: 'hunter2',
+      rootPath: '/Readest',
+      syncProgress: true,
+      syncNotes: true,
+      syncBooks: true,
+      strategy: 'send',
+      deviceId: 'device-uuid-9f3c',
+      lastSyncedAt: 1_700_000_001_500,
+      syncLog: log,
+    };
+
+    const next = buildWebDAVConnectSettings(previous, {
+      serverUrl: 'https://dav.example.com',
+      username: 'alice',
+      password: 'hunter2',
+      rootPath: '/Readest',
+    });
+
+    expect(next.enabled).toBe(true);
+    // Stable per-device id MUST survive — losing it makes the next sync
+    // look like a brand-new device and breaks cross-device clobber
+    // detection in `RemoteBookConfig.writerDeviceId`.
+    expect(next.deviceId).toBe('device-uuid-9f3c');
+    expect(next.syncBooks).toBe(true);
+    expect(next.strategy).toBe('send');
+    expect(next.syncProgress).toBe(true);
+    expect(next.syncNotes).toBe(true);
+    expect(next.lastSyncedAt).toBe(1_700_000_001_500);
+    expect(next.syncLog).toEqual(log);
+  });
+
+  test('updates the credentials when the user reconnects to a different account', () => {
+    const previous: WebDAVSettings = {
+      enabled: false,
+      serverUrl: 'https://old.example.com',
+      username: 'alice',
+      password: 'old-pw',
+      rootPath: '/Old',
+      deviceId: 'device-keep',
+      syncBooks: false,
+    };
+    const next = buildWebDAVConnectSettings(previous, {
+      serverUrl: 'https://new.example.com/',
+      username: 'bob',
+      password: 'new-pw',
+      rootPath: '/New',
+    });
+    expect(next.serverUrl).toBe('https://new.example.com/');
+    expect(next.username).toBe('bob');
+    expect(next.password).toBe('new-pw');
+    expect(next.rootPath).toBe('/New');
+    // The deviceId is intentionally NOT rotated even when the user
+    // reconnects to a different server/account: it identifies the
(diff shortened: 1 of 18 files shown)


chrox:
Nice work. Before merging this, I have a few concerns:

1. Does this also work in the web version?
2. Are there any diagnostic tools for investigating sync failures? Otherwise, this could easily result in a large amount of user feedback and difficult-to-debug issues.
3. How are conflicts resolved when both Readest Sync and WebDAV sync modify the same config.json for a book?
4. What happens when two devices come back online with different versions of the same config.json? How are overwrites and conflict resolution handled in that case?

loveheaven:
> Nice work. Before merging this, I have a few concerns:
> 
> 1. Does this also work in the web version?

Mostly work in the web version. It only doesn't work when you try to download a single book. It will show a clear toast on web that this feature is only for apps.
**Bottom line**: progress / annotations / library index sync work on web given a CORS-friendly server; book-binary sync works for small files but degrades for large ones; native streaming and the per-file download button are Tauri-only.


> 2. Are there any diagnostic tools for investigating sync failures? Otherwise, this could easily result in a large amount of user feedback and difficult-to-debug issues.

Yes — this was a real concern, so I added a sync history panel in the second commit.
image
This can help users know what exactly happened.

What's not there: a "Run Diagnostics" button that probes PROPFIND/PUT/HEAD against the configured root and prints each step's result. That's a natural P2 follow-up — it'd cut down on "my server returns 200 for HEAD but 405 for PUT" tickets — but the existing log is enough for users to file a useful bug report from a screenshot.

> 3. How are conflicts resolved when both Readest Sync and WebDAV sync modify the same config.json for a book?

Honest answer: the two backends don't coordinate, but they will rarely collide. They separately writes each other's own server. Webdav does not sync ViewSettings. It only sync progress, booknotes and books. so the collision is very rare. They will collide only when 
- Device A pushes to readest cloud successfully but fails to push to WebDAV (network blip).
- Device B comes online and pulls both. If WebDAV's writerUpdatedAt happens to read as newer than the cloud's updatedAt — possible because they live in different clock domains; WebDAV records the writer's wall clock, the cloud uses server time — the stale WebDAV payload could win and roll Device B's progress back by one push.

But this is rare. if you want, I can add a message on webdav settings to tell users that you should choose only one sync between webdav and readest cloud sync.


> 4. What happens when two devices come back online with different versions of the same config.json? How are overwrites and conflict resolution handled in that case?

Resolution is last-writer-wins by updatedAt at the top level, with separate per-note merge for booknotes.
Concretely:
Reading position (progress / location / xpointer) is whole-payload LWW. If A reads to 70% offline and B reads to 65% offline, whoever's updatedAt is later wins. The other side's progress is overwritten. This matches what most users expect for reading position — your phone session at midnight should be visible on the desktop in the morning.
Booknotes use field-level merge keyed by note id (mergeNotes). For each id, the side with the larger updatedAt (or deletedAt, for tombstones) wins. So concurrent annotations from A and B both survive a round-trip; one device deleting a note while the other edits it resolves to whichever timestamp is later.
viewSettings and other device-local preferences are not on the wire at all (see #3), so they cannot collide.
The wire envelope tags every payload with writerDeviceId and writerVersion, and the pull result surfaces remoteDeviceId. We capture this in the diagnostic Sync history (in the per-book failure entries) but don't currently surface it in any user-facing UI.

loveheaven:
What do you think of this pr? Do I need to make more improvement?

chrox:
It's generally fine. And I will test it more to finalize the PR.

ybbluesky:
good，期待早日集成进程序中,与书库集成了吗？

loveheaven:
Conflict resolved.

chrox:
Did a review pass and pushed one fix commit on top of your branch (also rebased onto current `main` to pick up the recent send/library fixes).

Two bugs caught:

1. **`WebDAVForm.handleConnect` wiped prior webdav state on reconnect.** The merge rebuilt the whole `webdav` block from the four form fields, dropping `deviceId`, `syncBooks`, `strategy`, `syncProgress`, `syncNotes`, `lastSyncedAt`, and `syncLog`. The deviceId rotation was the worst part — a disconnect + reconnect made the next sync look like a brand-new device and defeated the cross-device clobber check encoded in `RemoteBookConfig.writerDeviceId`. Extracted a pure `buildWebDAVConnectSettings` helper that spreads previous state first, matching the sibling pattern in `KOSyncForm.handleConnect`.

2. **`useWebDAVSync.ensureDeviceId` used a stale closure** to merge the freshly-minted deviceId back into settings, which can clobber a sibling write that landed between renders (`updateLastSyncedAt` and `persistWebdav` already use `useSettingsStore.getState()` to avoid this — `ensureDeviceId` now matches).

Added 3 unit tests covering the reconnect-preservation invariant (`webdav-connect-settings.test.ts`). Bug-reproduction trace: the tests fail against the original inline logic and pass after the spread fix.

Tests: 4400 passed / 0 failed. Lint: clean.

Nice work on the overall feature — the inline rationale comments make the trickier bits (Tauri-side streaming download, viewSettings invariant, closure-staleness pattern, regex XML parsing) easy to review. Some smaller observations I didn't push fixes for:

- Deleted books leak on the remote (the index propagation works for hiding them, but the per-hash dir is never `DELETE`d). Worth a TODO.
- `OPEN_PULL_SKIP_MS` comment claims it catches close-then-reopen, but the hook unmounts on close so `lastPulledAtRef` resets — the skip only triggers on same-instance re-runs.
- `WebDAVClient.encodePath` will re-encode literal `%` characters despite the "leave existing escapes alone" comment.
- `WebDAVForm.tsx` at 1444 lines is getting long; might be worth extracting the browser pane separately.

loveheaven:
> Did a review pass and pushed one fix commit on top of your branch (also rebased onto current `main` to pick up the recent send/library fixes).
> 
> Two bugs caught:
> 
> 1. **`WebDAVForm.handleConnect` wiped prior webdav state on reconnect.** The merge rebuilt the whole `webdav` block from the four form fields, dropping `deviceId`, `syncBooks`, `strategy`, `syncProgress`, `syncNotes`, `lastSyncedAt`, and `syncLog`. The deviceId rotation was the worst part — a disconnect + reconnect made the next sync look like a brand-new device and defeated the cross-device clobber check encoded in `RemoteBookConfig.writerDeviceId`. Extracted a pure `buildWebDAVConnectSettings` helper that spreads previous state first, matching the sibling pattern in `KOSyncForm.handleConnect`.
> 2. **`useWebDAVSync.ensureDeviceId` used a stale closure** to merge the freshly-minted deviceId back into settings, which can clobber a sibling write that landed between renders (`updateLastSyncedAt` and `persistWebdav` already use `useSettingsStore.getState()` to avoid this — `ensureDeviceId` now matches).
> 
> Added 3 unit tests covering the reconnect-preservation invariant (`webdav-connect-settings.test.ts`). Bug-reproduction trace: the tests fail against the original inline logic and pass after the spread fix.
> 
> Tests: 4400 passed / 0 failed. Lint: clean.
> 
> Nice work on the overall feature — the inline rationale comments make the trickier bits (Tauri-side streaming download, viewSettings invariant, closure-staleness pattern, regex XML parsing) easy to review. Some smaller observations I didn't push fixes for:
> 
**Thank you for your professional review. 
I added 6 commits above your commit to resolve the issues you mentioned:**

> * Deleted books leak on the remote (the index propagation works for hiding them, but the per-hash dir is never `DELETE`d). Worth a TODO.


The 5th commit [2cb75e4] resolves this issue. I add manual cleanup process to clean these dirs.

> * `OPEN_PULL_SKIP_MS` comment claims it catches close-then-reopen, but the hook unmounts on close so `lastPulledAtRef` resets — the skip only triggers on same-instance re-runs.


The first commit [90ddcf1] above your commit  resolved  this issue.

> * `WebDAVClient.encodePath` will re-encode literal `%` characters despite the "leave existing escapes alone" comment.

The first commit [90ddcf1] above your commit  resolved  this issue.

> * `WebDAVForm.tsx` at 1444 lines is getting long; might be worth extracting the browser pane separately.

The second commit [cb3bc83] above your commit resolved this issue.

The 3rd commit [d0753a3] resolves bugs for oom of large files syncing .
The 4th commit [e5e8489] resolves bugs for breaking sync process if you navigate to other pages.
The 6th commit [ea52486] adds some tests.


Please review this PR again. And I have a question: do you think  if syncHistoryPanel is necessary or not? I am thinking to remove it. but maybe it should be done in another PR.

chrox:
The syncHistoryPanel is still useful as a debug tool while this feature is stabilizing and before we resolve most of the compatibility issues.

loveheaven:
> The syncHistoryPanel is still useful as a debug tool while this feature is stabilizing and before we resolve most of the compatibility issues.

Please review it. I think it is stable now.

PR #4226, opened 2026-05-19
fix(library): clear nested-folder groups when deleting from bookshelf

loveheaven:
Deleting a group from the bookshelf right-click menu used to leave the group on screen whenever the import had any sub-directories. The cause: getBooksToDelete matched only `book.groupId === id`, but the bookshelf renders a top-level group with id = md5("MyDir") while books imported from a sub-folder carry groupId = md5("MyDir/sub"). Sub-folder books never got marked for deletion, refreshGroups re-built the parent group from their groupName on the next render, and the user saw an undeletable folder.

Fix: when an id resolves to a known group via getGroupName, also collect every book whose groupName equals that path or starts with `${path}/`. Hash-based dedup keeps a book from being queued twice when both rules match. Single-book deletes and flat-folder group deletes are unaffected.
image

### DIFF
--- a/apps/readest-app/src/app/library/components/Bookshelf.tsx
+++ b/apps/readest-app/src/app/library/components/Bookshelf.tsx
@@ -36,6 +36,7 @@ import {
   createWithinGroupSorter,
   ensureLibraryGroupByType,
   ensureLibrarySortByType,
+  expandBookshelfSelection,
   getBookSortValue,
   getGroupSortValue,
   compareSortValues,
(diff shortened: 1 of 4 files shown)


loveheaven:
Before my fix, it is impossible to delete a nested-folder group

apps/readest-app/src/app/library/components/Bookshelf.tsx
```diff
@@ -361,11 +361,38 @@ const Bookshelf: React.FC<BookshelfProps> = ({
...
+      // than `md5(groupName)` — without this prefix sweep the parent
+      // group could never be cleared because deleting "MyDir" would
+      // leave "MyDir/sub/book" untouched, and `refreshGroups` would
+      // re-create the parent group on the next render.
+      const groupName = getGroupName(id);
+      if (groupName) {
+        const prefix = `${groupName}/`;
+        filteredBooks.forEach((book) => {
+          if (!book.groupName) return;
+          if (book.groupName === groupName || book.groupName.startsWith(prefix)) {
```
chrox: This is too tricky. It is better to collect all the group ids or book ids into the `bookIdsToDelete` when selecting the whole group.

chrox (changes requested):

chrox:
I pushed a refactor commit on top of your fix that addresses the review comment. The bug fix itself is unchanged — what moved is *where* groups get expanded into their constituent book hashes:

- New helper `expandBookshelfSelection(ids, items)` in `libraryUtils.ts` resolves a group id to its rendered `group.books` rollup. `generateBookshelfItems` already folds nested-folder books into the top-level group, so this picks up your "MyDir/sub" case without a path-prefix sweep.
- The select-mode delete (`Bookshelf.deleteSelectedBooks`) runs the selection through the helper before populating `bookIdsToDelete`.
- The right-click group delete in `BookshelfItem` now dispatches `group.books`' hashes directly — the contextual menu already has the rollup in hand.
- `getBooksToDelete` collapses back to a flat `Set` lookup on book hashes — no `getGroupName` call, no prefix matching, no dedup bookkeeping inside the deletion path.

I also added unit tests for `expandBookshelfSelection` covering the nested-rollup case, dedup, soft-deleted skipping, and pass-through of unknown ids. `pnpm test` (4395 passed) and `pnpm lint` (tsgo + biome) are clean.

Thanks for tracking down the bug — let me know if you'd prefer a different shape.

chrox (approved):

PR #4229, opened 2026-05-19
feat(library): add Import from Folder dialog with format/size filters

loveheaven:
Replaces the silent "import every supported file recursively" behaviour of the directory import menu item with an explicit dialog that lets users pick which formats to include, set a minimum file size, and choose between mirroring subfolders as nested groups (legacy behaviour) or flattening every match into the current library view.

The selected directory is persisted in localStorage so re-opening the dialog seeds the path field with the user's last choice. Cancelling the dialog does not write to storage so an aborted pick won't pollute the next session.

Also hides the native number-input spinner via a small .no-spinner utility in globals.css; on macOS WebKit the spin buttons were drawing over the rounded input border and looked broken. The KB suffix now lives inside the input's bordered shell instead of beside it.
Two correctness fixes the dialog flow exposed:
    
The library importer treats groupId as a tri-state — undefined means "derive from 
basePath", '' means "explicitly the library root", any other string means a specific gro
up. Previously a falsy check conflated '' with undefined, so re-importing a deduped book
 under flatten mode silently kept its stale groupId/groupName from the prior keep-as-gro
ups run, making the book reappear in the old subfolder group instead of moving into the 
library root.
    
Imports of arbitrary user paths (e.g. ~/Downloads) now go through a new allow_path
s_in_scopes Tauri command that extends both fs_scope and asset_protocol_scope. The dialo
g plugin only auto-grants fs_scope, so reads through the asset protocol (RemoteFile / co
nvertFileSrc) used to fail with "asset protocol not configured to allow the path". The s
him is invoked after every selectFiles / selectDirectory call and once more at the start
 of runFolderImport so localStorage-restored paths are also covered. Granted scopes pers
ist across restarts via tauri_plugin_persisted_scope.

image
if you choose this option, it will behave like before.
image

image

if you choose the second option, it will recursively put all the books into bookshelf.
image
image

### DIFF
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -1608,5 +1608,18 @@
   "System": "النظام",
   "System dictionary integration is coming soon on this platform.": "تكامل قاموس النظام قادم قريباً على هذه المنصة.",
   "Waiting to be processed": "في انتظار المعالجة",
-  "Processing…": "جارٍ المعالجة…"
+  "Processing…": "جارٍ المعالجة…",
+  "Folder": "مجلد",
+  "Choose a folder": "اختر مجلداً",
+  "File Formats": "صيغ الملفات",
+  "File size larger than": "حجم الملف أكبر من",
+  "Minimum file size (KB)": "الحد الأدنى لحجم الملف (كيلوبايت)",
+  "KB": "كيلوبايت",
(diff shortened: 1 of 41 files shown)


loveheaven:
I developed this feature because I usually like putting all the books under one folder and import all of them into bookshelf.

loveheaven:
Please review this PR. I have refactored it.

PR #4244, opened 2026-05-20
fix(sel): clamp declared fontSize when it disagrees with rendered height

loveheaven:
The macOS system-dictionary HUD samples the underlying paragraph's typography via getRangeTextStyleInWebview so AppKit can re-draw the small label using the same font / size as the page text. The sampler trusted getComputedStyle().fontSize directly, which works for the typical EPUB inline box but breaks badly on pdf.js text layers: each glyph span carries an intrinsic font-size that reflects the document's unit-em size before transform: scale(...) shrinks it back to page- coordinate pixels, so the value can be many times larger than the on-screen glyph. Forwarded as-is to NSFont, that gives AppKit a giant attributed string and the yellow highlight rectangle behind the HUD ends up engulfing neighbouring paragraphs while the laid-out text overflows off-screen.

Cross-check the declared size against range.getBoundingClientRect(). height as a sanity bound. When the declared value exceeds the inline box height by more than 30 %, fall back to renderedHeight * 0.85 (roughly the cap-height-to-1.2-line-height ratio) so PDF lookups converge on a sane scale; otherwise keep the declared value untouched so normal EPUB body text is unaffected.
image

### DIFF
--- a/apps/readest-app/src/utils/sel.ts
+++ b/apps/readest-app/src/utils/sel.ts
@@ -162,9 +162,36 @@ export const getRangeTextStyleInWebview = (range: Range): RangeTextStyle | null 
     if (Number.isFinite(sy)) scaleY = sy!;
   }
 
-  const fontSizePx = parseFloat(style.fontSize) || 0;
+  // Cross-check the declared font-size against the range's actual
+  // visual height. In typical EPUB layout the inline box is roughly
+  // `font-size × line-height` tall (≈1.2× by default), so a declared
+  // font-size noticeably *larger* than the rendered height means the
+  // value isn't a real CSS pixel measurement we can hand to NSFont —
+  // pdf.js's text layer is the canonical offender: each glyph span
+  // can carry an intrinsic `font-size` that reflects the document's
+  // unit-em size before `transform: scale(...)` shrinks it back to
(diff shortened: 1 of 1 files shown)


loveheaven:
If I don't fix this bug, the yellow rectangle will be huge when the book is a pdf.

chrox:
image

@loveheaven Could you also fix the traffic lights position on macOS 26? There is a new trafficlightposition config in tauri now: https://v2.tauri.app/reference/config/#trafficlightposition, probably we don't need to rely on `apps/readest-app/src/hooks/useTrafficLight.ts` to make it align vertically in the header bar of a height of 48 pixels. As I don't have a macOS 26 device right now, I cannot test and verify it.

loveheaven:
> image
> @loveheaven Could you also fix the traffic lights position on macOS 26? There is a new trafficlightposition config in tauri now: https://v2.tauri.app/reference/config/#trafficlightposition, probably we don't need to rely on `apps/readest-app/src/hooks/useTrafficLight.ts` to make it align vertically in the header bar of a height of 48 pixels. As I don't have a macOS 26 device right now, I cannot test and verify it.

sure. no problem. I will try it.

chrox (approved):

PR #4247, opened 2026-05-20
fix(macos): fix traffic lights position on macOS 26

loveheaven:
Replaces the cocoa private-API positioning that drove traffic light placement through IPC with Tauri's supported trafficLightPosition window option, which routes through wry's macOS API and stays correct across versions including macOS 26 (Tahoe).

Position is now declared once at window creation: WebviewWindowBuilder.traffic_light_position in src-tauri/src/lib.rs for the initial main window, and trafficLightPosition on new WebviewWindow(...) in utils/nav.ts for reader windows and the recreated main window. The reader path mattered — those windows used to rely on the cocoa hack to place buttons after the on_window_ready hook fired, so any path that bypassed it left the buttons in AppKit's overlay default position (off-screen on macOS 26 until a resize).

The IPC surface narrows accordingly. set_traffic_lights now takes only visible: position is no longer a parameter and the WINDOW_CONTROL_PAD_X/Y static muts go away; setTrafficLightVisibility drops its position arg in trafficLightStore; useTrafficLight and HeaderBar drop their hard-coded { x: 10, y: 20 } magic numbers. position_traffic_lights stops touching the per-button NSWindowButton frames entirely and only collapses or restores the title-bar container view to hide / show buttons during reader chrome auto-hide. A short-circuit on the no-op transition keeps the cocoa setFrame from racing AppKit's own traffic-light tracking on every IPC call.

useTrafficLight stays — it still owns full-screen visibility synchronisation, the auto-hide visibility toggle, and feeds isTrafficLightVisible to the self-drawn <WindowButtons /> in the auth, library, OPDS, reader-sidebar, and user headers. None of those have an equivalent in the new declarative API. Only its 'where do the buttons sit' responsibility was moved out.

A single named constant TRAFFIC_LIGHT_RESTORE_Y_INSET is left behind in traffic_light.rs, used solely by the visible: false → true restore path to recompute the title-bar container height. It must agree with the y component of the two declarative trafficLightPosition values; a doc comment makes that contract explicit. Caching each window's natural title-bar height before the first collapse would let us delete the constant entirely, but the per-window state machine that requires is not worth the win for a single number.

y is tuned by eye to 24 to vertically center the buttons inside readest's ~48px header bar on macOS 26.1.
image
image

### DIFF
--- a/apps/readest-app/src-tauri/src/lib.rs
+++ b/apps/readest-app/src-tauri/src/lib.rs
@@ -14,6 +14,8 @@ mod android;
 
 use tauri::utils::config::BackgroundThrottlingPolicy;
 #[cfg(target_os = "macos")]
+use tauri::LogicalPosition;
+#[cfg(target_os = "macos")]
 use tauri::TitleBarStyle;
 
 use std::path::PathBuf;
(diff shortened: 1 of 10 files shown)


loveheaven:
Is this PR what you want?

chrox:
Looks great now. Thanks.

chrox:
@loveheaven I updated this PR to get rid of the hardcoded y padding for each macOS versions. It now works as expected on macOS 15.6, could you test it on macOS 26 and see if there is any unexpected behavior other than your original implementation?

loveheaven:
> @loveheaven I updated this PR to get rid of the hardcoded y padding for each macOS versions. It now works as expected on macOS 15.6, could you test it on macOS 26 and see if there is any unexpected behavior other than your original implementation?

It works great!
image
image

Excellent work!
I rebased my branch with the latest readest main branch and push our two commits again. nothing changes in our two commits.

PR #4250, opened 2026-05-21
Add GHCR and Docker Hub image publishing 

pourmand1376:
## Summary

This PR adds Docker image publishing for Readest and updates the web client/docker setup so self-hosted deployments can use prebuilt images without rebuilding for environment changes.

The problem was that I wanted to run this project but I really don't building docker packages. It would be much cleaner if it could be provided officially. The users just set the env variables and pull the image and done. 

### Docker release workflow
- adds a new GitHub Actions workflow to publish multi-arch Docker images
- publishes to GHCR by default
- optionally publishes to Docker Hub when `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets are configured

### Runtime config support for the web app
I also changed the project a little so that build-time configs be read at runtime instead of compile-time. 

## Why

Previously, the client relied on build-time `NEXT_PUBLIC_*` values, which made Docker image distribution impossible because changing deployment-specific values required rebuilding the image.

With this change:
- published images can be reused across environments
- self-hosters can configure the container through runtime env vars
- releases can automatically publish Docker images for easier deployment

## Notes
- GHCR publishing works out of the box with `GITHUB_TOKEN`
- Docker Hub publishing is optional and only enabled when the related secrets are present
- local build workflows are still supported through `docker/compose.build.yaml`

### Tests
- I have tested this inside codespaces and everything is fine. 
- Published images for testing are now available at two address:
  - GHCR Address: https://github.com/pourmand1376/readest/pkgs/container/readest
  - Docker Address: https://hub.docker.com/repository/docker/amirpourmand/readest/general

I'm also the maintainer of [al-folio](https://github.com/alshedivat/al-folio) project and I did a similar approach for that project to include it in the [dockerhub](https://hub.docker.com/r/amirpourmand/al-folio). You can create your own username and token for dockerhub. Or I can add you as maintainer to image that I've build and you update it however you want. No difference to me in this matter.


pourmand1376:
I appreciate it if you could take a look @chrox .

chrox (changes requested):
Thanks for the PR — the runtime-config refactor is the right shape for prebuilt images. Two issues I'd block on before merging:

### 1. `S3_ENDPOINT: http://minio:9000` breaks browser uploads/downloads on self-hosted MinIO

`docker/compose.yaml` changes `S3_ENDPOINT` from `http://${HOST_IP}:9000` to `http://minio:9000`. `S3_ENDPOINT` is server-only (no `NEXT_PUBLIC_` prefix), but the call graph is:

1. Browser → `POST /api/storage/upload` (`src/libs/storage.ts:50`)
2. Server runs `getUploadSignedUrl()` → `getSignedUrl(s3Client, putCommand, ...)` in `src/utils/s3.ts:64`
3. The `S3Client` is configured with `forcePathStyle: true` + `endpoint: process.env['S3_ENDPOINT']` (`s3.ts:17-22`). With path-style + SigV4 (the default), the host is part of the canonical signed request, so the returned URL is `http://minio:9000/<bucket>/<key>?X-Amz-Signature=...` and the host can't be rewritten without invalidating the signature.
4. Server hands that URL back as `{ uploadUrl }`
5. Browser does `xhr.open('PUT', uploadUrl, true)` in `src/utils/transfer.ts:22`

`minio` is only resolvable inside the docker network, so every browser-side PUT/GET against MinIO will fail. Same applies to `src/app/api/share/[token]/{download,cover}/route.ts`.

**Suggested fix:** keep `S3_ENDPOINT` browser-reachable (`http://${HOST_IP}:9000`), or introduce a separate `S3_PUBLIC_ENDPOINT` and pass it to a second `S3Client` used only for `getSignedUrl` calls, while the existing client keeps `http://minio:9000` for server-side SDK calls (`PutObject`/`CopyObject`/`DeleteObject`/`HeadObject` in `s3.ts:84,93,102,123`).

### 2. `<Script strategy="beforeInteractive">` in `src/pages/_app.tsx` may not honor the strategy

Per the [Next.js Pages Router docs](https://nextjs.org/docs/pages/api-reference/components/script#beforeinteractive), `beforeInteractive` must be placed in `pages/_document.js` to apply. Placing it in `_app.tsx` typically downgrades silently to `afterInteractive`.

If it downgrades, `window.__READEST_RUNTIME_CONFIG` may not be set before client modules that read it at module-eval time:
- `src/utils/supabase.ts:4` (top-level const)
- `src/utils/storage.ts` (`getStorageType`, called from `book.ts` at render)
- `src/services/environment.ts:15` (`getBaseUrl`)

When that happens, the chain falls through to `atob(process.env['NEXT_PUBLIC_DEFAULT_SUPABASE_URL_BASE64']!)`, meaning a self-hosted browser may briefly (or permanently, depending on which routes are still served by Pages Router) hit `readest.app`'s Supabase instead of the local one.

**Suggested fix:** add a `src/pages/_document.tsx` and move the `<Script>` there. The App-Router copy in `src/app/layout.tsx` is already in the right place.

### Smaller things (non-blocking)

- `docker/README.md:129` standalone-run example uses port 7000; should be 8000.
- `compose.yaml`'s new `SUPABASE_PUBLIC_URL` drops the `:-8000` default fallback on `KONG_HTTP_PORT`.
- `runtimeConfig.ts` doesn't fall back to old `NEXT_PUBLIC_OBJECT_STORAGE_TYPE` / `NEXT_PUBLIC_STORAGE_FIXED_QUOTA` / `NEXT_PUBLIC_TRANSLATION_FIXED_QUOTA` for users with hand-rolled docker envs (inconsistent with the supabase/api fallbacks already in place).

pourmand1376:
Thanks for your review. All your items are addressed. 

Please review Again. 

> note that I didn't use my dockerhub username (or yours in that regard) in the docs. It is better to decide where it is going to be and fix that.

chrox (approved):

chrox:
Thanks.

Issue #4261, opened 2026-05-21
Github Container Registry Package is not Public


pourmand1376:
I've recently added this. But as I wanted to test I understood that the created image isn't public. 

```
Error response from daemon: unknown: failed to resolve reference "ghcr.io/readest/readest:latest": unexpected status from HEAD request to https://ghcr.io/v2/readest/readest/manifests/latest: 403 Forbidden
```

Images are private by default and we should make them public.

pourmand1376:
This is related to this pull request. 
- https://github.com/readest/readest/pull/4250

chrox:
Done

PR #4263, opened 2026-05-21
ci: optimize build time for Docker and CI workflows

pourmand1376:
## Summary

Speeds up Docker multi-arch builds and CI workflows, fixes a pnpm version mismatch, and patches a known artifact-write CVE.

Docker Image Built time is now decreased significantly. (30 min -> 6 min)

### 🚀 Build Performance
- **Eliminate QEMU emulation** for arm64: replace single QEMU job with a native matrix build (`ubuntu-latest` for amd64, `ubuntu-24.04-arm` for arm64), removing ~21 min of emulated build time
- **Switch Docker layer cache** from `type=gha` (10 GB shared cap, no per-platform isolation) to `type=registry` on GHCR (no size cap, already authenticated, per-platform cache tags: `buildcache-linux-amd64` / `buildcache-linux-arm64`)
- **Add multi-arch merge job** that assembles the final manifest from per-platform digests
- **Slim production Docker stage** to copy only runtime artifacts (`.next`, `public`, `node_modules`, `package.json`, `next.config.mjs`), dropping `src/`, `src-tauri/`, `patches/`, and other source-only directories
- **Add `sharing=locked`** to the pnpm store cache mount to prevent cache corruption during concurrent builds

### ⚙️ CI Workflow Improvements
- Use `pnpm install --frozen-lockfile --prefer-offline` in `pull-request.yml` and `release.yml` for faster, reproducible installs
- Skip redundant ESLint pass during `next build` in `next.config.mjs` (lint already runs as a dedicated CI step)

### 🔧 Bug Fixes
- **Fix pnpm version in Dockerfile**: `10.29.3` → `11.1.1` (matches `package.json`), preventing corepack from re-downloading pnpm on every build

### 🔒 Security
- **Upgrade artifact actions to v7** (`upload-artifact` + `download-artifact`): patches arbitrary file write vulnerability in `actions/download-artifact >= 4.0.0 < 4.1.3` (SHA-pinned, consistent with the rest of the repo's workflows)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,8 +13,30 @@ target
 # Build outputs
 .next
 **/.next
+.open-next
+**/.open-next
 out
 **/out
+.vercel
+**/.vercel
+
+# Local env files
+docker/.env
+.env*.local
+**/.env*.local
+
+# Local credentials and tooling state
+*.pem
+certs
+**/certs
+.claude/settings.local.json
+**/.claude/settings.local.json
+.claude/worktrees
+**/.claude/worktrees
+.gstack
+**/.gstack
+.playwright-mcp
+**/.playwright-mcp
 
 # IDE
 .idea
@@ -28,4 +50,3 @@ Thumbs.db
 # Logs
 *.log
 npm-debug.log*
-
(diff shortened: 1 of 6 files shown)


pourmand1376:
Please also consider enabling ghcr to public:
- #4261

PR #4278, opened 2026-05-23
fix(layout): use 100vh fallback for .full-height to unbreak old Chromium WebView

### LINKED ISSUE #4274: [bug] .full-height collapses on Chrome <108 Android WebView (library page renders ~52px)
## Bug Description

On Android devices that ship with bundled WebView < 108 (e.g. many MTK Android 12 devices that lack Play Services and cannot update Android System WebView), the library page renders the top nav strip only — the bookshelf area below collapses to 0 height. The user sees the header (with "Search 6 books..." indicating the library is loaded), then nothing.

Same bug surfaces on any old-Chromium host: Tauri-on-old-Android, old WKWebView (iOS <15.4), and any embedded WebView (third-party launcher apps shipping `web.readest.com`).

Verified via injected DOM probe on a real Chrome 101 Android WebView:

```
<body.light>                                                            377×798
  <div#app>                                                             377×798
    <div.library-page.text-base-content.full-height>  flex, h=52px ✗    377×52   ← only nav
      <div.bookshelf.min-h-0.flex-grow>               flex-grow→0       377×0
        <div.bookshelf-items.grid>                    actually 635px    377×635  (clipped by parent)
          <div.virtuoso-grid-item>                                      104×212  (books DO render)
```

The bookshelf grid is rendered correctly (book items measure 104×212), but the `.library-page.full-height` parent has `overflow: hidden` and only allocated 52 px of height (= the nav bar), so everything else gets clipped.

## Expected behavior

`.library-page.full-height` should fill the viewport on every supported browser, including older Chromium WebViews still in active deployment on cheap Android hardware.

## Technical inputs

**Operating System (one reproduction):** Android 12 (WebView 101.0.4951.61), MTK SoC, no Play Services
**Readest Version:** `web.readest.com` as of 2026-05-22 (also reproducible on every Readest build since PR #2582)

### Root cause

PR #2582 (commit `e28dabce`, 2025-12-01) refactored `.full-height` for edge-to-edge support:

```css
.full-height {
  height: -webkit-fill-available;
}

@supports (height: 100dvh) {
  .full-height {
    height: 100dvh;
  }
}

:root.edge-to-edge .full-height {
  height: 100vh !important;
}
```

The intent is solid (progressively enhance to `100dvh` on modern; use `-webkit-fill-available` to avoid the iOS Safari URL-bar overflow problem with `100vh`). The breakage: `-webkit-fill-available` on **pre-108 Chromium Android WebView** does not produce a usable computed height — the spec/impl has been inconsistent across Chromium versions, and on these particular WebView builds it evaluates to a tiny value (effectively the parent's intrinsic content height) instead of the viewport-fill behaviour the rule expects. Combined with `@supports (height: 100dvh)` being false on the same browsers (dvh is Chrome 108+), `.full-height` falls through to the broken `-webkit-fill-available` rule with no recovery.

So the matrix is:

| Browser | `-webkit-fill-available` | `100dvh` | `.full-height` resolves to |
|---|---|---|---|
| Modern Chrome / Safari 15.4+ | (anything) | ✓ | `100dvh` ✓ |
| Old iOS Safari < 15.4 | ✓ (excludes URL bar) | ✗ | `-webkit-fill-available` ✓ |
| **Chrome / WebView 101 Android** | **broken** | ✗ | **`-webkit-fill-available` (broken)** ✗ |

### Proposed fix (please choose)

There's a real trade-off and I'd rather you pick the direction:

**Option A — Safe & universal (recommended):**

```diff
 .full-height {
-  height: -webkit-fill-available;
+  height: 100vh;
 }
 @supports (height: 100dvh) {
   .full-height { height: 100dvh; }
 }
 :root.edge-to-edge .full-height { height: 100vh !important; }
```

`100vh` is universally supported and produces the right viewport-fill behaviour on every browser. The cost: on **old iOS Safari (< 15.4)** the library content will overflow under the dynamic URL bar by ~50 px. dvh is supported on Safari 15.4 (March 2022) so this only affects iOS 15.3 and earlier, which should be a very small slice in 2026.

**Option B — Preserve iOS Safari behaviour:**

Use UA / platform sniff (via JS or a `data-platform` attribute on `<html>`) to pick `-webkit-fill-available` only on iOS, `100vh` elsewhere. More code, but no iOS Safari regression.

**Option C — `min-height: 100vh` everywhere:**

```diff
 .full-height {
+  min-height: 100vh;
   height: -webkit-fill-available;
 }
```

Provides a guaranteed floor on old Chromium while keeping the `-webkit-fill-available` fast-path where it works. Doesn't cap at viewport on iOS where the URL bar is present though — content could overflow downward.

I'd default to A. Happy to send a PR once you confirm the direction.

### Workaround already deployed downstream

For now I'm injecting a `<style>` polyfill at `onPageFinished` in our WebView host:

```css
.full-height        { height: 100vh !important; }
.min-h-dvh, .h-dvh, .h-svh, .h-lvh,
.min-h-svh, .min-h-lvh { min-height: 100vh !important; }
```

gated on `!CSS.supports('height', '100dvh')` so modern WebViews are untouched. Restores the bookshelf and any other dvh-using utility immediately. Same root cause as this issue.

iFocuspace:
## Summary

Closes #4274.

PR #2582 set `.full-height { height: -webkit-fill-available }` as the fallback before the `@supports (height: 100dvh)` upgrade. On pre-108 Chromium Android WebView (still shipped on a lot of MTK Android 12 devices that can't update Android System WebView) that property evaluates to the parent's intrinsic content height rather than viewport-fill — `.library-page.full-height` collapses to ~52 px (the nav bar) and the bookshelf is clipped by the parent's `overflow: hidden`.

Replace the fallback with `height: 100vh`. Universal support, correct viewport-fill on every browser. The `@supports (height: 100dvh)` rule still upgrades to dvh on modern browsers (Safari 15.4+, Chrome 108+), keeping URL-bar-aware behaviour where it's available.

## Diff

```diff
 .full-height {
-  height: -webkit-fill-available;
+  height: 100vh;
 }
 @supports (height: 100dvh) {
   .full-height { height: 100dvh; }
 }
 :root.edge-to-edge .full-height { height: 100vh !important; }
```

## Trade-off (per #4274 discussion, Option A confirmed)

iOS Safari < 15.4 will overflow under the dynamic URL bar by ~50 px when on `.full-height` pages — dvh has been available on Safari since 15.4 (March 2022), so the affected slice is small in 2026.

## Test plan

- [x] On Chrome 101 WebView (MTK Android 12, no Play Services): library page now renders the full bookshelf instead of just the nav bar
- [x] On modern Safari / Chrome: still upgrades to `100dvh` via `@supports`, no visual change
- [x] No other selectors touched, no JS changes

### DIFF
--- a/apps/readest-app/src/styles/globals.css
+++ b/apps/readest-app/src/styles/globals.css
@@ -74,7 +74,7 @@ body {
 }
 
 .full-height {
-  height: -webkit-fill-available;
+  height: 100vh;
 }
 
 @supports (height: 100dvh) {


iFocuspace:
Thanks for the quick merge! Just a heads-up: `web.readest.com` is still serving the pre-fix bundle (`c32edc9b703b468c.css` with `height: -webkit-fill-available`). The fix is on `main` but doesn't seem to have reached production yet. Happy to test on the same MTK Android 12 / WebView 101 device once it's live.
image

chrox:
Deployed. Please test it on your devices.

iFocuspace:
Confirmed fixed on my end. Working as expected now, appreciate the fast deployment!

PR #4290, opened 2026-05-25
feat(ai): improve RAG indexing UX and reader navigation flow

loveheaven:
## Summary

Improves the user experience around RAG indexing and a small Android-back navigation quirk in the reader. No new providers, no schema changes.

> **Stacked on #4289 (OpenRouter provider).** Until that PR merges, this PR's diff includes its commits as well; once PR1 is merged into `main`, GitHub will automatically narrow this PR's diff down to the changes described below.

## What changes

**Batched embedding with real progress** (`src/services/ai/ragService.ts`)
* `indexBook` now embeds chunks in batches (16 for Ollama, 64 otherwise) instead of one giant `embedMany` call. Concretely:
  * Progress reporting reflects real work — the progress bar advances per batch instead of jumping from 0 to 100 at the end.
  * We stay within per-request input limits that several providers enforce. Ollama's effective batch size is small in particular and the previous single-call path could OOM the embedding server on large books.
  * On batch failure, we throw an enriched error including the failing chunk range and the embedding model name, so the panel can show a useful message instead of a generic "Embedding failed".

**Indexing error surfaced in the AI panel** (`src/app/reader/components/notebook/AIAssistant.tsx`)
* Indexing errors now show in the panel as a dismissible block instead of being swallowed.
* The CTA label switches to **Retry Indexing** on failure.
* Progress-bar background uses a theme token so it follows light/dark mode correctly.

**Empty-state stops flickering** (`Thread`)
* Removes the custom `hasActiveConversation` gate around the empty state and relies on assistant-ui's own `ThreadPrimitive.Empty`. The placeholder hides immediately when the first message is sent, instead of waiting for the conversation to be persisted.

**Android back closes the right panel first** (`Reader`)
* When both the notebook and the sidebar are open on Android, back now closes the notebook first, then the sidebar. So users who opened the AI tab from the sidebar's chat-history list return to that list on the first back press, and close the sidebar on the second — matching native expectations.

## Manual testing

* Indexed a 600+ chunk book on both Ollama (local) and an OpenAI-Compatible gateway:
  * Progress bar moves smoothly batch by batch.
  * Forced an embedding error mid-way (server stop): panel shows `Embedding failed at chunk 257/612 (model: ...): connection refused` and the CTA becomes **Retry Indexing**. Retry succeeds without losing earlier progress.
* Verified the empty-state no longer flashes when sending the first message.
* Verified the Android two-step back flow on a Tauri Android build.

## Compatibility

* No schema migrations.
* Public API of `indexBook` unchanged.
* Behavior change is strictly additive (more granular progress, more informative errors).

### DIFF
--- a/apps/readest-app/package.json
+++ b/apps/readest-app/package.json
@@ -84,6 +84,7 @@
     "worktree:rm": "pnpm exec tsx scripts/worktree-rm.ts"
   },
   "dependencies": {
+    "@ai-sdk/openai-compatible": "^2.0.48",
     "@ai-sdk/react": "^3.0.49",
     "@assistant-ui/react": "0.11.56",
     "@assistant-ui/react-ai-sdk": "1.1.21",
--- a/apps/readest-app/src/__tests__/ai/providers.test.ts
+++ b/apps/readest-app/src/__tests__/ai/providers.test.ts
@@ -24,8 +24,18 @@ vi.mock('ai-sdk-ollama', () => ({
   }),
 }));
 
+// mock @ai-sdk/openai-compatible so OpenRouterProvider can be constructed
+// without going over the network during unit tests.
+vi.mock('@ai-sdk/openai-compatible', () => ({
+  createOpenAICompatible: vi.fn(() => ({
+    chatModel: vi.fn(),
+    textEmbeddingModel: vi.fn(),
+  })),
+}));
+
 import { OllamaProvider } from '@/services/ai/providers/OllamaProvider';
 import { AIGatewayProvider } from '@/services/ai/providers/AIGatewayProvider';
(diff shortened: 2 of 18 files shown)


chrox:
Please hold on refactoring the AI module for now, as there is an ongoing effort to migrate it to an agent-based stack.

loveheaven:
> Please hold on refactoring the AI module for now, as there is an ongoing effort to migrate it to an agent-based stack.

sure. no problem. What about the previous PR https://github.com/readest/readest/pull/4289 ? it is not refactoring. it adds a feature to support OpenRouter/OpenAI models.

chrox:
That PR should be OK to be merged.

PR #4311, opened 2026-05-26
fix(opds): make Android Back / Reader→Back behave correctly inside the OPDS browser

loveheaven:
## Summary

Three intertwined bugs all surfaced as "Back from inside an OPDS catalog escapes too far," plus a related visual issue with the Back/Forward buttons. All fixed together here.

### 1. Android Back / Esc no longer collapses the whole browser in one go

Previously the webview's default history-back fired, unmounted `/opds` and routed to `/library` *without* the URL flag needed to re-open the standalone OPDS catalogs dialog, so a second Back press also left `/library` — two presses to escape, dialog gone.

Now `/opds` mounts a `useKeyDownActions` hook that walks the in-page navigation stack when it has anything to pop (sub-feed, publication detail, search) and only closes the browser at the catalog root. The hook reads `historyIndex` via ref so the stale `useEffect` closure inside `useKeyDownActions` doesn't permanently see the first render's empty history. `closeOPDSBrowser` is also patched to merge `opds=true` into the restored library URL itself, instead of relying on `navigateToLibrary(navBack=true)` which overwrites `queryParams` with sessionStorage's `lastLibraryParams` and dropped the flag.

### 2. Reader → Back now lands the user back on the publication detail page rather than the bare book list

Opening a book leaves `/opds` entirely, so when the webview history-backs into `/opds` it remounts with an empty in-page history of length 1 and the user lost both the detail view and every sub-feed they had drilled through.

`handlePublicationSelect` now `pushState`s a `pub=<groupIndex>:<itemIndex>` marker into the URL before changing React state, and a mount-time restore effect re-applies the marker once the feed loads. `handleBack`/`handleForward` keep the marker in sync during normal in-page navigation, and the restore effect guards against malformed or no-longer-matching markers.

### 3. Unrelated settings churn no longer clobbers the restored detail

The mount-load `useEffect` depends on the whole zustand `settings` object, so any unrelated mutation (theme, font, etc.) re-rendered `OPDSCatalogPage`, re-fired the effect, re-fetched the same feed, reset `viewMode` to `'feed'`, cleared `selectedPublication`, and pushed a duplicate feed entry on top of the just-restored publication — the user saw a loading flash and dropped back to the list, with two spurious entries in the in-page history. The `loadingOPDSRef` in-flight guard didn't help because the original fetch had already completed by the time the spurious re-run hit.

Track the last `(catalogId, url)` pair the effect actually issued a fetch for in a ref and short-circuit re-entries with the same pair.

### 4. Back/Forward Navigation buttons looked enabled when actually disabled

The existing `disabled:bg-transparent` utility cancelled the daisyUI ghost-button disabled tinting, leaving the SVG chevrons at full opacity. Added `disabled:opacity-40` and `not-allowed` cursor so the disabled state is visible.

## Testing

- Existing OPDS test suite passes (4819 tests).
- Manually verified on Android: drill into a multi-level catalog, open a book, use system Back — lands on the publication detail; Back again returns to the previous feed; eventually returns to `/library` with the OPDS catalogs dialog re-opened.
- Verified Back/Forward chevrons now visibly dim when no further navigation is possible.

## Files changed

- `apps/readest-app/src/app/opds/page.tsx` — history restoration, key handler, settings-churn guard
- `apps/readest-app/src/app/opds/components/Navigation.tsx` — disabled-state styling
- `apps/readest-app/src/store/...` — `closeOPDSBrowser` URL flag fix

image
After fix, the forward/back button can show the disabled state.

### DIFF
--- a/apps/readest-app/src/app/opds/components/Navigation.tsx
+++ b/apps/readest-app/src/app/opds/components/Navigation.tsx
@@ -90,7 +90,7 @@ export function Navigation({
         <div className='flex gap-1'>
           {onBack && (
             <button
-              className='btn btn-ghost btn-sm px-1 disabled:bg-transparent'
+              className='btn btn-ghost btn-sm px-1 disabled:cursor-not-allowed disabled:bg-transparent disabled:opacity-40'
               onClick={onBack}
               disabled={!canGoBack}
               title={_('Back')}
@@ -100,7 +100,7 @@ export function Navigation({
           )}
           {onForward && (
             <button
-              className='btn btn-ghost btn-sm px-1 disabled:bg-transparent'
+              className='btn btn-ghost btn-sm px-1 disabled:cursor-not-allowed disabled:bg-transparent disabled:opacity-40'
               onClick={onForward}
               disabled={!canGoForward}
               title={_('Forward')}
--- a/apps/readest-app/src/app/opds/page.tsx
+++ b/apps/readest-app/src/app/opds/page.tsx
@@ -13,6 +13,7 @@ import { downloadFile } from '@/libs/storage';
 import { Toast } from '@/components/Toast';
 import { useThemeStore } from '@/store/themeStore';
 import { useTranslation } from '@/hooks/useTranslation';
+import { useKeyDownActions } from '@/hooks/useKeyDownActions';
 import { useLibraryStore } from '@/store/libraryStore';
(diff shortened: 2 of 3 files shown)


loveheaven:
I made a mistake that I put other fix in this PR. Please wait a moment. I will move other fix into another branch.

PR #4314, opened 2026-05-26
feat(ios): folder import with security-scoped bookmark persistence

loveheaven:
Adds iOS folder-picker support and persists access across app launches via security-scoped bookmarks. After this PR, any iOS code path that needs to read from a user-chosen folder root — Books/, Duokan/, Calibre/, Moon+ Reader/, an iCloud-mirror folder, etc. — can rely on the folder still being reachable on the next launch without re-prompting the user.

### What's in this PR

**Commit 1 — `feat(import): support folder picker on iOS via native-bridge`**

- Routes folder selection through a Swift `UIDocumentPickerViewController` (`.folder` content type) bridged from the web view, since `tauri-plugin-dialog` rejects folder picks on iOS with `FolderPickerNotImplemented`.
- `FolderPickerDelegate` retains the picked `URL` and calls `startAccessingSecurityScopedResource()` for the lifetime of the app process so subsequent Foundation / POSIX reads under that root succeed.
- Routes `NativeAppService.selectDirectory` through the bridge for both iOS and Android, then calls `allowPathsInScopes` so the picked directory is reachable via `fs_scope` and the asset protocol.
- In-process retention only — restarting the app loses access. Persistence is added in commit 2.

**Commit 2 — `feat(ios): persist security-scoped bookmarks for picked folders`**

- New Swift `FolderBookmarkStore`: serializes `URL.bookmarkData(.minimalBookmark)` keyed by POSIX path into `UserDefaults`; resolves on `NativeBridgePlugin.load(webview:)`, re-encodes on `isStale`, drops permanently unresolvable entries.
- The bookmark resolution loop calls `startAccessingSecurityScopedResource` for every persisted root and holds the URL alive in a process-scoped dictionary, so the WebView's first fs read after launch already succeeds.
- `allow_paths_in_scopes` gets an iOS branch that widens `fs_scope` / `asset_protocol_scope` for any path the frontend hands it, intentionally without the desktop-side "must already be in fs_scope" gate. The OS sandbox plus the bookmark token is the real access-control boundary on iOS; widening Tauri's in-memory scope cannot escalate access beyond what the OS already grants. Desktop behavior is unchanged.
- `allow_file_in_scopes` is now compiled for iOS too (previously desktop-only).

### Scope

This PR is platform plumbing only — folder picker + bookmark persistence + Tauri scope wiring. The frontend behavior that consumes these primitives (re-issuing `allowPathsInScopes` against the user's saved folder list at launch, surfacing those folders in the library, etc.) is intentionally left to a follow-up PR stacked on top of this one, to keep this change reviewable as a self-contained platform layer.

### Testing

- Manual: pick a folder on iPad → kill app → relaunch → folder is still accessible without a re-pick prompt.
- `isStale` path exercised by renaming the folder in Files.app between launches; bookmark gets re-encoded silently.
- Unresolvable bookmarks (folder deleted, provider uninstalled) are dropped from `UserDefaults` on the next launch instead of being retried forever.
- Desktop unchanged; existing `allow_paths_in_scopes` behavior and tests are unaffected.

### DIFF
--- a/apps/readest-app/src-tauri/plugins/tauri-plugin-native-bridge/ios/Sources/NativeBridgePlugin.swift
+++ b/apps/readest-app/src-tauri/plugins/tauri-plugin-native-bridge/ios/Sources/NativeBridgePlugin.swift
@@ -7,6 +7,7 @@ import StoreKit
 import SwiftRs
 import Tauri
 import UIKit
+import UniformTypeIdentifiers
 import WebKit
 import os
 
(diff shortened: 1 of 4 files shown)


loveheaven:
I will explain this PR completely in the next PR.

Here just explain a little bit.

I need folder importer on iOS to import books under Duokan or other folder into Readest.
image
image
image


image

chrox (approved):

Issue #1925, opened 2025-08-29
FR: support reading books directly from a directory on the device without importing books


devacame:
Apps like koreader use local storage as main directory and this being default behavior would help with alternate app usage and reduce the hassle of setting up folder structures again. 

~Also with webdav/opds support, if you could add download all from directory functionality, that's be great too.~

chrox:
This may be feasible on desktop environments, but it becomes increasingly impractical on Android and iOS, as these sandboxed systems don’t provide users with convenient tools for organizing files into directory tree structures. And on the Web platform, directory structures don’t exist. For a cross-platform reading app, this feature would therefore not be truly cross-platform at all.

devacame:
While I'm not sure sure why you think mobile devices are inconvenient for file organization, the Web part makes sense. Maybe you could support uploading entire folders then? For someone who have hundreds of books, having to upload them one at a time and having them unorganized is a bit deterring.

chrox:
> Maybe you could support uploading entire folders then? For someone who have hundreds of books, having to upload them one at a time and having them unorganized is a bit deterring.

This sounds reasonable.

hbirs:
I imported several hundred books in the Android version, but it takes quite a while for the cover previews to load in list view and other screens. On some devices, the app may even freeze.
It seems like the app is loading data (possibly JSON) on demand, so it might need its own caching mechanism similar to a browser cache. Please consider improving the response performance so that the screens can be viewed more smoothly.

Another possible approach would be to add an option in the settings that controls whether imported data is automatically added to the list when Readest loads it.
If users could instead open the data directly without importing—just browsing and opening books directly from the file explorer—it might help avoid issues like slow cover previews or delayed loading times.

chrox:
> I imported several hundred books in the Android version, but it takes quite a while for the cover previews to load in list view and other screens. On some devices, the app may even freeze. It seems like the app is loading data (possibly JSON) on demand, so it might need its own caching mechanism similar to a browser cache. Please consider improving the response performance so that the screens can be viewed more smoothly.
> 
> Another possible approach would be to add an option in the settings that controls whether imported data is automatically added to the list when Readest loads it. If users could instead open the data directly without importing—just browsing and opening books directly from the file explorer—it might help avoid issues like slow cover previews or delayed loading times.

Yes we have identified this performance issue as in list view all book covers are loaded at once we need to render only the visible items at least. We will fix this in the upcoming updates.

flanter21:
[...]

PR #4315, opened 2026-05-26
feat(library): in-place import from registered external folders

loveheaven:
> **Stacked on top of #4314.** This PR's branch is based on `feat/ios-folder-import` (PR #4314), so the diff currently includes those 2 commits at the bottom. **Please review and merge #4314 first** — after that I will rebase this branch onto `main` and the diff will shrink to only the 2 commits below.
>
> Commits owned by this PR (top of the stack):
> - `feat(library): in-place import with cloud sync and symmetric local delete`
> - `feat(library): one-tap "read in place" toggle in folder import`
>
> Commits inherited from #4314 (please ignore here — review in that PR):
> - `feat(import): support folder picker on iOS via native-bridge`
> - `feat(ios): persist security-scoped bookmarks for picked folders`

---

Lets users keep books inside an existing on-disk library (Duokan, Calibre, Moon+ Reader, iCloud mirror, …) and read them in place instead of duplicating every file into Readest's hash-keyed `Books/<hash>/` directory. Critical for users with large local libraries who don't want a second copy.

### What's in this PR

**Commit 1 — `feat(library): in-place import with cloud sync and symmetric local delete`**

The mechanism. Adds an `inPlace` option to `importBook`: a source file inside a registered external library folder is referenced directly via `book.filePath` instead of being copied. Sidecars (cover, config, nav) still live under `Books/<hash>/`.

`ingestService` routes through `shouldImportInPlace`, which marks an import in-place when the absolute source path lives under any of `settings.externalLibraryFolders` and is **not** inside a per-root `Books/` subtree. The Readest data dir (`customRootDir`) is intentionally excluded — that directory is Readest's home and should freely hold hash copies; in-place is for user-registered roots.

Cloud sync treats in-place books as first-class:

- `uploadBook` reads bytes from `(book.filePath, 'None')` when set. The cloud key is unchanged, so a peer downloading the book lands it under `Books/<hash>/` as a normal hash copy.
- `useBooksSync` strips `book.filePath` before pushing — it is a device-local path meaningless on any other device.
- `ingestService` no longer skips upload for in-place books; `autoUpload` / `forceUpload` behave like any other book. Only transient imports opt out.
- `deleteBook` `'local'`/`'both'` now physically removes the source file at `book.filePath` (base `'None'`). Local-delete semantics are symmetric with hash-copy books: the local copy is gone, the cloud backup remains, a future pull restores under `Books/<hash>/`. `removeFile` errors are swallowed.

New `SystemSettings.externalLibraryFolders?: string[]`. Added to `BACKUP_SETTINGS_BLACKLIST` alongside `localBooksDir` / `customRootDir` so device-local paths don't ride cloud backups.

**Commit 2 — `feat(library): one-tap "read in place" toggle in folder import`**

The user entrypoint. Adds a checkbox to the folder-import dialog. When the user opts in:

- The selected folder is auto-registered into `settings.externalLibraryFolders` (deduplicated).
- The choice is remembered as the default for the next import via `localStorage` (`readest:lastImportFolderReadInPlace`).
- Folders that are already registered force the checkbox ON and lock it (you cannot half-import a registered root).

### Testing

`cloud-service`, `ingest-service`, and `backup-settings` suites cover in-place delete, multi-root matching, per-root `Books/` guard, and the backup-strip. Full vitest suite passes (4858 tests).

### Compatibility notes

- **No UI yet for unregistering** a folder — that's a follow-up. Current entrypoint is via the import-dialog checkbox.
- The cloud `book` record's `filePath` is stripped on push; if any pre-PR dogfood client has already pushed a dirty `filePath`, the first pull after merge will bring it back into local state. Functionally harmless (`isBookAvailable` prefers the hash path), and the next push from that device strips it.
- iOS support requires #4314 (folder picker + bookmark persistence). On desktop and Android, this PR works standalone.

### DIFF
--- a/apps/readest-app/src/__tests__/services/backup-settings.test.ts
+++ b/apps/readest-app/src/__tests__/services/backup-settings.test.ts
@@ -24,6 +24,7 @@ function makeSettings(overrides: Partial<SystemSettings> = {}): SystemSettings {
     migrationVersion: 3,
     localBooksDir: '/Users/me/Books',
     customRootDir: '/Users/me/readest',
+    externalLibraryFolders: ['/Users/me/Duokan', '/Users/me/Calibre'],
     keepLogin: true,
     screenBrightness: 0.7,
     autoScreenBrightness: false,
@@ -85,6 +86,7 @@ describe('sanitizeSettingsForBackup - blacklist', () => {
     const out = rec(sanitizeSettingsForBackup(makeSettings()));
     expect(out['localBooksDir']).toBeUndefined();
     expect(out['customRootDir']).toBeUndefined();
+    expect(out['externalLibraryFolders']).toBeUndefined();
     expect(out['savedBookCoverForLockScreenPath']).toBeUndefined();
   });
 
--- a/apps/readest-app/src/__tests__/services/book-content-source.test.ts
+++ b/apps/readest-app/src/__tests__/services/book-content-source.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, test, vi } from 'vitest';
+import { exportBook, getBookFileSize, isBookAvailable } from '@/services/bookService';
+import { getLocalBookFilename } from '@/utils/book';
+import type { Book } from '@/types/book';
+import type { BaseDir, FileSystem } from '@/types/system';
+
+function makeBook(overrides: Partial<Book> = {}): Book {
+  return {
+    hash: 'bookhash',
+    format: 'EPUB',
+    title: 'sample',
+    author: 'Author',
+    createdAt: 1,
+    updatedAt: 1,
+    deletedAt: null,
+    downloadedAt: 1,
+    ...overrides,
+  };
+}
+
+function makeFs(options: {
+  existing?: Array<[string, BaseDir]>;
+  files?: Record<string, File>;
+}): FileSystem {
+  const existing = new Set((options.existing ?? []).map(([path, base]) => `${base}:${path}`));
+  const files = options.files ?? {};
+  return {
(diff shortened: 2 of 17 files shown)


loveheaven:
This PR just does one thing: to import books from an existing folder(where Duokan/Moon Reader/Calibre store their books)  and don't copy books into System/Readest/<book hash>/. I want to save my space on my android/iPhone.

image

1. The first rectangle  shows how  users  choose a folder on Iphone(where Duokan store its books). 

image

2. And the second red rectangle shows after uses choose the folder and select "Read books in place", the books will keep where they are.

**It doesn't change any other behaviour.**

chrox:
@copilot resolve the merge conflicts in this pull request

loveheaven:
I force-pushed the branch to resolve the conflicts. please merge please!

Before merging, I want to confirm what you think of this in-place import feature.

chrox:
This is a great feature and could be a good step forward to https://github.com/readest/readest/issues/1925.

loveheaven:
> This is a great feature and could be a good step forward to #1925.

Thank you very much!!!!

loveheaven:
Everything is Perfect! I verified your commit. Please close this PR.

chrox:
> Everything is Perfect! I verified your commit. Please close this PR.

Great. Please also test it in WebDAV before we release the next update.

loveheaven:
> > Everything is Perfect! I verified your commit. Please close this PR.
> 
> Great. Please also test it in WebDAV before we release the next update.

OK. No problem.

Issue #4198, opened 2026-05-17
Library pull-to-refresh shows stale reading progress from other devices until the source reader closes


Yee-Code:
## Bug Description

When a book is open and being read on one device (e.g. web), the in-reader auto-sync pushes `configs` (including `progress`, `location`, `updatedAt`) to the server successfully. However, on a second device (e.g. iPad), pulling-to-refresh the library page does **not** show the updated reading progress for that book. The progress only appears on the second device after the book is **closed** on the first device.

The user-visible symptom is: "as long as I keep the book open on web, no amount of pull-to-refresh on the iPad library updates the progress percentage on the cover."

### Root cause (confirmed by code reading)

Readest has two independent sync lanes (`books` and `configs`), and the library refresh path only touches one of them:

1. **In-reader push goes to `configs` only.**
   `apps/readest-app/src/app/reader/hooks/useProgressSync.ts` (`pushConfig`, L28-39) debounces and calls `syncConfigs([compressed], hash, metaHash, 'push')`. It does **not** update local `Book.progress` and does **not** push `books`.

2. **Library pull-to-refresh pulls `books` only.**
   `apps/readest-app/src/app/library/page.tsx` L185-192 registers `usePullToRefresh` → `pullLibrary()`.
   `apps/readest-app/src/app/library/hooks/useBooksSync.ts` L37-57: `pullLibrary` calls `syncBooks([], 'pull', since)` and nothing else. A repo-wide grep confirms `syncConfigs` is only invoked from `useProgressSync.ts` (i.e. only while a reader is mounted for that book).

3. **The library progress UI reads `book.progress`, not `config.progress`.**
   `apps/readest-app/src/app/library/components/ReadingProgress.tsx` L12-21 computes the percentage from `book.progress[0] / book.progress[1]`. Its `memo` comparator (L62-67) only re-renders when `book.hash / book.updatedAt / book.readingStatus` change.

4. **`book.progress` is only refreshed locally via `saveConfig`.**
   `apps/readest-app/src/store/bookDataStore.ts` L73-102: `saveConfig` is what copies `config.progress` onto a new `Book`, bumps `book.updatedAt`, and persists it. This runs on reader teardown / close, not on every page-flip.

5. **Consequence.**
[...]

PR #4341, opened 2026-05-28
fix(sync): prevent cross-device progress overwrite; retry first pull on flaky networks

### LINKED ISSUE #4222: Bug: Readest sync does not work at all and will overwrite itself with old location
## Bug Description

I posted about this on reddit last week and now that I have confirmed this doesnt just effect me (numerous other users confirmed identical behavior), I wanted to make an issue. 

In its current state, I cannot get Readest sync to actually work reliably (or sometimes at all) between two devices. 

I currently use Readest on Android via Boox Go 7 and a Samsung S26U. I have the latest version of Readest (0.11.1) installed on both devices. I am signed into my account on both devices and have the exact same epub files downloaded to both devices. 
Readest will not sync the location between the two devices. I can reproduce the below behavior almost 100% of the time. 

### Steps to reproduce
1. Open book on device 1. Read a chapter on the book, increasing progress from 3% to 4% (chapter 4 to 5). Confirm while in the book that a sync was completed. 
2. Go to Readest home on device 1. Confirm that the book now correctly shows 4% 
3. Open Readest on device 2. On home screen, the book will currently show 3% (where I previously let off on that device, as expected). Pull down to trigger sync. The book will update from 3% (chapter 4) to 4% (chapter 5), which is expected. 
4. Open book. Book opens to chapter 4 (where it was before I progressed further). It does not update the page position to the newest (correct) location.
5. While still inside the book, with no interaction on my part, a new sync is triggered, overwriting the correct location (chapter 5) with this older location (chapter 4). 
6. If I then go back to device 1 and re-open the book, about half the time it will then pull that old location and revert my progress on device 1. 

## Expected behavior
When a user triggers a sync on Readest, therefore pulling a new location on a book (as reflected in the ui by progress on the home screen increasing percentage), when the user opens that book it is expected that Readest would correctly display the updated page location. 

Currently, even though Readest seems to be aware of the updated location, it does not do anything. Even more troublesome is that it then overwrites the correct page location with the prior location, causing the user to lose progress on all devices. 

I am not saavy enough to fully troubleshoot what is going on here, but there are essentially two major sync issues that render Readest unusable in its current form: 

1. The UI doesn't seem to recognize that progress has been pulled, via sync, and therefore does not update the page position correctly
2. The sync will overwrite the current position with backwards progress. This should almost never happen (at least without some kind of prompt). 

## Technical inputs
Boox Go 7 (Black/White) on Android 13
Samsung S26 Ultra on OneUi 8.5 
Readest 0.11.1

chrox:
## Summary

- **Fix #4222**: stop the artificial `config.updatedAt` bump on book open so a stale local push can't overwrite a newer server-side config via last-writer-wins.
- **Retry the first pull on flaky networks**: `pullWithRetry` (1.5s / 4s / 10s backoff) replaces the bare initial pull and releases the push gate after exhaustion. Single-failure used to block every push for the whole reader session, which lined up well with the Android reports on the ticket.
- **Server piggyback**: `/api/sync POST` updates `books.progress` + `books.updated_at` for each upserted config, removing the second `syncBooks` round-trip from the in-reader auto-push (saves one POST per save, keeps the library row in sync while a reader stays open — related to #4198).
- Bumped sync `fetchWithTimeout` from 8s to 15s for the same Android cold-start reason.

## Why each change

| Symptom from #4222 | Fix |
|---|---|
| Open book on device 2; library row is fresh but the book opens to the old position and overwrites the other device's progress | `useProgressAutoSave` no longer fires `saveConfig` when the in-memory location still matches the loaded-from-disk position. With `config.updatedAt` un-bumped, the auto-push loses the server compare and the user's real progress survives. |
| Single failed pull blocks every push for the rest of the reader session (handleAutoSync only re-arms on `progress.location` changes) | `pullWithRetry` retries with backoff and releases the gate after exhaustion. Server-side last-writer-wins still protects the cross-device case for fresh-import scenarios. |
| Each in-reader auto-save sent two POSTs (configs + books) so other devices' library pull-to-refresh would see fresh progress (#4198 fix) | `/api/sync POST` updates the `books` row off the same configs push (gated by `.lt('updated_at')`). The client's second POST is dropped. |

## Test plan

- [x] `pnpm test` — 4888 pass (was 4885 — three new retry tests + four `useProgressAutoSave` tests; one `useProgressSync` test rewritten to assert the syncBooks round-trip is gone).
- [x] `pnpm lint` — clean.
- [x] Manual: open a book on device A → read → confirm push. Open same book on device B → confirm view jumps to A's position. Toggle airplane mode on B at the moment of book open → confirm device B's progress eventually pushes after the retry window, doesn't clobber A's progress before/after.
- [x] Manual: brand-new EPUB import on device B that's already at chapter 5 on A → confirm B doesn't push chapter-1 over A's chapter 5.

### DIFF
--- a/apps/readest-app/src/__tests__/hooks/useProgressAutoSave.test.tsx
+++ b/apps/readest-app/src/__tests__/hooks/useProgressAutoSave.test.tsx
@@ -1,0 +1,126 @@
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { act, cleanup, renderHook } from '@testing-library/react';
+
+type Progress = { location: string } | null;
+
+const h = vi.hoisted(() => {
+  const makeStore = <T,>(state: T) => {
+    const fn = () => state;
+    (fn as unknown as { getState: () => T }).getState = () => state;
+    return fn as (() => T) & { getState: () => T };
+  };
+
+  const state = {
+    config: { location: 'cfi-loc', updatedAt: 1000 } as { location: string; updatedAt: number },
+    progress: { location: 'cfi-loc' } as Progress,
+    previewMode: false,
+  };
+
+  return {
+    makeStore,
+    state,
+    saveConfigMock: vi.fn(async () => {}),
+  };
+});
+
+vi.mock('@/context/EnvContext', () => ({
+  useEnv: () => ({ envConfig: { name: 'env' } }),
+}));
+
+vi.mock('@/store/bookDataStore', () => ({
+  useBookDataStore: h.makeStore({
+    getConfig: () => h.state.config,
+    saveConfig: h.saveConfigMock,
+  }),
+}));
+
+vi.mock('@/store/readerStore', () => ({
+  useReaderStore: h.makeStore({
+    getProgress: () => h.state.progress,
+    getViewState: () => ({ previewMode: h.state.previewMode }),
+  }),
+}));
+
+vi.mock('@/store/settingsStore', () => ({
+  useSettingsStore: h.makeStore({ settings: { version: 1 } }),
+}));
+
+import { useProgressAutoSave } from '@/app/reader/hooks/useProgressAutoSave';
+
+const flushDebouncedSave = async () => {
+  await act(async () => {
+    // debounce 1000ms + inner setTimeout 500ms + slack
+    vi.advanceTimersByTime(2000);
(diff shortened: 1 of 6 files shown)


titandrive:
Awesome. Do you have an eta when this update will release by chance?

PR #4352, opened 2026-05-29
perf(sidebar): virtualize BooknoteView and memoize derivations

loveheaven:
Switching the annotation/bookmark sidebar to a flat virtualized list eliminates the per-item layout reads that caused multi-second jank when toggling tabs on books with hundreds of notes.

A. Virtualize the list with react-virtuoso
   - Flatten group headers + notes into a single FlatBooknoteRow array.
   - Embed an OverlayScrollbars instance inside the tab so scrollbar styling is preserved while Virtuoso owns the viewport (same nested pattern as TOCView).
   - Track the parent scroll-container's height with ResizeObserver to give Virtuoso a bounded viewport.
   - Replace the per-item useScrollToItem (which called getBoundingClientRect and closest() on every BooknoteItem on every progress tick — O(n) sync reflow on 1000+ items) with a single virtuosoRef.scrollToIndex driven by nearestCfi.

B. Stabilize derivations with useMemo / useCallback
   - filteredNotes, sortedGroups, flatItems, nearestCfi all useMemo so an unrelated config change (e.g. viewSettings autosave) no longer triggers a full sort + group rebuild.
   - handleBrowseBookNotes is now useCallback so BooknoteItem's React.memo can hit on prop equality.

C. Memoize BooknoteItem
   - Wrap the component in React.memo. With stable item / onClick references from the parent, re-renders triggered by sibling progress updates no longer cascade across every visible row.
   - Cache marked.parse(item.note) and dayjs(item.createdAt).fromNow() in useMemo. marked is the dominant per-render cost for note rows.
   - isCurrent moves to a useMemo over isCfiInLocation; the per-item scrollIntoView is removed since BooknoteView now drives scrolling.

useScrollToItem is intentionally left intact — SearchResults still uses it and its smaller list does not exhibit the same jank.

### DIFF
--- a/apps/readest-app/src/app/reader/components/sidebar/BooknoteItem.tsx
+++ b/apps/readest-app/src/app/reader/components/sidebar/BooknoteItem.tsx
@@ -1,6 +1,6 @@
 import clsx from 'clsx';
 import dayjs from 'dayjs';
-import React, { useRef, useState } from 'react';
+import React, { useMemo, useRef, useState } from 'react';
 import { MdEdit, MdDelete } from 'react-icons/md';
 
 import { marked } from 'marked';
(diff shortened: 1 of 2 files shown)


loveheaven:
I made 1100 annotations on  a book. The switch between different tabs in BookNoteview was very slow if I didn't do this optimization.

PR #4369, opened 2026-05-30
perf(import+open): native Rust EPUB/MOBI parser, OPF prefetch, parallel TOC enrichment

loveheaven:
## Summary

Push EPUB / MOBI / AZW3 metadata extraction and OPF prefetch into a native Rust pre-parser on Tauri targets, and parallelize the JS-side TOC fragment enrichment. The goal is to cut the import-time and book-open-time CPU spent on the JS thread (the bottleneck on iOS WebView and on low-end Android), while keeping web targets, fallback paths and existing on-disk formats unchanged.

5 commits, structured as one logical chain:

| # | Commit | Layer |
|---|---|---|
| 1 | `perf(epub): add native EPUB parser in Rust` | Rust infra |
| 2 | `perf(import): use native EPUB parser and downscale covers on Tauri targets` | JS consumer (import) |
| 3 | `perf(reader): prefetch EPUB OPF/nav from Rust on book open` | JS consumer (open) |
| 4 | `perf(nav): parallelize section scans and memoize fragment lookups` | JS-only optimization (works without the Rust path) |
| 5 | `perf(import): use native MOBI/AZW/AZW3 parser on Tauri targets` | Same pattern, second format |

## Motivation

On iOS WebView and low-end Android, the foliate-js EPUB / MOBI parsers block the JS main thread on:

- a full ZIP central-directory parse (EPUB) or full record-stream inflate (MOBI), per `importBook` and again per `openBook`;
- a second pass over the file to compute `partialMD5`;
- a per-section ZIP inflate inside `computeBookNav` for fragment CFI calculation.

For a 5 MB EPUB on an iPhone 11, this is the dominant cost in the import / open hot path. Moving the metadata extraction and the OPF / nav fetch into native Rust lets us:

- inflate at native speed instead of via zip.js running in the WebView;
- compute `partialMD5` from the same buffer the parser already has, without a second read;
- return the OPF, EPUB3 nav, NCX and a manifest-size table in a single IPC, so the spine TOC walk on the JS side is served from an in-memory cache instead of inflating again per chapter.

### 1. `perf(epub): add native EPUB parser in Rust` (`9d04707`)

Adds three Tauri commands in a new `epub_parser` module:

- `parse_epub_metadata(path)` – returns `{ title, author, cover_jpeg, partial_md5 }` for the import hot path. Cover is decoded and re-encoded to a webview-friendly JPEG when its long edge exceeds the library-grid size, so the on-disk thumbnail drops from multi-MB to ~30–60 KB.
- `parse_epub_full(path)` – returns the OPF bytes, EPUB3 nav bytes, NCX bytes and a `{ entry-name → uncompressed-size }` table, for the reader open hot path.
- `extract_epub_cover_full(path)` – returns the original full-resolution cover bytes, for the lock-screen wallpaper writer.

All three avoid ferrying multi-MB blobs across the JS↔Rust IPC boundary. No JS callers yet — wired up in commits 2 and 3.

New Rust deps (all pure-Rust, ship to every Tauri target):
`zip 2` (deflate-only) · `quick-xml 0.36` · `md-5 0.10` · `base64 0.22` · `percent-encoding 2` · `image 0.25` (default features off, jpeg/png/gif only).

### 2. `perf(import): use native EPUB parser and downscale covers on Tauri targets` (`e3fc476`)

`bookService.importBook` now forwards EPUB metadata + cover extraction to `parse_epub_metadata` and reuses the `partialMD5` it returns, skipping both the foliate-js full-archive parse and the second read for hashing.

Side effect: the cover written to `cover.png` is now downscaled to a webview-friendly JPEG. To keep the lock-screen wallpaper feature unchanged, `useAutoSaveBookCover` now pulls the original full-resolution cover via `extract_epub_cover_full` and falls back to the (now-downscaled) thumbnail when the native path is unavailable.

Web targets and non-EPUB formats keep the existing path. New file: `src/utils/tauriEpubBridge.ts`.

### 3. `perf(reader): prefetch EPUB OPF/nav from Rust on book open` (`3f1d0ac`)

When opening an EPUB on Tauri targets, `DocumentLoader` calls `parse_epub_full` up-front. The foliate-js zip loader is wrapped so that `loadText()` of the OPF, EPUB3 nav, NCX and a synthetic `META-INF/container.xml` is served from that in-memory cache without inflating through zip.js. All other assets keep flowing through the original loader.

A small in-flight dedupe is added to the spine-text loader so the nav pipeline (`loadText` + `createDocument` back-to-back on the same href) doesn't pay for two zip.js inflate calls per chapter on first open.

Plumbing: `readerStore.openBook` resolves an absolute on-disk path via the new `appService.resolveNativeBookFilePath` / `bookService.resolveNativeBookFilePath` helper and threads it into `DocumentLoader` as `nativeFilePath`. Web targets, non-EPUB formats and books without a managed/external on-disk path skip the prefetch.

### 4. `perf(nav): parallelize section scans and memoize fragment lookups` (`51ac997`)

Pure JS-side optimization, works whether or not the Rust prefetch is available:

- `computeBookNav` processes sections via `Promise.all` instead of a sequential for-loop, and within each section issues `loadText()` and `createDocument()` concurrently. Combined with the in-flight `loadText` dedupe added in commit 3, each chapter pays for a single zip inflate per nav build, and inflates of different chapters overlap.
- `enrichTocFromNavElements` is restructured into two phases: a cheap `<nav` substring filter on the inflated text, and a parsed-document walk for the survivors. Most chapters fall out in phase 1 without ever being parsed.
- `fragments.ts::calculateFragmentSize` consults a per-section position cache (`makeFragmentPositionCache`), so the N-fragment loop is `O(N)` over the chapter HTML instead of `O(N²)`.
- A small `isCfiAddressable` guard skips elements that foliate-js's CFI generator can't address (`documentElement`, `body` itself, detached nodes, nodes outside `<body>`) — these previously threw and spammed `console.warn` for every fragment, now they silently fall back to the section CFI.

### 5. `perf(import): use native MOBI/AZW/AZW3 parser on Tauri targets` (`c296f65`)

Mirrors commit 2 for Kindle formats. New `mobi_parser` (built on the `mobi 0.8` crate, KF7 + KF8 reader) reads title, author, publisher, ISBN, ASIN, publish date, language, subjects and description from the MobiHeader + EXTH records, resolves the EXTH 201 cover offset against the PDB image-record table (with `ThumbOffset` / first-image fallbacks), and strips KindleGen's HTML wrapping in EXTH 103 so the description goes into the library DB as plain text.

Cover-resize / `partialMD5` / `RawCoverImage` are extracted into a new `parser_common` module shared between `epub_parser` and `mobi_parser`, so a single tweak applies to every native importer and the `partialMD5` algorithm can't drift between formats (a divergent algorithm would silently re-import every existing book under a new hash on first run).

`bookService.importBook` tries the EPUB native path first, then the MOBI native path; both bridges fall back to the foliate-js DocumentLoader on web targets, parse error, or format mismatch. New deps: `mobi 0.8`. New file: `src/utils/tauriMobiBridge.ts`.

## Compatibility / fallback matrix

| Platform / format | Path |
|---|---|
| Tauri (desktop/iOS/Android) + EPUB | Rust fast-path; falls back to foliate-js on parse error |
| Tauri + MOBI/AZW/AZW3/PRC | Rust fast-path; falls back to foliate-js on parse error |
| Tauri + PDF / FB2 / CBZ / TXT | Unchanged (foliate-js) |
| Web target, all formats | Unchanged (foliate-js) |
| Books without a managed on-disk path | Skips Rust prefetch, falls back |
| Encrypted / DRM EPUB | Falls back (Rust parser doesn't decrypt) |

## Risks

- **New Rust deps**: 7 pure-Rust crates (zip, quick-xml, md-5, base64, percent-encoding, image with default features off, mobi). No C/C++ toolchain. Binary size impact is small (<1 MB across all targets in practice — `image` with only jpeg/png/gif decoders is the largest contributor).
- **Cover thumbnail size change**: `cover.png` written during import is now a downscaled JPEG (~30–60 KB) rather than the original cover. Lock-screen wallpaper still uses the full-resolution cover via `extract_epub_cover_full`. Existing libraries are not migrated — covers stay at their previous size until re-imported. The `getCoverFilename` extension was kept as `.png` to avoid invalidating existing covers.
- **`partialMD5` parity**: the Rust implementation is byte-exact with the JS one (shared via `parser_common`). A divergence would cause silent re-imports under new hashes, which is why the algorithm is centralized.
- **EPUB / MOBI edge cases**: encrypted EPUBs, malformed central directories, MOBI files without EXTH 201 — all return `Err` from Rust and trigger the JS fallback.

## Testing

- Unit tests added for `epub_parser`, `mobi_parser` and the JS bridges.
- Manual smoke test on macOS desktop and iOS device build with: standard EPUBs, EPUBs with CJK paths, percent-encoded OPF hrefs, MOBI (KF7), AZW3 (KF8), large covers, missing covers.
- No real-device latency benchmarks are included in this PR — the optimization rationale is mechanistic (fewer JS-thread inflates and IPC round-trips), not benchmark-driven. Maintainers who want numbers, please ask and I will run a profile on iOS.

## Out of scope

- PDF native parsing (deliberately not included — `pdf.js` still owns the cover render path on PDFs and replacing it requires a different trade-off).
- Cover format change for already-imported books (not migrated).
- Encrypted / DRM EPUB support.


loveheaven:
The benefits: 
1. The import speed is 3 time faster. When I tried to import 550 epub and pdf files, it took me 10 minutes to import all the books. After this optimization, it took me 2 minutes to import all the books.
2. The cover of every epub/mobi/azw3 is resized to 512 px at most. The memory the library consumed plummets. The library scrolls very fast.
3.  The open speed of a epub/mobi is 30% faster.
4. pdf is left unchanged.

chrox:
Thanks for this — the native pre-parser is a strong perf idea, and the open-time `parse_epub_full` prefetch (Rust hands foliate the OPF/nav bytes, foliate parses them) is exactly the right shape.

Before it lands, I want to flag one architectural concern on the **import** path.

On Tauri (desktop + iOS + Android), `bookService.importBook` uses the native Rust parser as the **source of truth** for metadata, with foliate-js only as a fallback (`tryNativeParseEpub` → `tryNativeParseMobi` → `DocumentLoader`). So `parse_epub_metadata` / `parse_mobi_metadata` — a from-scratch reimplementation of OPF/MOBI metadata extraction — becomes what populates every imported book's title / author / identifier / series / cover on our primary platforms, demoting the battle-tested foliate-js parser to fallback.

The risk/reward there feels lopsided:

- **The expensive work isn't the metadata parse.** The costs this PR targets are the zip central-directory parse + inflate (slow in the WebView), the second read for `partialMD5`, and cover decode/resize. Parsing the *already-inflated* OPF XML is cheap — so reimplementing it in Rust buys very little speed while taking on the full correctness surface of a fresh EPUB/MOBI parser.
- **A quick review already turned up several divergences from foliate-js**, each of which is wrong metadata in a real library:
  - `dcterms:modified` mapped onto `published` (foliate keeps them separate) → EPUB3 books with no `<dc:date>` get a bogus publication date;
  - multi-author books collapse to a single `metaHash` (`authors.join(', ')` vs foliate's per-author array) → cross-platform/sync dedup misses, duplicate library entries;
  - HTML entities (`&amp;`) aren't decoded, and cp1252 MOBI metadata mojibakes (foliate decodes via the file's encoding header);
  - the synthetic `container.xml` doesn't XML-escape the OPF path — a path containing `&`, `<`, `>` or `"` produces malformed XML and the book fails to open, with no fallback.

  Finding this many in a shallow pass strongly suggests a long tail (refinements, `file-as`/`role`, alternate-script titles, multiple languages, `belongs-to-collection`, calibre columns, …) — each a metadata bug that only shows up on the primary platform.

**Suggested direction:** keep Rust doing the heavy mechanical work, but let foliate-js stay the single source of truth for metadata — have the import command return the OPF bytes (+ cover + `partialMD5`) and let foliate parse them, exactly like `parse_epub_full` already does for the open path. You keep essentially all the perf win, and the whole class of metadata-parity bugs disappears by construction (it's literally the same parser — nothing to keep in sync).

I've got a re-architecture along these lines ready (foliate parses metadata on both import and open; Rust keeps the `parse_epub_full` prefetch + `partialMD5`; cover downscaling moves to JS) and can push it onto this branch if that direction works for you.

loveheaven:
**Thanks for the careful review — you're right on the architectural point.** I agree with you on everything (especially about foliate-js becomes the single source of truth for OPF metadata on both import and open paths) **except one thing** which is **cover downscaling** I'd like to keep on the Rust side. The reasoning:

- on import, scaling and re-encoding cover images in JS runs serially on the renderer thread (no real Worker-level parallelism for createImageBitmap + canvas toBlob without spinning up workers, and even then it stalls on the main thread for blob transfer);
- bulk imports (a few hundred books from a Calibre library) are the scenario where this actually shows up — Rust's image crate decoding + resizing in parallel via rayon is meaningfully faster on multi-core devices, especially Android mid-tier;
- cover bytes don't participate in metadata correctness, so keeping them on the Rust side doesn't expand the parser-parity surface;
- the IPC payload is identical either way (downscaled bytes vs. raw bytes), so there's no marshalling overhead trade-off.

Concretely I'd suggest the Rust command continues to return { partialMd5, opfBytes, opfPath, navBytes?, navPath?, coverBytes, coverMime } where coverBytes is already downscaled to the thumbnail target dimensions, and JS just stores it as-is. Happy to change my mind if you have benchmarks showing the JS path is competitive on Android — but my instinct from the bulk-import profiling I did earlier is that it isn't.

Everything else as you've described it sounds right to me. Thanks again for catching this before it shipped.

If you agree, I can push it onto this branch. it is also okay if you push it. Please feel free to let me know your opinion and your decision.

chrox:
Thanks, please update this PR for it.

loveheaven:
> Thanks, please update this PR for it.

I updated this PR. Now use foliate.js to parse meta for epub/mobi/azw3 and use rust for cover resizing and partial md5. So now this PR depends on https://github.com/readest/foliate-js/pull/19.

Importing 550 epub/pdf books costs 5.5min currently on main branch.  After this update, it costs 2.5 min.



Please review it again

chrox:
We might need more carefully performance profiling to identify the hot spot in the importing process. And then either optimize it in foliate-js or replace it with rust implementation.

loveheaven:
> We might need more carefully performance profiling to identify the hot spot in the importing process. And then either optimize it in foliate-js or replace it with rust implementation.

I already did a performance profiling by importing 550 epubs/pdfs(500 epubs and 50 pdfs).



That's the time cost of every stage which pure JS took before this PR. 

Stage | Total | Average | Median| p90 | Max
-- | -- | -- | -- | -- | --
js parse epub | 480.0s | 878 ms | 835 ms | 1374 ms | 6948 ms
write cover.png | 311.5s | 571 ms | 430 ms | 964 ms | 6353 ms
getCover() | 191.9s | 352 ms | 213 ms | 673 ms | 6563 ms
partialMD5 (js) | 142.3s | 260 ms | 256 ms | 471 ms | 2166 ms
open file handle | 70.7s | 129 ms | 132 ms | 161 ms | 230 ms
save book config | 61.3s | 112 ms | 34 ms | 302 ms | 1212 ms
mkdir Books/ | 43.8s | 80 ms | 58 ms | 148 ms | 964 ms
generateCoverImageUrl | 0.1s | — | — | — | —



That's why I decide to resize cover in Rust to reduce the cost of reading/writing cover and why I decide to do partialMD5  in Rust. and that's also in the beginning why I decided to use rust to parse meta.

Now the time cost for importing is 2 or 3 times faster with keeping Folitate.js to parse meta. There should be very little risk to accept this PR. If you want to optimize further in folitate.js, I think I can do it in another PR.

If you think this PR is big, I can split it into several small PRs so that you can review them easily.

loveheaven:
> We might need more carefully performance profiling to identify the hot spot in the importing process. And then either optimize it in foliate-js or replace it with rust implementation.

Hey. what do you think of this PR? 
After the last refactor, it is working without any risk. Want to know your opinion.

chrox:
Yeah, we will merge it as soon as possible.

PR #4391, opened 2026-05-31
fix(reader): scroll wide EPUB tables horizontally

Jesusm1229:
image

## Summary

Wide EPUB/HTML tables (e.g. many columns without per-cell `width`) could overflow the reading column and overlap body text. The previous `applyTableStyle()` path scaled tables via CSS `transform` only when cell widths were known, so tables like “Table 3-1” often stayed full width with no containment.

This PR:

- Wraps each `<table>` in a `.readest-table-scroll` container with `overflow-x: auto` and `width: max-content` on the table (layout CSS in `getPageLayoutStyles`, same layer as `pre` overflow).
- Removes transform-based scaling from `applyTableStyle()`.
- Adds `applyTableTouchScroll()` so horizontal swipes on scrollable tables scroll the table in **paginated** mode instead of turning the page (capture-phase listeners, same pattern as the brightness gesture / foliate touch docs). At the left/right scroll edge, page turns still work.

## Impact

| Before | After |
|--------|--------|
| Wide tables bleed across the page | Horizontal scroll within the column |
| Tables with explicit widths often scaled down (tiny text) | Natural size + scroll when needed |
| Mobile paginated: horizontal swipe on table = page turn | Swipe on table scrolls; at edges, page turn |

Applies to reflowable EPUB/HTML sections via `FoliateViewer` on load (web + Tauri). PDF / fixed-layout unchanged.

## Tests

- Updated `style-dom.test.ts` for wrapper + touch helpers
- `pnpm --filter @readest/readest-app exec vitest run src/__tests__/utils/style-dom.test.ts` — 48 passed

## Test plan

- Please Test on mobile devices since I was unable to.


Jesusm1229:
Book used AI Engineering https://z-lib.sk/book/r9b0GvKjqB/ai-engineering.html

Jesusm1229:
Please, tell me if this approach is approved by the Maintainers and community.

chrox (commented):
Added two follow-up commits addressing issues found while testing this PR:

**`195f02f2f` — keep wheel/trackpad table scrolling from turning the page**
- `applyTableTouchScroll` only routed *touch* events; trackpad/mouse **wheel** scrolling over a table (forwarded to pagination via `handleWheel` → `iframe-wheel`) both scrolled the table and turned the page. Added capture-phase `wheel` routing that consumes a horizontal wheel over a scrollable table, incl. at the scroll edge so trackpad momentum never chains into a page turn.
- `findWrapper` used `target instanceof Element`, which is always `false` for iframe event targets (this module runs in the top-window realm), so the touch routing never fired. Switched to duck-typing on `closest`.

**`c2454ccfb` — don't show a spurious scrollbar on layout tables**
- Wrapping every table made layout/character tables that wrap to fit the column show a spurious horizontal scrollbar.
- The wrapper now clips (no scrollbar) for a table that fits within a few px of the column, and a `ResizeObserver` re-evaluates as the column width settles — so a table genuinely wider than the column always scrolls and is never clipped, while one that wraps to fit shows no scrollbar.

Added a Chromium browser test (`paginator-table-layout.browser.test.ts`) over `sample-table-layout.epub` (layout tables must not scroll) and `sample-table-wide.epub` (a too-wide table must scroll, not clip). Full suite green (unit 5063, browser 158), lint clean.

Jesusm1229:
Thank you so much, chrox

PR #4392, opened 2026-05-31
fix(reader): replace light callout backgrounds in dark mode

Jesusm1229:
## Summary

In dark mode, EPUB callout boxes (sidebars, tips, formula blocks) often keep white or light gray backgrounds from publisher CSS or inline styles, while Readest applies the theme foreground on `html`/`body` and rewrites black text. The result is light text on a light box and poor contrast.

This change applies dark theme backgrounds when **Override book colors** is off (the default):

- Override inline `background` / `background-color` values that are white or `rgb(255, …)` in the reader stylesheet
- Set `body.theme-dark` background to the theme color
- Rewrite light backgrounds in EPUB stylesheets during `transformStylesheet()` when the app is in dark mode (same pipeline as existing black-text rewrites)

When override book colors is enabled, existing full color override behavior is unchanged.

**Before:**

image

**After:**

image

## Tests

- `style-get-styles.test.ts`: dark-mode inline white background overrides; no overrides in light mode
- `style.test.ts`: `transformStylesheet` rewrites `#ffffff` in dark mode; leaves `#222` unchanged

## Test plan

- [ ] Open a book with a white/light callout box (e.g. O’Reilly-style tip or formula sidebar) in **dark mode** with override book colors **off**
- [ ] Confirm callout background matches the reading theme and body text is readable
- [ ] Confirm **light mode** callouts are unchanged
- [ ] Toggle **Override book colors** on in dark mode and confirm behavior remains correct
- [ ] Spot-check a chapter with no callout boxes (no visual regression)

### DIFF
--- a/apps/readest-app/src/__tests__/utils/style-get-styles.test.ts
+++ b/apps/readest-app/src/__tests__/utils/style-get-styles.test.ts
@@ -571,6 +571,22 @@ describe('getColorStyles branches (via getStyles)', () => {
     expect(css).not.toMatch(/body\.pbg\s*\{[^}]*background-color:[^}]*!important/);
   });
 
+  it('overrides inline white backgrounds in dark mode when overrideColor is false', () => {
+    const vs = makeViewSettings({ overrideColor: false });
+    const theme = makeThemeCode({ isDarkMode: true, bg: '#1a1a1a', fg: '#e0e0e0' });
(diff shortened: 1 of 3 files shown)


Jesusm1229:
Book used AI Engineering: https://z-lib.sk/book/r9b0GvKjqB/ai-engineering.html

PR #4413, opened 2026-06-02
fix(readwise): use Tauri HTTP transport in desktop app

momadacoding:
## Summary

Switch `ReadwiseClient` to `@tauri-apps/plugin-http` when running in Tauri desktop mode, matching the pattern already used by WebDAV, Hardcover, AI providers, translators, OPDS, and KOReader sync.

## Why

In a Tauri webview context the standard `window.fetch` is still subject to CORS preflight rules and — on Android — the platform's cleartext-traffic policy. `@tauri-apps/plugin-http` sends the request from the Rust side via reqwest, bypassing the renderer entirely (no Origin header, no preflight, no cleartext block).

`ReadwiseClient` was one of the few remaining services that had not yet adopted this transport, so users on the desktop build could hit CORS or network errors when validating tokens or pushing highlights.

## Changes

- `ReadwiseClient.request()`: use `tauriFetch` instead of `globalThis.fetch` when `isTauriAppPlatform()` is true
- Added unit test to verify the Tauri HTTP path is taken in desktop mode

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


chrox:
Good catch.

PR #4530, opened 2026-06-10
fix(webdav): never push or pull for transient book sessions

loveheaven:
## Problem

When a user opens a file via Open-with (Android VIEW intent / iOS share sheet) or previews a draft from the /send flow, `appService.importBook(..., { transient: true })` creates an ephemeral `Book` whose bytes are NOT copied into `Books/<hash>/`. The Book carries an absolute `filePath` pointing at the original URI plus an immediately-set `deletedAt`, so the entry is filtered out of the bookshelf on next library load.

The Reader-level WebDAV sync hook (`useWebDAVSync`) didn't know about this distinction and treated the transient session like any other open book:

- `pushNow` debounce-flushed progress to `Readest/books/<hash>/config.json` after a few page-turns or on close.
- `pushBookFileNow` / `pushBookCoverNow` HEAD-probed and uploaded the binary + cover when `syncBooks` was on.
- The first-open `pullNow → pushNow` bootstrap created the per-hash directory on the remote even before any reading happened.

Two user-visible side effects follow:

1. The directory-listing fallback in `syncLibrary` (the loop that scans `Readest/books/` for hashes the local library doesn't know about) picks the orphan hash up on every sibling device's next Sync now and materialises a phantom shelf row — "a book I never imported keeps re-appearing across my devices".
2. `pullNow` could fetch an unrelated real book that happens to share the same hash on another device and merge its progress into the preview session, clobbering the in-memory location with state the user can't see in any UI.

## Fix

Add a `isTransientBook(book)` helper in the Reader hook that fingerprints transient imports as `!!book.filePath && !!book.deletedAt`. This combination uniquely identifies the transient path:

- In-place imports also set `filePath` but keep `deletedAt: null`, so they MUST sync (the import comments and `ingestService`'s upload logic both treat them as first-class books). The guard correctly leaves them alone.
- User-deleted books carry `deletedAt` but rarely `filePath`; they're already filtered earlier as tombstones, so a false positive is harmless either way.
- Brand-new transient previews are the only state where both fields are populated together at creation time (`bookService.ts:392, 525-529`).

`pushNow`, `pushBookFileNow`, `pushBookCoverNow` and `pullNow` early-return for transient books. The lock refs (`fileSyncedRef` / `coverSyncedRef`) are flipped only after the transient check, so a future real import of the same hash still gets a fresh upload attempt.

## What is intentionally NOT changed

`syncLibrary` (the library-wide "Sync now") is left as-is. `WebDAVForm`'s caller already filters out `deletedAt` rows before invoking it, and `bookService` writes `deletedAt` synchronously on every transient import, so the existing `!b.deletedAt` filter inside the service already keeps transient entries off the library-level push path. Adding a parallel `filePath && deletedAt` filter there would be a strict subset of the existing one — pure noise without a behaviour change.

## Test plan

New `webdav-skip-transient.test.ts` covers three scenarios end-to-end against `syncLibrary` with a stubbed fetch:

- `filePath + deletedAt` (transient) — `loadConfig` is never called and no PUT to the transient hash directory fires.
- `deletedAt` only (regular tombstone) — also skipped, locking the existing behaviour.
- `filePath` only with `deletedAt: null` (in-place import) — still reaches the upload pipeline as expected.

Existing 5182-test vitest suite passes via the pre-push hook.

## Caveat

Hash directories that earlier builds already pushed for transient sessions remain on the WebDAV server and won't be auto-cleaned by this change. Users can sweep them via `WebDAVBrowsePane`'s existing orphan-cleanup mode; this fix only closes the source of new pollution.

### DIFF
--- a/apps/readest-app/src/__tests__/services/webdav-skip-transient.test.ts
+++ b/apps/readest-app/src/__tests__/services/webdav-skip-transient.test.ts
@@ -1,0 +1,145 @@
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { syncLibrary } from '@/services/webdav/WebDAVSync';
+import type { WebDAVSettings } from '@/types/settings';
+import type { Book, BookConfig } from '@/types/book';
+
+/**
+ * `syncLibrary` must never upload anything for a transient import — those
+ * Books carry both `filePath` (source URI outside Books/<hash>/) and
+ * `deletedAt` (set at creation time by `bookService.importBook` when
+ * `transient: true`). Pushing config.json under `Readest/books/<hash>/`
+ * for such a hash creates a per-hash directory on the remote, which the
+ * directory-listing fallback in `syncLibrary` then materialises as a
+ * phantom shelf row on every sibling device's next sync.
+ *
+ * These tests exercise the library-wide push gate in isolation: we pass
+ * a tiny `books` array containing one transient and one normal entry,
+ * stub all network primitives, and assert that exactly the non-transient
+ * book ends up on the wire.
+ */
+
+const ORIGINAL_FETCH = globalThis.fetch;
+
+const settings: WebDAVSettings = {
+  enabled: true,
+  serverUrl: 'https://dav.example.com',
+  username: 'alice',
+  password: 'secret',
+  rootPath: '/',
+};
+
+const makeBook = (overrides: Partial<Book>): Book => ({
+  hash: 'h0',
+  format: 'EPUB',
+  title: 'Untitled',
+  author: 'Unknown',
+  createdAt: 1,
+  updatedAt: 1,
+  ...overrides,
+});
+
+const emptyConfig = (): BookConfig => ({ updatedAt: 0, booknotes: [] });
+
+let fetchMock: ReturnType<typeof vi.fn>;
+
+beforeEach(() => {
+  fetchMock = vi.fn();
+  // Default: every call is a benign 404 (nothing on the remote yet).
+  // Individual tests can override per call via `mockImplementation`.
+  fetchMock.mockResolvedValue(new Response(null, { status: 404 }));
+  globalThis.fetch = fetchMock as unknown as typeof fetch;
+});
+
+afterEach(() => {
+  globalThis.fetch = ORIGINAL_FETCH;
+  vi.restoreAllMocks();
+});
+
+describe('syncLibrary transient guard', () => {
+  test('skips books that have both filePath and deletedAt set', async () => {
+    const transient = makeBook({
+      hash: 'transient-hash',
+      filePath: 'content://shared/file.epub',
+      deletedAt: Date.now(),
+    });
+    const normal = makeBook({ hash: 'normal-hash' });
+
+    const loadConfig = vi.fn<(book: Book) => Promise<BookConfig>>(async () => emptyConfig());
+    const loadBookFile = vi.fn(async () => null);
+
+    await syncLibrary(settings, [transient, normal], {
+      strategy: 'send', // push-only path keeps the assertion focused
+      syncBooks: false,
+      deviceId: 'device-A',
+      loadConfig,
+      loadBookFile,
+    });
+
+    // `loadConfig` is the first per-book step in the push loop, so it's
+    // a clean signal for "did this book reach the upload pipeline?"
+    const configCalls = loadConfig.mock.calls.map(([book]) => book.hash);
+    expect(configCalls).toContain('normal-hash');
+    expect(configCalls).not.toContain('transient-hash');
+
+    // No PUT to `Readest/books/transient-hash/...` should ever fire.
+    const transientPuts = fetchMock.mock.calls.filter((call) => {
+      const [url, init] = call as [string, RequestInit | undefined];
+      return (init?.method ?? '').toUpperCase() === 'PUT' && url.includes('transient-hash');
+    });
+    expect(transientPuts).toHaveLength(0);
+  });
+
+  test('still skips when only deletedAt is set (regular tombstone)', async () => {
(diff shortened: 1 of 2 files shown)


loveheaven:
I have another question:

Now when you open an epub via Open-with (Android VIEW intent / iOS share sheet), Readest creates a <book hash> folder in the appdata folder. Do you think if it is a problem? 
If you think it is a problem, I will fix it.

chrox:
The config.json is expected to be saved in the appdata directory so this should be expected behavior.

loveheaven:
> The config.json is expected to be saved in the appdata directory so this should be expected behavior.

But these books are transient books. After you open it, you will possibly never read it again. After time passes, there will be a lot of <book bash> folders in the appdata directory. Is it not a problem?

chrox:
Users will expect the app to return to the last opened position, which requires preserving the config.json file across transient opens. Also, the overhead of keeping this config file is negligible on modern devices.

chrox:
Readest is designed to avoid expensive full-directory scans. The overhead is negligible compared to the storage and performance characteristics of modern APFS, ext4, and NTFS filesystems which are designed to support extremely large numbers of files and directories often in the billions or more.

chrox:
Transient books should still have a sidecar data file, which ideally should be synced across devices so that the same book opened transiently on different devices can share reading progress, annotations, and other metadata. In practice, “transient” simply means the book file itself is not imported into the Readest Books directory, the behavior is equivalent to enabling “Read books in place” when importing from a directory.

Given this understanding, could you re-evaluate whether this PR is actually necessary?

loveheaven:
> Transient books should still have a sidecar data file, which ideally should be synced across devices so that the same book opened transiently on different devices can share reading progress, annotations, and other metadata. In practice, “transient” simply means the book file itself is not imported into the Readest Books directory, the behavior is equivalent to enabling “Read books in place” when importing from a directory.
> 
> Given this understanding, could you re-evaluate whether this PR is actually necessary?

Okay. I basically agree with you. but I have one more question to confirm with you:
 Every day I browse lots of web pages randomly. Some of them are PDF files, which get automatically downloaded to my phone. I usually decide whether to keep them only after opening them in Readest. If I don't want to keep a file, I'll delete it — but by then, Readest has already created a <book hash> folder for it in the app data directory, and that folder just stays there.

So I'm wondering whether this design is really appropriate for my use case.

For books I do want to keep, I import them into my library — in that case, retaining the <book hash> folder makes total sense. But for books I don't want, I simply never import them. These are effectively transient books. If Readest keeps accumulating <book hash> folders for all of them, I worry that it could waste a significant amount of storage on my phone.

What do you think about this scenario?

chrox:
There is a 'Manage Cache' function in the advanced settings in the main menu of the library page. We can add an option to clear cache of deleted books there, this could save some storage by deleting the config.json file or the cover.png file if the book is already deleted in the library.

loveheaven:
> There is a 'Manage Cache' function in the advanced settings in the main menu of the library page. We can add an option to clear cache of deleted books there, this could save some storage by deleting the config.json file or the cover.png file if the book is already deleted in the library.

yes, I agree with you about we should add an option to clear the cache. 
But when we talk about WebDAV sync, why should we sync the transient books since I don't want to keep them? The books I want to keep will be imported into the library. Transient books are temporary books I don't want.

What do you think of it?

chrox:
For consistency with Readest's built-in sync behavior, the metadata of transient books is synced across devices as well. "Transient" simply means that the book file itself is not imported into the library; it does not mean that the reading state is disposable. Many users expect that when they open the same file on another device, their reading progress, annotations, and other metadata will be available automatically.

A particularly important use case is on desktop platforms, where users may disable Auto Import Books and instead use Finder or their file manager to organize their library, opening books directly in Readest as needed. In this workflow, all books are technically "transient," yet users would still reasonably expect reading progress, annotations, and other reading data to sync correctly across devices when using WebDAV sync.

More broadly, one of Readest's core goals is to remove the boundaries between devices and provide a seamless reading experience. Users should be able to treat all of their devices as a single reading environment without having to think about where a book was opened or worry about losing progress and fragmenting their reading data.

chrox:
If needed, we could introduce a dedicated preview mode or parameter for this use case. In preview mode, a file would be opened for temporary viewing only, without creating or retaining any reading data in the Readest data vault.

loveheaven:
> If needed, we could introduce a dedicated preview mode or parameter for this use case. In preview mode, a file would be opened for temporary viewing only, without creating or retaining any reading data in the Readest data vault.

yes.  I agree. so 
1. should I close this PR ? 
2. should I or you create another PR that is without creating or retaining any reading data in the Readest data vault.

chrox:
> 1. should I close this PR ?

I think we can close this PR.

> 2. should I or you create another PR that is without creating or retaining any reading data in the Readest data vault.

Based on your specific use case, which appears to be primarily desktop-oriented, I would prefer a CLI parameter that opens files in preview mode. This keeps the implementation simple.

loveheaven:
> > 1. should I close this PR ?
> 
> I think we can close this PR.

Okay. I will close this PR.

> 
> > 2. should I or you create another PR that is without creating or retaining any reading data in the Readest data vault.
> 
> Based on your specific use case, which appears to be primarily desktop-oriented, I would prefer a CLI parameter that opens files in preview mode. This keeps the implementation simple.

I primarily use android. that's why I care about my storage.

chrox:
An ordinary config.json file for a transient book typically takes up no more than a few kilobytes of storage. In practice, this is unlikely to make a noticeable difference unless you accumulate many thousands of transient books. For example, even 10,000 such files would only consume a few dozen megabytes.

If storage usage is the primary concern on Android, it may be more practical to use a dedicated preview application for files that you do not intend to keep reading.

PR #4556, opened 2026-06-12
perf(reader): throttle library.json writes and cache known dirs to cut IPC

loveheaven:
## Summary

`useProgressAutoSave` fires `saveConfig` ~once per second of reading. Two write-time sources were doubling its IPC cost on a release Android build (Chrome DevTools Bottom-Up profile showed `processIpcMessage` chewing ~25% of main-thread time during a reading session):

1. `saveConfig` wrote the whole `library.json` (plus its backup) on every call, so a user with N books paid `2 * JSON.stringify(N)` per save.
2. `nativeFileSystem.writeFile` / `copyFile` defensively called `plugin:fs|exists` before every write to ensure the parent dir existed. The same dir was probed once per save — about half of the IPC time per save was `exists()` round-trips against directories that had been there since the book was opened.

## Fix

- **`LIBRARY_SAVE_THROTTLE_MS = 30s`** coalesces a swipe burst into a single `library.json` write. Per-book `config.json` is still written eagerly because it is the sync source-of-truth and is small. `flushPendingLibrarySave()` is called on hook unmount and `window` blur so closing the book always flushes.
- **In-process `knownExistingDirs` Set** caches verified directories per app session. `createDir` adds; `removeDir` (including recursive) clears matching keys. Cold start still does the original `exists` + `createDir` dance once per directory.

## Testing

- All existing tests pass (`pnpm test`).
- Verified on a release Android build (aarch64): `processIpcMessage` share in Bottom-Up drops substantially during a sustained reading session.

This is the storage-layer half of a larger reader perf effort; React-side reductions (progress store split, per-field selectors, BookCell memoization) follow in separate PRs.

### DIFF
--- a/apps/readest-app/src/__tests__/store/book-data-store.test.ts
+++ b/apps/readest-app/src/__tests__/store/book-data-store.test.ts
@@ -8,7 +8,7 @@ vi.mock('@/utils/md5', () => ({
   md5Fingerprint: (value: string) => `md5_${value}`,
 }));
 
-import { useBookDataStore } from '@/store/bookDataStore';
+import { useBookDataStore, flushPendingLibrarySave } from '@/store/bookDataStore';
 import type { BookData } from '@/store/bookDataStore';
 import type { BookConfig, BookNote, Book } from '@/types/book';
 import { useLibraryStore } from '@/store/libraryStore';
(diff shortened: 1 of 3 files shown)


loveheaven:
I have 550 books in my library. Page turn animation often get stuck because saving library.json is a heavy work. and IPC(writefile, exists) make UI thread become stuck too. 
so that's why I need to make this PR.

PR #4557, opened 2026-06-12
perf(reader): split progress into its own store to cut React commit storm

loveheaven:
## Summary

Move per-book `progress` out of `readerStore.viewStates[key].progress` into a dedicated `readerProgressStore`. The previous setup wrote `progress` into `readerStore` on every `relocate` event during a swipe burst, fanning out re-renders to ~65 components subscribed to `useReaderStore()` (most via destructure) — the bulk of them did not care about progress at all, but were re-rendered anyway because Zustand subscribes the destructure-form to the entire state object.

## Changes

- New tiny store `store/readerProgressStore.ts` holds the per-book `BookProgress` map. `setBookProgress` only fires its own subscribers.
- `readerStore.setProgress` now writes progress to the new store and only touches `bookDataStore` for the primary view (secondary parallel views shouldn't overwrite the shared config).
- `readerStore.getProgress` is kept as a delegating facade so existing imperative call sites don't break.
- Components / hooks that genuinely need to react to progress changes subscribe via the new `useBookProgress(bookKey)` hook. The handful of call sites that just want a one-shot read use `getBookProgress(key)`.
- Test mocks updated.

## Why this matters

On a release Android build, profiling showed the swipe-burst hot path went through `setProgress` → entire `readerStore` setState → re-render of every `useReaderStore()` consumer. Splitting progress into its own store keeps the high-frequency write isolated from the much wider `readerStore` subscriber set.

## Depends on #4556

This PR's `useProgressAutoSave.ts` uses `flushPendingLibrarySave` from #4556 to flush on hook unmount. Please merge #4556 first.

## Testing

All existing tests pass (`pnpm test` — 5288 passed).

### DIFF
--- a/apps/readest-app/src/__tests__/components/BooknoteView.test.tsx
+++ b/apps/readest-app/src/__tests__/components/BooknoteView.test.tsx
@@ -19,13 +19,21 @@ let mockProgress: { location: string } | null;
 let mockBooknotes: BookNote[];
 
 // ---------- Mocks ----------
-vi.mock('@/store/bookDataStore', () => ({
-  useBookDataStore: () => ({ getConfig: () => ({ booknotes: mockBooknotes }) }),
-}));
+// Production code uses per-field selectors; mock must apply them.
+vi.mock('@/store/bookDataStore', () => {
+  const state = { getConfig: () => ({ booknotes: mockBooknotes }) };
+  return {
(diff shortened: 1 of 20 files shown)


loveheaven:
Make this PR to prevent page turn animation get stuck. Lots of components which depends on progress will need update after you turn a page no matter the state is changed or not.

PR #4561, opened 2026-06-12
perf(cfi): bucket booknotes per chapter and batch-collapse location matcher

loveheaven:
## Summary

Each `relocate` event recomputed `booknotes.filter(...isCfiInLocation(item.cfi, location))` over the full annotation list. For a book with hundreds of annotations spread across many chapters this walked the whole array on every page turn, and `isCfiInLocation` itself parsed both CFIs on every comparison.

## Fix

Two complementary changes in `utils/cfi.ts` and the callers (`Annotator`, `useBooknotesNav`, `useSearchNav`):

- New `getCfiSpinePrefix(cfi)` extracts the spine-pos prefix (e.g. `epubcfi(/6/8`) for cheap chapter-level keying.
- New `createCfiLocationMatcher(location)` parses the current location's start/end CFIs once and returns a matcher closure. Callers can build a per-chapter bucket of booknotes and only run the matcher against the bucket whose prefix matches the current location's prefix.
- `Annotator`'s `progress` effect, `useBooknotesNav`, and `useSearchNav` now build/use a `Map<spinePrefix, BookNote[]>` keyed bucket, so the per-relocate filter touches only the chapter's worth of notes instead of the whole book.

## Why this matters

Builds on the progress-store split (#4557, now merged). Once `useBookProgress` is the per-cell reactive surface, `Annotator` re-runs its progress effect on every page turn. The filter step is the dominant cost in that effect for users with large annotation sets — `relocate` becomes O(notes-in-chapter) instead of O(all-notes), and CFIs in unrelated chapters are no longer parsed at all.

## Testing

All existing tests pass (`pnpm test` — 5299 passed). New unit tests in `cfi.test.ts` cover `getCfiSpinePrefix` edge cases (null/empty/multi-step CFIs).

### DIFF
--- a/apps/readest-app/src/__tests__/utils/annotation-index.test.ts
+++ b/apps/readest-app/src/__tests__/utils/annotation-index.test.ts
@@ -1,0 +1,94 @@
+import { describe, it, expect } from 'vitest';
+import {
+  buildAnnotationIndex,
+  selectLocationAnnotations,
+} from '@/app/reader/utils/annotationIndex';
+import { uniqueId } from '@/utils/misc';
+import { BookNote } from '@/types/book';
+
+describe('buildAnnotationIndex / selectLocationAnnotations', () => {
+  const note = (overrides: Partial<BookNote>): BookNote => ({
+    id: uniqueId(),
+    type: 'annotation',
+    cfi: 'epubcfi(/6/6!/4/2/2,/1:0,/1:10)',
+    note: '',
+    createdAt: 1,
+    updatedAt: 1,
+    ...overrides,
+  });
+
+  // A relocate always reports a single-section location (foliate joins the
+  // section base CFI with the in-document range), so the spine prefix keys
+  // the bucket. `epubcfi(/6/6)` covers everything anchored in section /6/6.
+  const location = 'epubcfi(/6/6)';
+
+  it('buckets a styled annotation under its spine prefix', () => {
+    const highlight = note({ style: 'highlight', color: 'yellow' });
+    const { bySection, globals } = buildAnnotationIndex([highlight]);
+    expect(bySection.get('/6/6')).toEqual([highlight]);
+    expect(globals).toEqual([]);
+  });
+
+  it('buckets a note-only annotation (no style) so its bubble survives relocate', () => {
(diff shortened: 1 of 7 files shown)


loveheaven:
I have more than 1100 annotations in one book. So when I turn a page of the book, the animation gets stuck. That's why I make this PR.

chrox:
Please rebase this branch on to the origin main so that we can review it easier.

loveheaven:
> Please rebase this branch on to the origin main so that we can review it easier.


Done. 

Sorry. Did you push any commit into my branch? Sorry. I forced-pushed this branch.... Could you please push it again?

chrox:
Never mind, I just rebase the branch via the GitHub merge panel.

chrox:
Reviewed and pushed a follow-up fix (8961c15) for a behavior regression in the bucketing.

**Issue:** The per-chapter bucket gated entries on `item.style`, but the original `Annotator` effect maintained **two independent filters** — `annotations` (required `style`) and `notes` (required only a non-empty `note`, *not* `style`). Note-only annotations (a `note` with no highlight style/color, e.g. created via the Notebook "add note" flow — `Notebook.tsx::handleSaveNote`) were therefore dropped from the bucket entirely, so their note bubble (`Overlayer.bubble`, drawn for any annotation with `.note`) was no longer re-applied on the relocate / booknotes-change path. Impact is narrow because `onCreateOverlay`, import, and primary-section sync still draw bubbles, but a note-only annotation arriving via background sync into an already-rendered (non-primary) section would not reappear until that section re-rendered.

**Fix:** Restore the original two-list semantics — bucket on `style` **OR** non-empty `note`, then classify per location (`annotations` needs a `style`, `notes` needs a `note`; globals still require a `style`). Single-section invariant of `progress.location` (foliate's `joinIndir(baseCFI, fromRange(range))`) is preserved, so the single-bucket lookup stays correct.

Also extracted the inline logic into a unit-tested `annotationIndex` module (`buildAnnotationIndex` + `selectLocationAnnotations`) under `reader/utils/`, with regression tests in `annotation-index.test.ts` (verified red against the old `style`-only gate).

`pnpm test` (5308 passed) and `pnpm lint` (tsgo + biome) both green.

PR #4562, opened 2026-06-12
perf(reader): coalesce relocate events and memoize BookCell to stop per-swipe storm

loveheaven:
## Summary

Two complementary changes that stop a swipe burst from re-rendering the full reader tree on every intermediate `relocate`.

## FoliateViewer — coalesce relocate

foliate fires `relocate` multiple times during a swipe burst (snap steps plus the final stabilize). Each one ended up in `setProgress`, which writes to `readerProgressStore` plus `bookDataStore`. This change coalesces them into a single commit per animation frame via `requestAnimationFrame`, so only the final viewport state is persisted.

An earlier iteration used `requestIdleCallback`, but profiling on Android showed "Fire Idle Callback" ballooning to ~2 s of total time per ~28 s session: rIC backed up under sustained pressure and dumped the whole queue into the post-swipe pause, producing exactly the "feels sluggish right after I let go" jank the change was trying to fix. `rAF` runs once per frame, gets scheduled by the browser's normal vsync loop, and doesn't accumulate when the page is busy.

## BooksGrid → BookCell

Previously `BooksGrid` subscribed to the entire `progresses` map and rendered all per-book chrome inline, so every page turn re-rendered every cell in the grid (cheap but multiplicative when parallel views are open).

Extract a memoized `BookCell` component that:

- subscribes only to its own book's progress via `useBookProgress(bookKey)` — a page turn re-renders one cell;
- consumes per-field store selectors so action identities are stable and `React.memo` doesn't get defeated;
- receives memoized `gridInsets` / `contentInsets` from `BooksGrid` so memoization on prop identity is meaningful.

## Why this matters

Pairs with the progress-store split (#4557). #4557 isolates the high-frequency write source from the wider reader store; this PR makes sure the now-tight reactive surface only fires on full-frame transitions and that only one cell at a time gets the update.

## Depends on #4557

This PR's first commit duplicates #4557 (split progress store) so the second commit can compile in isolation (it uses `useBookProgress`). Once #4557 lands, GitHub will collapse the diff to only the coalesce / BookCell changes. Please merge #4557 first.

## Testing

All existing tests pass (`pnpm test` — 5288 passed).

### DIFF
--- a/apps/readest-app/src/app/reader/components/BooksGrid.tsx
+++ b/apps/readest-app/src/app/reader/components/BooksGrid.tsx
@@ -89,27 +89,32 @@ const BookCellInner: React.FC<BookCellProps> = ({
   onGoToLibrary,
 }) => {
   // Per-field selectors — see store/readerProgressStore.ts header for the
-  // "destructure-subscribes-the-whole-store" rationale. Without these,
-  // every BookCell would re-render per page turn and defeat React.memo.
+  // "destructure-subscribes-the-whole-store" rationale.
   const getConfig = useBookDataStore((s) => s.getConfig);
   const getBookData = useBookDataStore((s) => s.getBookData);
-  const getViewState = useReaderStore((s) => s.getViewState);
-  const getViewSettings = useReaderStore((s) => s.getViewSettings);
 
-  // Per-cell reactive subscription. Only re-renders when THIS book's
-  // progress changes, which is the only piece of data we need from
-  // the progress store for the section / pageinfo / sectionLabel
-  // props passed down to SectionInfo and FooterBar below.
+  // Per-cell reactive subscriptions. This cell re-renders when THIS book's
+  // progress changes (page turns) OR its view state changes. Both are
+  // needed: viewState carries `viewSettings` and `ribbonVisible`, which
+  // gate the chrome this cell mounts (Show Header / Show Footer, Double
+  // Border, bookmark Ribbon). Those settings save with applyStyles=false
+  // and the ribbon toggle writes no progress, so without a viewState
+  // subscription the toggles wouldn't take effect until the next page turn.
+  //
+  // Subscribing to the per-book slice is safe now that progress lives in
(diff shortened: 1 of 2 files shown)


chrox:
Reviewed and pushed a follow-up fix (`714c10a0d`).

**Issue:** the `BookCell` extraction subscribes reactively only to `useBookProgress(bookKey)` and reads `viewState`/`viewSettings` imperatively. On current `main` the grid re-rendered on *any* `readerStore` change (whole-store destructure subscription), which is what kept the per-book chrome in sync. After this PR, per-book state that doesn't flow through the progress store no longer re-renders the cell:

- `Show Header` / `Show Footer` and `Double Border` / `Border Color` save via `saveViewSettings(..., applyStyles=false)` → no `setStyles` → no `relocate` → no progress write.
- Bookmarking writes `setBookmarkRibbonVisibility` → `viewStates[key]` only, no progress.

Since `SectionInfo` / `ProgressBar` / `Ribbon` are mount-gated inside `BookCell` (and don't self-gate), toggling those left stale chrome until the next page turn.

**Fix:** `BookCell` now also subscribes to its own `viewStates[bookKey]` slice. This is safe precisely because #4557 moved progress into its own store — `viewStates[key]` now bumps only on low-frequency events (settings toggles, ribbon, init, sync), never on the per-swipe relocate path — so it doesn't reintroduce the commit storm the split removed.

The `FoliateViewer` rAF coalescing looks correct (stable refs/callback, register-once wiring, unmount flush; `previewMode` clearing is on the untouched renderer-relocate path).

`pnpm test` (5288 passed) and `pnpm lint` (tsgo + Biome) pass.

Minor, left as-is: `ReaderContent`'s `handleCloseBook` / `handleCloseBooksToLibrary` aren't `useCallback`'d, so `React.memo(BookCell)` is defeated whenever `ReaderContent` re-renders. It's off the per-page-turn path, and memoizing it correctly also requires stabilizing the `throttle(...)`, so I left it.

Issue #3742, opened 2026-04-04
Brightness/font size settings not showing on Android Tablet


tfriedel:
## Bug Description

On my Xiaomi Pad Mini tablet I don't see buttons for adjusting brightness, font size etc at the bottom of the screen like I do on my Pixel 10. I click in the middle of the page to make the bottom navigation bar show.

## Expected behavior

See not only a navigation bar and tts icon but also icons for changing font size, brightness etc like I do on my Pixel.

## Technical inputs

Operating System: Android 16 WebView 146.0.7680.177 , Xiaomi hyper os 3.0.5.0

Readest Version: 0.10.1

![Screenshot_2026-04-04-12-23-54-092_com.bilingify.readest.jpg](https://github.com/user-attachments/assets/db83603d-247b-4dc3-a945-019ffb059018)

PR #4598, opened 2026-06-15
fix(footer): keep reader toolbar stable across orientation changes

loveheaven:
## Problem

The bottom footer bar swaps layouts whenever the user rotates the device:

- portrait → mobile bar (TOC / color / progress / font / TTS panels)
Screenshot_20260613_154011_com bilingify readest
- landscape → desktop bar (chevron navigation + range slider)
Screenshot_20260613_154332_com bilingify readest

- rotate again → flip back

That's an unintentional side-effect of #3759, which made `forceMobileLayout` true only when `innerWidth <= innerHeight`:

```ts
const forceMobileLayout =
  !!appService?.isMobile && window.innerWidth >= 640 && window.innerWidth <= window.innerHeight;
```

The aspect-ratio guard solved the original tablet-in-portrait regression from #3742 / #3746, but it also keys the entire toolbar off the current viewport aspect ratio — so any mobile device (phone, tablet, foldable) re-skins its footer on every rotation. Users expect rotating the device to leave the toolbar alone.

## Fix

Drop the orientation guard. `forceMobileLayout` becomes simply:

```ts
const forceMobileLayout = !!appService?.isMobile;
```

Every iOS / Android instance keeps the mobile footer bar regardless of orientation, which is the more useful default because that bar exposes brightness / font / color / progress panels that the desktop layout omits.

Desktop / web is unaffected — `appService.isMobile` is false there, and the `sm:` (640px) Tailwind breakpoint still collapses a narrow window to the mobile layout the same way it always did.

| Surface | Before | After |
| --- | --- | --- |
| Phone portrait | mobile bar | mobile bar (unchanged) |
| Phone landscape | desktop bar | **mobile bar** (changed) |
| Tablet / foldable portrait | mobile bar (#3759) | mobile bar (unchanged) |
| Tablet / foldable landscape | desktop bar | **mobile bar** (changed) |
| Desktop / web | desktop bar (or mobile under `sm:`) | unchanged |

## Open question for reviewers

#3759 was merged with an explicit contract: `forceMobileLayout` must be false for phones (`innerWidth < 640`) so that all `!forceMobileLayout && 'sm:…'` overrides evaluate to the original class strings and `MobileFooterBar`'s slide-down animation behaves exactly as before #3746. That contract was the reason #3759 did not re-introduce the phone panel animation regression that got #3746 reverted.

This change makes `forceMobileLayout` true on phones too, so it crosses that boundary. In local testing I did not observe a phone-panel-animation regression — by my reading, since #3759 the `MobileFooterBar` subtree no longer wraps children in the extra `<div>` that broke #3746, so the `forceMobileLayout=true` branch is now safe on phones as well. But I have only tested on Android. I'd appreciate confirmation on:

1. iOS phone portrait + landscape — does the bottom panel still slide cleanly behind the navigation bar on dismissal?
2. Is there any other place where the codebase still relies on `forceMobileLayout=false` on phones (CSS, layout, hit-testing) that I missed?

If either turns up an issue, an alternative scope would be `appService.isMobile && innerWidth >= 640` (tablets/foldables in both orientations, phones unchanged) — that keeps the #3759 contract intact while still removing the orientation flip on the larger devices.

## Verification

- `pnpm lint` clean
- Manual: Android phone portrait/landscape, Android tablet portrait/landscape — toolbar stays on the mobile layout across rotation in all four states; desktop window unchanged

### DIFF
--- a/apps/readest-app/src/app/reader/components/footerbar/FooterBar.tsx
+++ b/apps/readest-app/src/app/reader/components/footerbar/FooterBar.tsx
@@ -195,12 +195,28 @@ const FooterBar: React.FC<FooterBarProps> = ({
   const footerBarRef = useRef<HTMLDivElement>(null);
   useSpatialNavigation(footerBarRef, isVisible);
 
-  // Force the mobile footer bar on mobile tablets/foldables in portrait mode
-  // where the viewport width exceeds the `sm:` (640px) breakpoint. Phones
-  // (innerWidth < 640) are intentionally excluded so their styling and panel
-  // slide-down animation remain exactly as before — see #3742 / #3746.
-  const forceMobileLayout =
-    !!appService?.isMobile && window.innerWidth >= 640 && window.innerWidth <= window.innerHeight;
+  // Always use the mobile footer bar on mobile devices (phones, tablets, and
+  // foldables on iOS / Android). Previously this flag also factored in the
+  // current viewport's portrait/landscape state — it was only true when
+  // `innerWidth <= innerHeight`. That made the entire bottom toolbar swap
+  // contents on every orientation change: a phone in landscape (or a
+  // foldable in landscape) would suddenly drop the brightness / font /
+  // color / progress panels and render the desktop layout instead, then
+  // swing back to the mobile panels when rotated to portrait. Users
+  // expect rotating the device to leave the toolbar alone, so we no
+  // longer key off orientation here.
+  //
+  // Phones in landscape (innerWidth >= 640, innerHeight < 640) used to
+  // fall through to the `sm:` breakpoint and pick up the desktop layout;
+  // they now stay on the mobile layout, which is the more useful default
+  // because the mobile layout still exposes the per-book panels (this
+  // is the same fix #3759 applied for the tablet-portrait case, just
+  // generalized to all mobile orientations).
+  //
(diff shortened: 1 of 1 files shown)


chrox:
> That's an unintentional side-effect of https://github.com/readest/readest/pull/3759

No this is intentional. Previously on both portrait and landscape of iPad/tablets it will use the desktop-style of bottom bar, now on portrait it will use the mobile style bottom bar as requested in #3742, but on landscape of these devices it will still use the desktop-style which is more natural on wider screen.

loveheaven:
> > That's an unintentional side-effect of #3759
> 
> No this is intentional. Previously on both portrait and landscape of iPad/tablets it will use the desktop-style of bottom bar, now on portrait it will use the mobile style bottom bar as requested in #3742, but on landscape of these devices it will still use the desktop-style which is more natural on wider screen.

Okay. But I don't understand it. Why should I use desktop-style footbar? I want to be able to adjust brightness/font/toc on my Phone no matter it is on landscape or portrait.

chrox:
The mobile-style bottom bar may need to grow vertically, which requires more screen height. Otherwise, on some phones in landscape mode, the bar could be partially clipped, resulting in a poor user experience. A better solution might be to add screen brightness controls to the desktop-style bottom bar and keep the functionality of both styles consistent.

loveheaven:
okay. I understand it now. I will close this PR.

PR #4627, opened 2026-06-17
fix(library): never let a routine save shrink library.json (cold-start "Open with" wipe)

chrox:
## Problem

Opening a book via Android **"Open with"** on a **cold start** could wipe the entire library — and it stayed wiped after a normal restart.

**Root cause** (a three-PR interaction):
- **#4407** introduced `openTransient`, which builds an ephemeral entry from `useLibraryStore.getState().library`. On a cold start the store isn't populated yet, so it builds on top of an *empty* library.
- The library page's load-skip gated on `libraryBooks.length > 0` (#4325), so the now-non-empty store made it **skip `loadLibraryBooks()`** — the real library is never read from disk.
- A later `saveLibraryBooks` then persists the empty/partial set, and `safeSaveJSON` overwrites **both** `library.json` *and* `library.json.bak`.
- **#4527** (reliable cold-start intent delivery) made this reliably reproducible; released **v0.11.4 lacks #4527**, which is why it does not reproduce there.

## Fix (two layers of defense)

**1. Write-level floor — `saveLibraryBooks` can no longer shrink `library.json`.**
A routine save now merges with the on-disk library (union by hash, incoming wins): it can add or modify rows (including `deletedAt` tombstones) but **never drops a book**. Deliberate, authoritative rewrites (restore, tombstone GC, account reset) opt in via a new `{ replace: true }`. This layer alone makes the wipe impossible regardless of the caller.

**2. Stay out of the bad state.**
- `openTransient` loads the real library from disk *before* importing a transient book. This also fixes a secondary cold-start bug: the "already-imported?" hash-match ran against the empty store, so previously-imported books were re-imported as transient copies.
- The library page's load-skip now gates on the store's `libraryLoaded` flag instead of `length > 0`.

Compatible with #4626 "Purge Data", which tombstones via `updateBook` (a modify, not a shrink) and so needs no `replace`.

## Testing
- New `saveLibraryBooks` suite cases: never-drop (merge floor), no-wipe on empty save, tombstone preserved, incoming-wins, plus both `{ replace: true }` paths.
- Full unit suite green (`pnpm test`), `tsgo --noEmit` + Biome clean.
- Built and installed on a Xiaomi (Android) for manual verification of the cold-start "Open with" flow.

### DIFF
--- a/apps/readest-app/src/__tests__/services/suites/library-tests.ts
+++ b/apps/readest-app/src/__tests__/services/suites/library-tests.ts
@@ -90,12 +90,12 @@ export function libraryTests(getService: () => AppService) {
       expect(loaded).toEqual([]);
     });
 
-    it('should overwrite previous library on save', async () => {
+    it('should overwrite (and shrink) the library only when replace is set', async () => {
       const service = getService();
       await service.saveLibraryBooks([makeBook({ hash: 'old' })]);
 
       const newBooks = [makeBook({ hash: 'new1' }), makeBook({ hash: 'new2' })];
-      await service.saveLibraryBooks(newBooks);
+      await service.saveLibraryBooks(newBooks, { replace: true });
 
       const loaded = await service.loadLibraryBooks();
       expect(loaded).toHaveLength(2);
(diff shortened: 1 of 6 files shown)


chrox:
@loveheaven Checkout this.

Issue #4643, opened 2026-06-18
Resdest0.11.4版，出现epub格式电子书封面丢失和无法全屏显示问题


wjx5228:
样书里的封面两种代码，之前的Resdest版本，都能完整加载显示，实现封面加载。
但升级到0.11.4版本后，样书B的代码模式，封面无法显示；样书A的代码模式无法全屏显示。
特此提供样书反馈！

https://wwamr.lanzoum.com/ixvUW3l4tpeh
密码:cwe8

wjx5228:
https://wwamr.lanzoum.com/itjXh3sb22re

wjx5228:
样书测试，在这个链接，前文的链接发错了！
https://wwamr.lanzoum.com/itjXh3sb22re

chrox:
Both are fixed in the latest code and the fixes will be available in the next update.

wjx5228:
辛苦大佬！

wjx5228:
新版0.11.10，仍没有修复封面问题，特此反馈！！

chrox:
Please share the screenshots of the covers also the screenshot of your About Readest page.

wjx5228:
https://wwamr.lanzoum.com/iWg113se9lbc

wjx5228:
不是一本书，
更新后，
是所有电子书都出现这问题了！
影响阅读体验。
还请及时修复！

chrox:
Please paste screenshots here directly. I cannot download the zip package you uploaded.

wjx5228:
<img width="1260" height="2844" alt="Image" src="https://github.com/user-attachments/assets/da9c2d90-4dbb-4a09-8c17-fb332ce538fd" />

chrox:
The sample A book is expected, now the image will fit the screen other than stretched to full screen. Please adjust the aspect ratio of image if it cannot cover the full screen.

chrox:
I can reproduce the sample B rendering issue now.

chrox:
> 是所有电子书都出现这问题了！

Does it happen all in the sample B case?

wjx5228:
<img width="1260" height="2844" alt="Image" src="https://github.com/user-attachments/assets/0e62f71d-1ffe-4b07-94c3-0bbe59cbc3dc" />

wjx5228:
更新后，都不行

wjx5228:
正常的话，
封面都是全屏显示，
不会不显示和无法全屏显示的

[...]

Issue #4379, opened 2026-05-31
In scroll mode, the cover image cannot be displayed.


riceteasun:
## Bug Description
>In scroll mode, the cover image cannot be displayed. Switching back to page-turning mode allows the cover to appear normally.

## Technical inputs
Operating System:
origin OS 6

Readest Version:
> Operating System: Android 16 (WebView 148.0)
> Readest Version: 0.11.2

chrox:
Please share the EPUB file that has the problem.

riceteasun:
> Please share the EPUB file that has the problem.

In fact, this happens with every EPUB file I've tested.
[epub](https://cf-pan.acgdmzy.com/s/BYKfP)

PR #4679, opened 2026-06-20
fix(reader): stretch fullscreen cover to fill the page

chrox:
## Summary
Paginated Duokan full-page covers (`data-duokan-page-fullscreen`) were rendered with `object-fit: contain`, which letterboxes the cover when its aspect ratio doesn't match the page. This bumps `foliate-js` so they render with `object-fit: fill` instead — the cover fills the whole page, **distorting to fit** when the aspect ratio differs, matching Duokan's native full-page render.

Scoped strictly to the paginated fullscreen branch of `setImageSize()`:
- Ordinary inline images everywhere keep `object-fit: contain`.
- Scrolled-mode Duokan covers keep `contain` bounded by `max-height` (the #4379 flow is untouched — there's no fixed page height to fill in scrolled mode).

## foliate-js change
Depends on readest/foliate-js#33 (submodule pointer bumped to its branch SHA `d2f7f98`). The pointer will need re-bumping to the squash-merge SHA once that PR lands.

## Test
Adds a browser test (`paginator-duokan-cover.browser.test.ts`) asserting the fullscreen cover computes `object-fit: fill`. Verified test-first: it fails as `contain` without the foliate-js change and passes with it.

- `pnpm test` — 5884 passed
- `pnpm lint` (tsgo + Biome) — clean

### DIFF
--- a/apps/readest-app/src/__tests__/document/paginator-duokan-cover.browser.test.ts
+++ b/apps/readest-app/src/__tests__/document/paginator-duokan-cover.browser.test.ts
@@ -104,6 +104,25 @@ describe('Paginator Duokan fullscreen cover (#4379)', () => {
     expect(img!.offsetHeight).toBeGreaterThan(0);
   });
 
+  it('stretches the fullscreen cover to fill the page (object-fit: fill)', async () => {
+    paginator = createPaginator();
+    paginator.open(book);
+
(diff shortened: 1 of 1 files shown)


chrox:
This PR fixes https://github.com/readest/readest/issues/4643.

PR #4696, opened 2026-06-21
fix(library): preserve original files when deleting "read in place" books

chrox:
## Problem

A user imported a folder via **Import From Directory** with **Read books in place**, then later deleted those books from the library. Readest **permanently deleted the user's original files on disk** — they were not moved to the Recycle Bin, and cloud sync had not yet uploaded them ("Book File Not Uploaded"), so they were unrecoverable.

"Read in place" books are read straight from their original location (`book.filePath`, base `'None'`) instead of being copied into `Books/<hash>/`. That source file belongs to the user; deleting the library entry must never destroy it.

## Root cause

`deleteBook` in `src/services/cloudService.ts` resolved the book's content source and, when the source was `external` (an in-place or transiently-opened book), unconditionally called `fs.removeFile()` on it for `local`/`both`/`purge` deletes.

This behavior was previously **deliberate and tested** — the in-place test block asserted the source file *was* removed, rationalized as "symmetric with deleting a normal book's managed copy." That assumption is wrong for files Readest never copied.

## Fix

`deleteBook` now only removes files Readest itself created:

- the **managed** copy under `Books/<hash>/`, and
- app-generated **sidecars** (`cover.png` on `both`; the whole `Books/<hash>/` dir on `purge`).

`external` sources (`book.filePath`, base `'None'` — in-place and transient imports) are never touched. Also corrected the misleading `readInPlace` JSDoc in `ImportFromFolderDialog.tsx` that documented the destructive behavior as intended.

## Tests

`src/__tests__/services/cloud-service.test.ts` — flipped the in-place assertions to require the source file is **preserved** across `local`/`both`/`purge`, while sidecar removal (cover on `both`, dir wipe on `purge`) is unchanged. Confirmed red before the fix, green after. Full file: 28/28 pass.

## Follow-ups (not in this PR)

The original report also noted the deletion flow lacks any warning/disclaimer, and deletions bypass the OS Recycle Bin even for managed copies. Worth addressing separately.

### DIFF
--- a/apps/readest-app/src/__tests__/services/cloud-service.test.ts
+++ b/apps/readest-app/src/__tests__/services/cloud-service.test.ts
@@ -216,7 +216,7 @@ describe('cloudService', () => {
         expect(book.uploadedAt).toBe(1000);
       });
 
-      test('removes the in-place source file and the sidecar dir for in-place books', async () => {
+      test('wipes the sidecar dir but NEVER the in-place source file', async () => {
         const book = createMockBook({ filePath: '/Users/me/Library/sample.epub' });
         vi.mocked(mockFs.exists).mockImplementation(async (path, base) => {
           if (base === 'None' && path === book.filePath) return true;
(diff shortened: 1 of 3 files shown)


chrox:
@loveheaven We will never remove files not owned/copied by the Readest app, so when deleting in place imported books we won't delete the original book files, in order to not surprise users. FYI.

PR #4779, opened 2026-06-25
fix(reader): stop a quick-deleted highlight from being re-drawn (#4773)

### LINKED ISSUE #4773: instantly highlighted text bug
## Bug Description

When text is instantly highlighted and the mark is deleted within a very short time, the mark remains visible on the page and only disappears after reopening the book.

> Operating System: Android 14 (WebView 145.0)
> Readest Version: 0.11.12-20260625

chrox:
## Summary

Fixes #4773. Deleting a highlight within a very short time left the mark painted on the page; it disappeared only after reopening the book.

## Root cause

`Annotator` re-draws on-page annotations after every relocate from a memoized `annotationIndex`, via `selectLocationAnnotations` then `view.addAnnotation`. `buildAnnotationIndex` filters `deletedAt` when the index is built, but `selectLocationAnnotations` trusted that and never re-checked.

Deleting a highlight (`handleHighlight(false)`) stamps `deletedAt` in place on the booknote object that is still sitting in the index bucket, and removes the overlay. `Annotator` does not re-render on booknote changes (it only subscribes to the stable `getConfig`), so the memo stays stale. If the re-apply effect scheduled from the popup-open render flushes after the delete (the "very short time" window, with passive effects deferred under rapid taps), it reads the deleted object from the pre-deletion snapshot and re-draws the overlay, orphaning it until reopen. Instant Highlight just makes the race easy to hit; the delete plus re-apply path is shared with normal highlights.

## Fix

Re-check `deletedAt` at the read site, not just at index build:

- `selectLocationAnnotations` (annotationIndex.ts): skip `deletedAt` before classifying, covering both highlights and notes.
- The sibling `annotationIndex.globals` re-apply loop in `Annotator` gets the same guard (global highlights have the identical stale-snapshot hazard).

## Testing

- Unit: `annotation-index.test.ts` builds an index, marks the booknote `deletedAt` in place, and asserts `selectLocationAnnotations` returns nothing. Red before the fix, green after.
- `pnpm test` (full suite) and `pnpm lint` pass.
- On a Xiaomi 13 Pro (CDP lane, fixed build): real create then delete then immediate page-turn over 4 iterations left overlay count 6 -> 7 -> 6 each time (no orphan); a stray-overlay probe confirmed the counter is not blind.

### DIFF
--- a/apps/readest-app/src/__tests__/utils/annotation-index.test.ts
+++ b/apps/readest-app/src/__tests__/utils/annotation-index.test.ts
@@ -54,6 +54,22 @@ describe('buildAnnotationIndex / selectLocationAnnotations', () => {
     expect(globals).toEqual([]);
   });
 
+  it('skips an annotation deleted in place after the index was built (stale index)', () => {
+    // #4773: the re-apply effect holds a memoized index. A quick delete stamps
+    // `deletedAt` on the SAME object that is still sitting in the bucket (the
+    // memo has not recomputed yet). If the read trusts the build-time filter,
+    // the just-deleted highlight gets re-drawn and its overlay is orphaned —
+    // visible on the page until the book is reopened. The read must re-check.
+    const highlight = note({ style: 'highlight', color: 'yellow', note: 'hi' });
+    const index = buildAnnotationIndex([highlight]);
+    // The user deletes the highlight: `deletedAt` is stamped in place on the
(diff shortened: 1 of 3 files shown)


ybbluesky:
0.11.20-2026072506 (WebView 147.0.7727.137)，the bug again

PR #4803, opened 2026-06-26
fix(settings): keep global settings in sync across windows (#4580)

### LINKED ISSUE #4580: Pagination settings are occasionally reverted to default settings
## Bug Description

Readest keeps re-enabling pagination settings. In my specific example: click to paginate keeps re-enabling. I have not yet managed to reliably predict when it happens, but over the course of a day I will suddenly notice it reverted to "not what I set it to".

## Expected behavior

I expect readest to keep the settings as: Click to Paginate (off), Swipe to Paginate (off), Show Page Navigation Buttons (on) when I set it to this.

## Technical inputs

Operating System: Win 11

Readest Version: Version 0.11.4 (Edge 149.0.0.0)

chrox:
## Problem

Fixes #4580. On desktop, global view settings (Click to Paginate, Swipe to Paginate, Show Page Navigation Buttons) occasionally revert to their defaults. As the reporter confirmed, this only happens with **multiple windows open** (`1 + n_open_books` windows).

## Root cause

Each Tauri window (library + one per open book) keeps its own **in-memory** `useSettingsStore.settings`, loaded once when the window opens. Global settings persist to a single shared `settings.json`, and every window writes the **whole** object via the store's `saveSettings`.

A reader window opened *before* the user customized pagination still holds the default `disableClick=false`. When that window later saves — most visibly `handleCloseBooks` on window close, but any settings write does it — it clobbers the user's `disableClick=true` back to the default. That matches the symptom exactly: reverts to *default*, intermittently, only with multiple windows.

## Fix

Cross-window broadcast so no window holds stale globals:

- **`src/utils/settingsSync.ts`** (new) — `broadcastGlobalSettings` emits `globalViewSettings`/`globalReadSettings` (with the source window label) to all windows; `subscribeSettingsSync` ignores the emitter's own echo; `mergeSyncedGlobalSettings` adopts the two global blobs while **preserving every device/window-local field** (filesystem paths, `lastOpenBooks`, sync cursors, brightness). No-op off Tauri.
- **`store/settingsStore.ts`** — `saveSettings` broadcasts after persisting.
- **`hooks/useSettingsSync.ts`** (new), mounted in **`Providers.tsx`** (shared root for both window types) — adopts incoming broadcasts via `setSettings`.

Scope is intentionally minimal: only the two global Settings-dialog objects sync (covers the reported pagination bug + sibling read settings). The receive path only updates the in-memory store, so there is no save/broadcast loop; the replica publisher pushes to the network only (no disk write), and pagination fields are not in `SETTINGS_WHITELIST` anyway. Live cross-window view refresh of already-open books is out of scope (the bug is persistence, not live propagation).

## Testing

- Test-first: added a failing test for the merge logic, then implemented.
- New `src/__tests__/utils/settings-sync.test.ts` (merge adopts globals, preserves device-local fields, no mutation) + extended store test (`saveSettings` persists and broadcasts).
- `pnpm test` (6339 passed), `pnpm lint` (tsgo + Biome), and `pnpm format:check` all clean.

### DIFF
--- a/apps/readest-app/src/__tests__/store/settings-store.test.ts
+++ b/apps/readest-app/src/__tests__/store/settings-store.test.ts
@@ -10,13 +10,20 @@ vi.mock('@/utils/time', () => ({
   initDayjs: vi.fn(),
 }));
 
+vi.mock('@/utils/settingsSync', () => ({
+  broadcastGlobalSettings: vi.fn(),
+}));
+
 import i18n from '@/i18n/i18n';
 import { initDayjs } from '@/utils/time';
+import { broadcastGlobalSettings } from '@/utils/settingsSync';
 import { useSettingsStore } from '@/store/settingsStore';
+import type { EnvConfigType } from '@/services/environment';
 import type { SystemSettings } from '@/types/settings';
 
 const mockChangeLanguage = vi.mocked(i18n.changeLanguage);
 const mockInitDayjs = vi.mocked(initDayjs);
+const mockBroadcastGlobalSettings = vi.mocked(broadcastGlobalSettings);
 
 function makeSettings(overrides: Partial<SystemSettings> = {}): SystemSettings {
   return {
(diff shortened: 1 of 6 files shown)


ivannunyadambiz:
thanks guys!

PR #4819, opened 2026-06-27
fix(node): ensure correct cross-platform path resolution in NodeAppService

### LINKED ISSUE #4815: NodeAppService tests fail due to mixed path separators on Windows
## Bug Description

Running the `readest-app` test suite locally on Windows results in several failing tests in `node-app-service.test.ts` due to mixed path separators when a custom root directory is used.

> **Example:**
> Expected: C:\\...\\Readest\\Books
> Received: C:\\.../Readest/Books
> The path contains both `\` and `/` causing the equality assertion to fail on Windows.

## Expected behavior

`NodeAppService` should return native Windows filesystem paths when running on Windows so that resolved paths are consistent with `path.join()` and the existing test expectations.

I investigated the issue locally and have a working fix that allows the full test suite to pass on Windows while leaving the shared `BaseAppService` unchanged.

Before preparing the PR, I wanted to check whether Windows-native path handling in `NodeAppService` is something worth supporting, since the current CI appears to run only on Ubuntu.

## Technical inputs

Operating System: Windows 11 25H2

Readest Version: main (commit 97868f0)

luiscortesdev:
Fixes #4815 

## Summary
On Windows, `NodeAppService` produced mixed path separators when resolving paths under custom root directories. This caused several test failures in `node-app-service.test.ts` because expected paths used `path.join()` (platform native separators), while implementation concatenated strings using `/`.

## Solution
- Replaced manual string concatenation with `path.join()` in `NodeAppService` path resolution
- Ensured `resolveFilePath` returns platform-native paths on Windows
- Prevented mixed-separator paths when using custom root directories

## Testing
- Added/validated test ensuring resolved paths match platform-native separators on Windows
- All tests in `node-app-service.test.ts` pass on Windows
- Entire test suite passes on Windows (see note below)

## Note
- One unrelated test (`src/__tests__/helpers/updater.test.ts`) fails locally due to an existing syntax error in the repository.
- This issue is unrelated to the changes in this PR and does not affect `NodeAppService` behavior.

## Environment 
Operating System: Windows 11 25H2

### DIFF
--- a/apps/readest-app/src/services/nodeAppService.ts
+++ b/apps/readest-app/src/services/nodeAppService.ts
@@ -109,7 +109,7 @@ const getPathResolver = ({ customRootDir }: { customRootDir?: string } = {}) => 
           'Dictionaries',
         ];
         const leafDir = dataDirs.includes(base) ? '' : base;
-        return leafDir ? `${customRootDir}/${leafDir}` : customRootDir!;
+        return leafDir ? nodePath.join(customRootDir!, leafDir) : customRootDir!;
       }
     : undefined;
 
@@ -120,7 +120,7 @@ const getPathResolver = ({ customRootDir }: { customRootDir?: string } = {}) => 
         return {
           baseDir: 0,
           basePrefix: async () => custom ?? getAppConfigDir(),
-          fp: custom ? `${custom}${fp ? `/${fp}` : ''}` : fp,
+          fp: custom ? nodePath.join(custom, fp) : fp,
           base,
         };
       case 'Cache':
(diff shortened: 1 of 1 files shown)


chrox:
Thanks for the fix, @luiscortesdev! The core change (using `path.join` in the resolver plus the new `resolveFilePath` override) is correct and resolves the mixed-separator paths on Windows.

I pushed a small cleanup (force-update) to keep the PR scoped to the actual fix:

- Dropped the `chore(biome)` commit. The `biome.json` ignore-glob changes and the removed `'use strict'` in `extract-i18n.js` are unrelated to #4815, and lint passes fine without them (they look like artifacts of a different local Biome version).
- Dropped the added 'native platform separators' test. `resolved.split(path.sep).join(path.sep)` is the identity for any string, so it passes regardless of the bug. Coverage already exists: the test suite constructs the service with a custom root dir, so `should resolve file paths correctly` (asserting against `path.join(...)`) exercises exactly the branch this PR fixes and would fail on Windows before the change.

The branch now touches only `nodeAppService.ts` (+18/-14). Full test suite (6382 passed) and lint pass locally.

PR #4883, opened 2026-07-02
fix(nix): get nix devshell working

dastarruer:
I've run into some issues while trying to build readest using the nix flake.

Firstly, when running `pnpm tauri dev`, this is what the app looks like:

 image

Secondly, the pre-push git hook is broken, as it tries to modify files in the immutable nix store:

```plaintext
 pnpm -C apps/readest-app format:check

$ pnpm -w format:check
$ biome format .
/nix/store/1c15krms4dvjsasbdsxc3inrb0iv8hqy-source/editors/code/package.json format ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Formatter would have printed the following content:
  
       1    1 │   {
       2      │ - ····"name":·"rust-analyzer",
       3      │ - ····"displayName":·"rust-analyzer",
       4      │ - ····"description":·"Rust·language·support·for·Visual·Studio·Code",
       5      │ - ····"private":·true,
       6      │ - ····"icon":·"icon.png",
...
```

Lastly, the devshell does not run `pnpm install`, `pnpm --filter @readest/readest-app setup-vendors`, or `git submodule update --init --recursive` upon entering. The flake also does not set a formatter.  I'd like to add these two things, but have no idea whether the changes would be welcome.

### DIFF
--- a/ops/flake.nix
+++ b/ops/flake.nix
@@ -17,7 +17,7 @@
   outputs = { self, nixpkgs, flake-utils, android, devshell, fenix }:
     {
       overlay = final: prev: {
-        inherit (self.packages.${final.stdenv.hostPlatform.system}) android-sdk android-studio;
+        inherit (self.packages.${final.stdenv.hostPlatform.system}) android-sdk;
       };
     }
     //
@@ -36,8 +36,6 @@
             self.overlay
           ];
         };
-        # android-studio is not available in aarch64-darwin
-        androidConditionalPackages = if pkgs.stdenv.hostPlatform.system != "aarch64-darwin" then [ pkgs.android-studio ] else [ ];
 
         commonPackages = with pkgs; [
           pnpm
@@ -183,9 +181,6 @@
             system-images-android-34-google-apis-x86-64
             system-images-android-34-google-apis-playstore-x86-64
           ]);
-        } // lib.optionalAttrs (system == "x86_64-linux") {
-          # Android Studio in nixpkgs is currently packaged for x86_64-linux only.
-          android-studio = pkgs.androidStudioPackages.stable;
         };
 
         devShells = {
@@ -224,7 +219,7 @@
               pkgs.android-sdk
               pkgs.gradle
               pkgs.jdk
-            ] ++ androidConditionalPackages;
+            ];
             extraEnv = [
               {
                 name = "ANDROID_HOME";
--- a/ops/flake.nix
+++ b/ops/flake.nix
@@ -38,18 +38,12 @@
         };
         # android-studio is not available in aarch64-darwin
         androidConditionalPackages = if pkgs.system != "aarch64-darwin" then [ pkgs.android-studio ] else [ ];
+
         commonPackages = with pkgs; [
           pnpm
           nodejs_24
           clang
           pkg-config
-          (pkgs.fenix.complete.withComponents [
-            "cargo"
-            "clippy"
-            "rust-src"
-            "rustc"
-            "rustfmt"
-          ])
           pkgs.rust-analyzer-nightly
           xdg-utils
           self.formatter.${system}
@@ -97,7 +91,9 @@
 
         mkCommonShell =
           { name
+          , postInit ? ""
           , extraPackages ? [ ]
+          , extraTargets ? [ ]
           , extraEnv ? [
               {
                 name = "PKG_CONFIG_PATH";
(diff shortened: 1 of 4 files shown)


chrox:
Thanks for pointing this out. These sound like reasonable improvements to me.

dastarruer:
Will add them when I get a chance to. However, I’m not sure how to fix the
first two issues, and I’d like some help figuring it out.

chrox:
1. Please check the WebKitGTK version in your environment, this is definitely an issue caused by outdated WebKitGTK version.

2. You might need to git push in an ordinary file system for the git repo. Nix might not be a good choice for development, as I understand?

dastarruer:
1. What webkitgtk version do I need?
2. I was thinking along the lines of ignoring /nix/store paths. I just
don’t know where biome ignores go.

chrox:
1. You might want to use the latest version available.

2. Please check the documentation of biome.

ops/flake.nix
```diff
@@ -14,14 +14,22 @@
...
-  outputs = { self, nixpkgs, flake-utils, android, devshell, fenix }:
+  outputs =
+    {
+      self,
+      nixpkgs,
+      flake-utils,
+      android,
+      devshell,
+      fenix,
+    }:
```
chrox: Is the code re-formatting really necessary?

chrox:
Could you keep the code change minimal to the nix fix so that it works as expected?

chrox (changes requested):

ops/flake.nix
dastarruer: After adding `nixfmt` as the formatter and running `nix fmt flake.nix`, that was the output. I can remove it if you like, but then I'd need to know what formatter the last contributer used. I used `nixfmt` since it's usually the go-to for most projects.

dastarruer:
I asked about the required version of webkitgtk since nixpkgs already pulls in the latest version:

```sh
pkg-config --modversion webkit2gtk-4.1
2.52.4
```

dastarruer:
Also, when pushing, there were four failing tests on my end that were totally unrelated to my changes. Is this something I need to be worried about?

chrox:
Please also check the CI status on GitHub and keep the version of nodejs/pnpm and rust the same version with the CI.

ops/flake.nix
chrox: For now, we might want to remove the formatting to keep this fix simple enough.

dastarruer:
I was going off the `CONTRIBUTING.md` (which was incorrect), so I've updated it to require `v24` instead of `v22`.

dastarruer:
Can I also move `flake.nix` and `flake.lock` to the root of the project? Putting it inside the `./ops` directory is highly unusual. These two files are usually found in the root, not in a subdirectory. If `flake.nix` needs to be split up, those files go inside `./nix`, not `./ops`.

ops/flake.nix
```diff
@@ -118,34 +139,47 @@
...
           pkgs.devshell.mkShell {
             inherit name;
             packages = commonPackages ++ extraPackages;
             env = extraEnv;
+            devshell.startup.init_project.text = ''
+              git submodule update --init --recursive
+
+              if [ ! -d "node_modules" ]; then
+                pnpm install
+                pnpm --filter @readest/readest-app setup-vendors
```
chrox: https://github.com/readest/readest#2-install-dependencies

The `git submodule update --init --recursive` and `pnpm --filter @readest/readest-app setup-vendors` should be a prerequisite that needs to be run before any build script no matter if it's nix env or a bare system env.

chrox (changes requested):

chrox:
> Can I also move `flake.nix` and `flake.lock` to the root of the project? Putting it inside the `./ops` directory is highly unusual. These two files are usually found in the root, not in a subdirectory. If `flake.nix` needs to be split up, those files go inside `./nix`, not `./ops`.

Please keep them under ./ops. These are supporting files for the Nix development environment. Readest is neither a Nix package nor primarily built on or for Nix, and we don’t want to give the impression that it is.

ops/flake.nix
dastarruer: Removed the if statement.

dastarruer:
Is android studio necessary for this project, or can I remove it?

dastarruer:
Running `pnpm tauri icon ../../data/icons/readest-book.png` is necessary for `pnpm tauri android dev` to run without errors. However, the files generated are not ignored by git. Not sure if this is intended.

chrox:
> Is android studio necessary for this project, or can I remove it?

Android studio is not necessary to build for Android.

chrox:
> Running `pnpm tauri icon ../../data/icons/readest-book.png` is necessary for `pnpm tauri android dev` to run without errors. However, the files generated are not ignored by git. Not sure if this is intended.

This is intended.

ops/flake.nix
dastarruer: Btw not sure if it notified you, but I removed the formatting.

chrox:
Looks good to me now.

dastarruer:
Just a disclaimer: I don't have access to a mac, so I can't test the ios shell.

chrox (approved):

PR #4903, opened 2026-07-02
fix(window): enter fullscreen from maximized windows (#4034)

### LINKED ISSUE #4034: Full Screen button does not work in Phosh
## Bug Description

The fullscreen button does nothing in Phosh.
Could not attach video of what is going on. Here is the link https://www.dropbox.com/scl/fi/qwai6ns9ubs2phljw2hla/VID_20260502_014406_01.mp4?rlkey=oxvsr7chuw7x2yf619hg4d4f9&st=2e3fhd5h&dl=0

## Expected behavior

Fullscreen button should make the app fullscreen. This works with all the libadwaita apps and also with Firefox (which does not use libadwaita).

## Technical inputs

Operating System: Debian 14 (Phosh)
Readest Version: Version 0.10.6 (WebView 605.1.15) (installed via Flatpak)

---
Maybe there is a very simple setting which needs to be added to make the app go fullscreen in phosh?

chrox:
## Problem

Fixes #4034. The Fullscreen button did nothing on **Phosh** (Wayland mobile shell), and on **Windows** it only worked when the window was *not* maximized (per the issue comment).

## Root cause

`tauriHandleToggleFullScreen` in `src/utils/window.ts` had an `isMaximized` branch (introduced by #872) that called `unmaximize()` and **never** `setFullscreen()` when the window was maximized:

```js
if (isMaximized) {
  await currentWindow.unmaximize();   // never enters fullscreen
} else {
  await currentWindow.setFullscreen(!isFullscreen);
}
```

Phosh windows are always maximized (it's a mobile shell), so the button always hit the `unmaximize()` branch and appeared to do nothing. On Windows, a maximized window was likewise only unmaximized instead of going fullscreen.

## Fix

Toggle fullscreen unconditionally:

```js
const isFullscreen = await currentWindow.isFullscreen();
await currentWindow.setFullscreen(!isFullscreen);
```

The maximize-button handler (`tauriHandleToggleMaximize`) already exits fullscreen first when in fullscreen, so the two window controls stay consistent without the fullscreen handler special-casing maximize. The Linux transparent-background workaround is untouched, and `isMaximized` is still used elsewhere (`themeStore` rounded-window), so no capability changes are needed.

## Testing

- Added a `tauriHandleToggleFullScreen` block to `src/__tests__/utils/window.test.ts` covering the maximized (Phosh/Windows), already-fullscreen, and plain cases. The maximized-window test fails against the old code and passes with the fix.
- `pnpm test` (window suite) and `pnpm lint` (tsgo + biome) pass.

Note: I don't have Phosh hardware to reproduce directly, so this is validated by unit tests and reasoning about the tao/winit fullscreen path. The reporter confirmed the compositor supports fullscreen (libadwaita apps and Firefox work), so entering it via `setFullscreen(true)` should now take effect.

### DIFF
--- a/apps/readest-app/src/__tests__/utils/window.test.ts
+++ b/apps/readest-app/src/__tests__/utils/window.test.ts
@@ -24,7 +24,7 @@ vi.mock('@/utils/event', () => ({
 
 import { getCurrentWindow } from '@tauri-apps/api/window';
 import { type as osType } from '@tauri-apps/plugin-os';
-import { tauriHandleOnCloseWindow } from '@/utils/window';
+import { tauriHandleOnCloseWindow, tauriHandleToggleFullScreen } from '@/utils/window';
 
 type CloseHandler = (event: { preventDefault: () => void }) => Promise<void> | void;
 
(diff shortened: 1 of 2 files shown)


archisman-panigrahi:
I am travelling now but once I'm back I'll test this fix on phosh

Issue #4905, opened 2026-07-02
FR: Make book covers look like actual book covers


dastarruer:
**Does your feature request involve difficulty for you to complete a task? Please describe.**

> No, but I think it'd be a nice feature.

**Describe the solution you'd like**

> <img width="750" height="1334" alt="Image" src="https://github.com/user-attachments/assets/732277dc-37df-4f79-885f-9b3fa99c6614" />
>
> Basically, I'd like the book covers to look less like just icons, but more like this. Personally, I think this makes my collection feel much cozier to just look at, and is the main thing I miss from apple books.

PR #5042, opened 2026-07-09
refactor: rename ColorPanel to ThemePanel

dastarruer:
Since the color panel in the settings view does a lot more than just setting colors (ie. background images, reading ruler, etc.), it makes more sense to label it as a theme panel in accordance with its current use.

It would also make the panel a more appropriate place to put other customization options down the line, such as #4905. In fact, currently, these two setting relating to library view customization:

image

Would benefit from being placed in the theme panel for better discoverability.


dastarruer:
All tests are passing on my end with `pnpm test`, not sure what's happening w the checks.

chrox:
Please read the CI error log. The CI runs much more than pnpm test.

chrox:
Please also make sure `tsgo` is installed:
```
➜  readest-app git:(dev) git push origin dev
Connection to ssh.github.com port 443 [tcp/https] succeeded!
$ pnpm -w format:check
$ biome format .
Checked 1707 files in 1067ms. No fixes applied.
$ tsgo --noEmit && biome lint .
Checked 1674 files in 1384ms. No fixes applied.
$ dotenv -e .env -e .env.test.local -- vitest
```
so that the git hook can lint the repo before you push to the remote repo. And type errors like this can be identified earlier.

dastarruer:
Should I make a separate PR to add tsgo to `package.json`?

chrox:
Not for now. We’ll upgrade to the newly released TypeScript 7 after we’ve completed full testing, and then replace the tsgo command with tsc.

dastarruer:
I just had to run `pnpm test` in `apps/readest-app` instead of the project root. It's fixed now

chrox:
Yes, source of truth and harness is in the CI.

PR #5058, opened 2026-07-10
fix(opds): fix logic for temporary destination filename (#5024)

### LINKED ISSUE #5024: OPDS Auto-download fails (2)
## Bug Description

Same context as #4988.

Things are a bit better in 0.11.18, but this isn't completely solved.

- I have 9 books in my magic shelf.
- The 9 `HEAD` requests still `401` (missing `authorization` header), but I believe this is not a problem.
- All `GET` requests are `200`. So it seems the problem is within the app.

=> Only 4 books land in readest, the other 5 fail. I tried again, and the count was different: respectively 6 and 3.

I can reproduce the bug on desktop (macOS) and mobile (iPadOS).

Image

There might be some concurrency issue here.

Here are the logs, server side:

```
192.168.0.22 - client [08/Jul/2026:20:25:47 +0000] "GET /cwa/opds/magicshelf/5 HTTP/2.0" 200 11966 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:48 +0000] "HEAD /cwa/opds/download/570/epub/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:48 +0000] "HEAD /cwa/opds/download/555/epub/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:48 +0000] "HEAD /cwa/opds/download/572/epub/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:48 +0000] "GET /cwa/opds/download/570/epub/ HTTP/2.0" 200 166192 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:49 +0000] "HEAD /cwa/opds/download/553/epub/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:49 +0000] "GET /cwa/opds/download/555/epub/ HTTP/2.0" 200 757359 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:49 +0000] "HEAD /cwa/opds/download/552/pdf/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:49 +0000] "GET /cwa/opds/download/553/epub/ HTTP/2.0" 200 1039340 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:49 +0000] "GET /cwa/opds/download/572/epub/ HTTP/2.0" 200 7865466 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:49 +0000] "HEAD /cwa/opds/download/551/pdf/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:49 +0000] "HEAD /cwa/opds/download/364/pdf/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:50 +0000] "GET /cwa/opds/download/552/pdf/ HTTP/2.0" 200 7997386 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:50 +0000] "GET /cwa/opds/download/551/pdf/ HTTP/2.0" 200 18717538 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:50 +0000] "HEAD /cwa/opds/download/57/epub/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - - [08/Jul/2026:20:25:50 +0000] "HEAD /cwa/opds/download/4/epub/ HTTP/2.0" 401 0 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:51 +0000] "GET /cwa/opds/download/364/pdf/ HTTP/2.0" 200 26191582 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:51 +0000] "GET /cwa/opds/download/4/epub/ HTTP/2.0" 200 1070135 "-" "Readest/1.0 (OPDS Browser)" "-"
192.168.0.22 - client [08/Jul/2026:20:25:51 +0000] "GET /cwa/opds/download/57/epub/ HTTP/2.0" 200 4055546 "-" "Readest/1.0 (OPDS Browser)" "-"
```

chloeroform:
Previously, the destination filename for OPDS downloads was derived from the request URL. In applications such as Calibre Web Automated, download URLs have the form `/opds/download/364/pdf/`, causing temporary files to be named all `pdf.pdf` or `epub.epub`, leading to race conditions.

This patch uses a random UUID for temporary filenames instead of relying on the URL, ensuring unique filenames regardless of the download endpoint.

Fixes #5024.

### DIFF
--- a/apps/readest-app/src/services/opds/autoDownload.ts
+++ b/apps/readest-app/src/services/opds/autoDownload.ts
@@ -17,6 +17,7 @@ import { upsertOPDSSourceMapping } from './sourceMap';
 import { isRetryEligible, DOWNLOAD_CONCURRENCY, MAX_RETRY_ATTEMPTS } from './types';
 import type { PendingItem, SyncResult, OPDSSubscriptionState, FailedEntry } from './types';
 import { runWithConcurrency } from '@/utils/concurrency';
+import { uniqueId } from '@/utils/misc';
 
 /**
  * Download a single item and import it into the library.
@@ -62,9 +63,7 @@ async function downloadAndImport(
   // Use the last non-empty path segment as the base; falling back to the
   // entry id avoids producing 200+ char filenames from deep URLs and keeps
   // us comfortably under the ~255-byte filesystem limit.
-  const lastSegment = pathname.split('/').filter(Boolean).pop() ?? '';
-  const sanitized = (lastSegment || item.entryId).replaceAll(/[/\\:*?"<>|]/g, '_').slice(0, 200);
-  const basename = sanitized || 'opds-download';
+  const basename = uniqueId();
   const filename = ext ? `${basename}.${ext}` : basename;
   let dstFilePath = await appService.resolveFilePath(filename, 'Cache');
 
--- a/apps/readest-app/src/app/opds/page.tsx
+++ b/apps/readest-app/src/app/opds/page.tsx
@@ -57,6 +57,7 @@ import { normalizeOPDSCustomHeaders } from './utils/customHeaders';
 import { closeOPDSBrowser, stashOPDSReturnTarget } from './utils/opdsClose';
 import { findExistingBookForPublication } from './utils/findExistingBook';
 import Dialog from '@/components/Dialog';
+import { uniqueId } from '@/utils/misc';
 
 type ViewMode = 'feed' | 'publication' | 'search' | 'loading' | 'error';
 
@@ -588,7 +589,7 @@ export default function BrowserPage() {
 
           const pathname = decodeURIComponent(new URL(url).pathname);
           const ext = getFileExtFromMimeType(parsed?.mediaType) || getFileExtFromPath(pathname);
-          const basename = pathname.replaceAll('/', '_');
+          const basename = uniqueId();
           const filename = ext ? `${basename}.${ext}` : basename;
           let dstFilePath = await appService?.resolveFilePath(filename, 'Cache');
           console.log('Downloading to:', url, dstFilePath);


chrox:
We don't need a random part on the file name, maybe it's better to take the last two or three parts of the url pathname as the base filename  and we need to keep the extension name as for some file types this is needed to parse the book metadata when importing.

chloeroform:
The final filename is `newFilePath`, which corresponds to the remote filename. The temporary file is deleted right away.

https://github.com/readest/readest/blob/36ea007e15018e4f334b273d1c321db398792063/apps/readest-app/src/services/opds/autoDownload.ts#L90-L91

`dstFilePath` is only used by `downloadFile()`.

chrox:
Yes, you are right. The race condition might also happen for manual download, could you also fix it?

chloeroform:
Sure! 🙂

chloeroform:
`handleImportBookFromUrl` uses `clip_url`.
`handleImportBooksFromFiles` and `handleImportBooksFromFiles` don't use temporary files. 

I found another call site in `apps/readest-app/src/app/opds/page.tsx`, which I fixed the same way.

We're probably good.

chrox:
Please use uniqueId for the random part of the file name and keep the extension name in case that the OPDS server does not provide the full filename in the response header, which might cause failed import for TXT and PDF files.

apps/readest-app/src/services/opds/autoDownload.ts
```diff
@@ -50,25 +50,8 @@ async function downloadAndImport(
...
 
-  const parsed = parseMediaType(item.mimeType);
-  const rawPathname = new URL(url).pathname;
-  let pathname: string;
-  try {
-    pathname = decodeURIComponent(rawPathname);
-  } catch {
-    pathname = rawPathname;
-  }
-  const ext = getFileExtFromMimeType(parsed?.mediaType) || getFileExtFromPath(pathname);
```
chrox, line 61: I mean keep the original ext in the destination file name.

chrox:
I mean keep this ext in the destination file name.

chrox:
> The extension will be kept in the destination file name (`newFilePath`).

```
const probedFilename = await probeFilename(responseHeaders);
  if (probedFilename) {
    const newFilePath = await appService.resolveFilePath(probedFilename, 'Cache');
    await appService.copyFile(dstFilePath, 'None', newFilePath, 'None');
    await appService.deleteFile(dstFilePath, 'None');
    dstFilePath = newFilePath;
  }

  const book = await appService.importBook(dstFilePath, books);
```

The `newFilePath` requires that responseHeaders contains `content-disposition` response header which is not always the case. If probeFilename returns empty string, importBook will import a file with a name of a random string without extension name, and if will fail for TXT and PDF.

chrox:
> Yes, I understand, but as stated above, you cannot infer the extension from the URL:
> 
> > The [OPDS spec](https://github.com/opds-community/specs/blob/master/opds-2.0.md) doesn't guarantee the extension is in the URL. We cannot therefore, and unfortunately, assume anything from the Calibre endpoint. Other servers could also follow other conventions.
> 
> The URL could be `opds/dc27a635-9086-4326-a02a-c748b29ead69`, or `/opds/book/24`, etc.

No, the code you deleted does try its best to parse the extension from the MIME and the path, and it works in most cases: 

`const ext = getFileExtFromMimeType(parsed?.mediaType) || getFileExtFromPath(pathname);`

chrox (changes requested):

chloeroform:
Sorry, I misunderstood what you meant.

PR #5060, opened 2026-07-11
feat: add nicer issue templates

dastarruer:
Convert existing issue templates to yaml. This makes opening issues a more user-friendly process, allows us to enforce certain fields, and makes sure all issues follow a specific, easy-to-read format. 

~~Stolen~~ Borrowed from [nixpkgs' issue template](https://github.com/NixOS/nixpkgs/blob/master/.github/ISSUE_TEMPLATE/01_bug_report.yml).


dastarruer:
Just a heads up, when previewing these templates on github, the following message appears:

image

Even though `name` is very much unique in both files, so I couldn't tell you what's happening here. And when I push both templates to a separate repo, there are no issues. Github seems to still think that the old `.md` files still exist, and is throwing an error.

chrox (approved):

PR #5071, opened 2026-07-11
feat(settings): Increase margin upper bounds

ghost:
The current maximum margin value is too restrictive on larger mobile devices. For example, on a [Samsung S11 Ultra](https://en.wikipedia.org/wiki/Samsung_Galaxy_Tab_S11#Display)(1848×2960), the existing limit corresponds to approximately 4% of the screen width and 3% of the screen height, resulting in margins that are visually insignificant.

This change increases the maximum configurable margin value. The new upper bound matches [the range](https://github.com/readest/readest/blob/ee727b0cb315b57a587cf0b6bcad331be2b7b6bc/apps/readest-app/src/components/settings/LayoutPanel.tsx#L667) used for "Maximum Column Height" and "Maximum Column Width", providing consistency across related layout settings.

**Potential regression:** none identified, the change won't impact the current installed base, no logic changed.
**Testing performed:** modified margin settings and opened ebooks using Readest [v0.11.18](https://github.com/readest/readest/releases/tag/v0.11.18) (+ patch) on both Android (S11 Ultra) and Linux (ArchLinux).
**Blast radius:** minimal (Settings > Layout, ebook reader).

### DIFF
--- a/apps/readest-app/src/components/settings/LayoutPanel.tsx
+++ b/apps/readest-app/src/components/settings/LayoutPanel.tsx
@@ -611,7 +611,7 @@ const LayoutPanel: React.FC<SettingsPanelPanelProp> = ({ bookKey, onRegisterRese
               ? Math.max(0, Math.round((16 - gridInsets.top) / 4) * 4) - gridInsets.top
               : 0
           }
-          max={88}
+          max={144}
           step={4}
         />
         <NumberInput
@@ -623,7 +623,7 @@ const LayoutPanel: React.FC<SettingsPanelPanelProp> = ({ bookKey, onRegisterRese
               ? Math.max(0, Math.round((16 - gridInsets.bottom) / 4) * 4) - gridInsets.bottom
               : 0
           }
-          max={88}
+          max={144}
           step={4}
         />
         <NumberInput
@@ -631,7 +631,7 @@ const LayoutPanel: React.FC<SettingsPanelPanelProp> = ({ bookKey, onRegisterRese
           value={showFooter && isVertical ? marginLeftPx : compactMarginLeftPx}
           onChange={showFooter && isVertical ? setMarginLeftPx : setCompactMarginLeftPx}
           min={0}
-          max={88}
+          max={144}
           step={4}
         />
         <NumberInput
@@ -639,7 +639,7 @@ const LayoutPanel: React.FC<SettingsPanelPanelProp> = ({ bookKey, onRegisterRese
           value={showHeader && isVertical ? marginRightPx : compactMarginRightPx}
           onChange={showHeader && isVertical ? setMarginRightPx : setCompactMarginRightPx}
           min={0}
-          max={88}
+          max={144}
           step={4}
         />
         <NumberInput


chrox:
A maximum margin value of 9999 is not reasonable and can lead to serious performance issues if users actually set it. We should enforce a practical upper limit, such as 128 (or at most 144), to prevent layout and rendering issues on some devices.

ghost:
That's a fair point of course, but can't we let users find the numbers that work best for them both in terms of performance and comfort?

144px is still less than 5% of 2960, that puts the text very close to the edge.

I think in the end it's very hard to find values that will work for every user, book, device. The user should probably be the final judge.

chrox:
> 144px is still less than 5% of 2960, that puts the text very close to the edge.
> 

The 1848×2960 figure refers to the physical screen resolution. The settings here are specified in CSS logical pixels, which are divided by your device’s pixel ratio (at least 3 in your case). That means a margin of 144 logical pixels corresponds to roughly 15% of the vertical resolution and 23% of the horizontal resolution on your device.

ghost:
Thanks for the explanation.

I still believe it's better to trust the user and let them make that decision.

From my perspective, this doesn't introduce additional complexity to the codebase or to the app usage. It simply gives users a bit more control over their reading environment.

chrox:
The option itself implies that values up to 9999 pixels are supported, encouraging users to select them, but in practice doing so only leads to layout and performance problems. I don’t think allowing a maximum value of 9999 pixels is a reasonable design choice.

chrox:
Please provide a concrete use case to verify that a margin more than 144 px is needed in order to increase the upper limit. Note that the Column Gap is also added to the horizontal margin if the horizontal margins are not enough.

ghost:
> The option itself implies that values up to 9999 pixels are supported, encouraging users to select them, but in practice doing so only leads to layout and performance problems.

The app allows you to inject CSS to customize both book content and the reader interface though...

But anyway, I understand 9999 might not be reasonable. I resubmitted with 960 as ceiling value.

chrox:
I’m sorry that 960 is still not reasonable, it’s a random number without careful design and test. We won’t merge this.

ghost:
Let's not overstate the issue here.

I tested this carefully on two devices using the values *I* wanted to use. I also included a brief analysis at the end of my commit explaining the reasoning behind the change and its impact. Looking through the backlog, I don't think every change has received this same level of scrutiny.

Ultimately, this comes down to a difference in philosophy. I prefer leaving the limit open so users can choose the settings that work best for them, whereas you prefer constraining the range to what you consider a reasonable value. I think both are valid approaches.

My goal is to give people the flexibility to choose the settings they prefer. As I've mentioned before, I don't think there's a single value that will work well across all books, devices, and users. There is no scientific reasoning behind this. If we set 144px, someone still might come tomorrow wanting for more. In the end I'm not looking for values that work, I'm looking for a maximum value that will never be reached.

chrox:
We need a real-world edge case demonstrating that a 144 px margin is genuinely insufficient, just as you previously showed that 88 px wasn’t enough.

Our design philosophy is to keep settings practical and well justified. The goal of this project is not to expose every possible option to users or overwhelm users with unnecessary configuration. That’s one of the biggest lessons I’ve learned from years of developing the KOReader project.

chrox:
Please also keep in mind that there is a Paragraph Mode in case you are not aware of, which will show only one paragraph each time if you want to focus on the current paragraph.

ghost:
I'm not sure how else to explain this: users **don't have to use** the extreme value. They simply choose the value that works best for them within the available range of 0 to 960.

I also feel we're approaching this discussion with very different assumptions. I don't think we're being equally open to considering each other's perspective, which makes it difficult to have a productive conversation. You should be more open to your users/contributors as a maintainer.

To be honest, I've found this discussion quite frustrating.

Have a good one.

chrox:
The key difference in our thinking is that I believe every option we expose should be safe to use. A maximum value of 960 px isn’t just an option that most people won’t choose, it can create real problems if someone does.

Because this is a global setting, an extreme value can affect every EPUB on the device. In the worst case, the paginator may have to lay out each chapter as thousands of pages, leading to excessive memory usage, poor performance, or even making books effectively impossible to open. That would be far more frustrating for users than limiting the setting to a practical range.

I’m not opposed to increasing the limit if we can justify it with a real use case. If someone can demonstrate that 144 px is genuinely insufficient, I’m happy to revisit the upper bound. But I’d prefer not to expose values that we already know can break the reading experience for no practical benefit.

I appreciate the discussion, even if we don’t end up agreeing on this particular point.

chrox:
The reason I insist on a much lower upper bound for the margin value is that I’ve already run into this pitfall in practice and settled on what I believe is a safer, more reasonable limit. You can also view it as a limitation of the current paginator: its performance degrades rapidly as the margins become larger. This is largely due to how CSS column layout works, and it’s not something that would be obvious to a new contributor.

The Readest community is always open to contributions, and I hope this disagreement doesn’t discourage you from contributing in the future.

chrox:
LGTM now, thanks.

PR #5079, opened 2026-07-13
feat(sorting): add toggle to filter by time remaining

### LINKED ISSUE #5056: FR: Sort by page count
I'm switching up from Librera where a similar feature is present. I tend to travel a lot and this feature helped me pick a book from my library that I could read in the time I was on the road.

The way that this feature works in Librera is that when you open a EPUB file it calculates the amount of page turns it would take for you to reach the end of a given file, and allows you to sort your personal library in the shortest to longest file you have. When you know you reading speed this could facilitate picking which book to start in the time you will have to finish it.

Currently, I see have my library both in the Librera and Readest apps and whenever I want to start a new read I sort it through Librera and move to find the book in Readest to read.

dastarruer:
Closes #5056.

I also plan on adding a setting to show time left in the `ReadingProgress` component to complement this feature. Also going to add tests.

I think I could probably also remove the `Progress` sort type in favor of this, since this is a better version of it anyways.

### DIFF
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -2040,5 +2040,9 @@
   "Readest only accesses the files it creates in your OneDrive.": "لا يصل Readest إلا إلى الملفات التي ينشئها في OneDrive الخاص بك.",
   ". Make OneDrive the active cloud provider.": ". اجعل OneDrive مزود السحابة النشط.",
   "Sentence Pause": "وقفة بين الجمل",
-  "Paragraph Gap": "وقفة بين الفقرات"
+  "Paragraph Gap": "وقفة بين الفقرات",
+  "{{minutes}}m left": "بقيت {{minutes}} دقيقة",
+  "{{hours}}h left": "بقيت {{hours}} ساعة",
+  "Time Remaining": "الوقت المتبقي",
+  "Theme": "السمة"
 }
--- a/apps/readest-app/src/app/library/components/ReadingProgress.tsx
+++ b/apps/readest-app/src/app/library/components/ReadingProgress.tsx
@@ -8,6 +8,7 @@ import { getTimeRemainingMinutes } from '../utils/libraryUtils';
 
 interface ReadingProgressProps {
   book: Book;
+  showTimeRemaining: boolean;
 }
 
 const getProgressPercentage = (book: Book) => {
@@ -22,15 +23,15 @@ const getProgressPercentage = (book: Book) => {
 };
 
 const ReadingProgress: React.FC<ReadingProgressProps> = memo(
-  ({ book }) => {
+  ({ book, showTimeRemaining }) => {
     const _ = useTranslation();
     const progressPercentage = useMemo(() => getProgressPercentage(book), [book]);
 
     const minutes = getTimeRemainingMinutes(book);
     const hours = minutes ? Math.floor(minutes / 60) : undefined;
     const mins = minutes ? minutes % 60 : undefined;
-    const timeRemainingLabel =
-      hours && mins
+    const progressLabel =
+      showTimeRemaining && hours && mins
         ? _(`${progressPercentage}% · ${hours}h ${mins}m left`)
         : `${progressPercentage}%`;
 
@@ -78,7 +79,7 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
         role='status'
         aria-label={`${progressPercentage}%`}
       >
-        <span>{timeRemainingLabel}</span>
+        <span>{progressLabel}</span>
       </div>
     );
   },
--- a/apps/readest-app/src/app/library/components/ReadingProgress.tsx
+++ b/apps/readest-app/src/app/library/components/ReadingProgress.tsx
@@ -4,7 +4,7 @@ import type { Book } from '@/types/book';
 import { useTranslation } from '@/hooks/useTranslation';
 import { SHOW_UNREAD_STATUS_BADGE } from '@/services/constants';
 import StatusBadge from './StatusBadge';
-import { getTimeRemainingMinutes } from '../utils/libraryUtils';
+import { getDisplayedTimeRemaining } from '../utils/libraryUtils';
 
 interface ReadingProgressProps {
   book: Book;
@@ -27,12 +27,18 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
     const _ = useTranslation();
     const progressPercentage = useMemo(() => getProgressPercentage(book), [book]);
 
-    const minutes = getTimeRemainingMinutes(book);
-    const hours = minutes ? Math.floor(minutes / 60) : undefined;
-    const mins = minutes ? minutes % 60 : undefined;
+    const minutes = getDisplayedTimeRemaining(book);
+    const formatTimeLeft = (total: number) => {
+      if (total < 60) return _('{{minutes}}m left', { minutes: total });
+      const hours = total / 60;
+      // One decimal below 10h (1.6h), whole hours above it (23h) — a tenth of an
+      // hour stops being meaningful once there are dozens of them left.
+      const rounded = hours < 10 ? Math.round(hours * 10) / 10 : Math.round(hours);
+      return _('{{hours}}h left', { hours: rounded });
+    };
     const progressLabel =
-      showTimeRemaining && hours && mins
-        ? _(`${progressPercentage}% · ${hours}h ${mins}m left`)
+      showTimeRemaining && minutes
+        ? `${progressPercentage}% · ${formatTimeLeft(minutes)}`
         : `${progressPercentage}%`;
 
     if (book.readingStatus === 'finished') {
@@ -75,11 +81,11 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
 
     return (
       <div
-        className='text-neutral-content/70 flex justify-between text-xs'
+        className='text-neutral-content/70 flex min-w-0 justify-between text-xs'
         role='status'
         aria-label={`${progressPercentage}%`}
       >
-        <span>{progressLabel}</span>
+        <span className='truncate'>{progressLabel}</span>
       </div>
     );
   },
@@ -87,7 +93,8 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
     return (
       prevProps.book.hash === nextProps.book.hash &&
       prevProps.book.updatedAt === nextProps.book.updatedAt &&
-      prevProps.book.readingStatus === nextProps.book.readingStatus
+      prevProps.book.readingStatus === nextProps.book.readingStatus &&
+      prevProps.showTimeRemaining === nextProps.showTimeRemaining
     );
   },
 );
--- a/apps/readest-app/src/app/library/components/ReadingProgress.tsx
+++ b/apps/readest-app/src/app/library/components/ReadingProgress.tsx
@@ -25,19 +25,11 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
     const _ = useTranslation();
     const progressPercentage = useMemo(() => getProgressPercentage(book), [book]);
 
-    let timeRemainingHours;
-    let timeRemainingMinutes;
-    if (book.timeRemainingMinutes) {
-      timeRemainingHours = Math.floor(book.timeRemainingMinutes / 60);
-      timeRemainingMinutes = book.timeRemainingMinutes % 60;
-    }
-
-    let timeRemainingLabel;
-    if (timeRemainingHours && timeRemainingMinutes) {
-      timeRemainingLabel = `${progressPercentage}% · ${timeRemainingHours}h ${timeRemainingMinutes}m left`;
-    } else {
-      timeRemainingLabel = `${progressPercentage}%`;
-    }
+    const minutes = book.timeRemainingMinutes;
+    const hours = minutes ? Math.floor(minutes / 60) : undefined;
+    const mins = minutes ? minutes % 60 : undefined;
+    const timeRemainingLabel =
+      hours && mins ? `${progressPercentage}% · ${hours}h ${mins}m left` : `${progressPercentage}%`;
 
     if (book.readingStatus === 'finished') {
       return (
--- a/apps/readest-app/src/app/library/components/ReadingProgress.tsx
+++ b/apps/readest-app/src/app/library/components/ReadingProgress.tsx
@@ -25,6 +25,20 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
     const _ = useTranslation();
     const progressPercentage = useMemo(() => getProgressPercentage(book), [book]);
 
+    let timeRemainingHours;
+    let timeRemainingMinutes;
+    if (book.timeRemainingMinutes) {
+      timeRemainingHours = Math.floor(book.timeRemainingMinutes / 60);
+      timeRemainingMinutes = book.timeRemainingMinutes % 60;
+    }
+
+    let timeRemainingLabel;
+    if (timeRemainingHours && timeRemainingMinutes) {
+      timeRemainingLabel = `${progressPercentage}% · ${timeRemainingHours}h ${timeRemainingMinutes}m left`;
+    } else {
+      timeRemainingLabel = `${progressPercentage}%`;
+    }
+
     if (book.readingStatus === 'finished') {
       return (
         <div className='flex justify-start'>
@@ -69,7 +83,7 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
         role='status'
         aria-label={`${progressPercentage}%`}
       >
-        <span>{progressPercentage}%</span>
+        <span>{timeRemainingLabel}</span>
       </div>
     );
   },
--- a/apps/readest-app/src/app/library/components/ReadingProgress.tsx
+++ b/apps/readest-app/src/app/library/components/ReadingProgress.tsx
@@ -4,6 +4,7 @@ import type { Book } from '@/types/book';
 import { useTranslation } from '@/hooks/useTranslation';
 import { SHOW_UNREAD_STATUS_BADGE } from '@/services/constants';
 import StatusBadge from './StatusBadge';
+import { getTimeRemainingMinutes } from '../utils/libraryUtils';
 
 interface ReadingProgressProps {
   book: Book;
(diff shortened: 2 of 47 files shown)


dastarruer:
Update: This is what I have so far:

image

Should I add a setting to toggle this time remaining section or can it be default behavior?

dastarruer:
Tests will fail, but I'll get to it when I get to it.

chrox:
Please make sure the `.husky/pre-push` hook is actually working on your side. Most type errors in your PR should be caught with the lint and test phase before each commit push.

dastarruer:
> Please make sure the `.husky/pre-push` hook is actually working on your side. Most type errors in your PR should be caught with the lint and test phase before each commit push.

The pre-push hook is expensive to run on my machine, so I'm not regularly running it. But the only thing getting caught is the failing tests. There aren't any type errors getting thrown.

dastarruer:
Also lmk if I need to add any more tests.

apps/readest-app/src/app/library/components/ReadingProgress.tsx
```diff
@@ -25,6 +25,20 @@ const ReadingProgress: React.FC<ReadingProgressProps> = memo(
     const _ = useTranslation();
     const progressPercentage = useMemo(() => getProgressPercentage(book), [book]);
 
+    let timeRemainingHours;
```
dastarruer: I don't know what the best way to write this kind of logic in typescript would be, so please let me know if this should be changed.

apps/readest-app/src/types/book.ts
```diff
@@ -119,6 +119,7 @@ export interface Book {
 
   lastUpdated?: number; // deprecated in favor of updatedAt
   progress?: [number, number]; // Add progress field: [current, total], 1-based page number
+  timeRemainingMinutes?: number; // Time left until the end of the book
```
chrox: There is no need for this field. `timeRemainingMinutes` can be calculated from the `progress?: [number, number]` like this:

```
const timeLeftStr = showPagesLeft
    ? remainingInBook
      ? _('{{time}} min left in book', {
          time: formatNumber(
            Math.round((pagesLeft * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT),
            localize,
            lang,
          ),
        })
      : _('{{time}} min left in chapter', {
          time: formatNumber(
            Math.round((pagesLeft * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT),
            localize,
            lang,
          ),
        })
    : '';
```

see: `apps/readest-app/src/app/reader/components/ProgressBar.tsx#L96`

chrox (changes requested):

apps/readest-app/src/app/library/components/ReadingProgress.tsx
chrox: Should be:
```
const minutes = book.timeRemainingMinutes;
const hours = minutes ? Math.floor(minutes / 60) : 0;
const mins = minutes ? minutes % 60 : 0;
const timeRemainingLabel = hours && mins
    ? `${progressPercentage}% · ${hours}h ${mins}m left`
    : `${progressPercentage}%`;
```

chrox: But any user facing strings need to be i18n translated in this project. 

see: https://github.com/readest/readest/blob/main/apps/readest-app/docs/i18n.md

apps/readest-app/src/types/book.ts
dastarruer: Should I move this into a function in `apps/readest-app/src/types/book.ts` taking `pagesLeft` as an arg?

chrox:
> The pre-push hook is expensive to run on my machine, so I'm not regularly running it. But the only thing getting caught is the failing tests. There aren't any type errors getting thrown.

```
$ tsgo --noEmit && biome lint .
Error: src/__tests__/store/library-store.test.ts(98,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(108,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(116,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(125,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(135,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(146,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(161,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(169,34): error TS2554: Expected 4 arguments, but got 3.
Error: src/__tests__/store/library-store.test.ts(181,34): error TS2554: Expected 4 arguments, but got 3.
[ELIFECYCLE] Command failed with exit code 2.
Error: Process completed with exit code 2.
```

This should be caught earlier.

apps/readest-app/src/types/book.ts
chrox: No, `apps/readest-app/src/types` are pure type interfaces. Since `time = Math.round(page * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT` is simple and straightforward enough, you can just inline it.

chrox:
Also, I would suggest moving the remaining time of the book into the book details to keep the bookshelf clean and concise.

dastarruer:
> Also, I would suggest moving the remaining time of the book into the book details to keep the bookshelf clean and concise.

So move `time = Math.round(page * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT` there instead of recalculating multiple times, right? In that case, should I also remove `time = Math.round(page * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT` from `ProgressBar.tsx`?

dastarruer:
> > The pre-push hook is expensive to run on my machine, so I'm not regularly running it. But the only thing getting caught is the failing tests. There aren't any type errors getting thrown.
> 
> ```
> $ tsgo --noEmit && biome lint .
> Error: src/__tests__/store/library-store.test.ts(98,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(108,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(116,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(125,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(135,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(146,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(161,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(169,34): error TS2554: Expected 4 arguments, but got 3.
> Error: src/__tests__/store/library-store.test.ts(181,34): error TS2554: Expected 4 arguments, but got 3.
> [ELIFECYCLE] Command failed with exit code 2.
> Error: Process completed with exit code 2.
> ```
> 
> This should be caught earlier.

Likely because I passed `--no-verify` to skip the pre-push hook. I'll stop doing that, since it's just causing me issues.

apps/readest-app/public/locales/ar/translation.json
```diff
@@ -2040,5 +2040,7 @@
   "Readest only accesses the files it creates in your OneDrive.": "لا يصل Readest إلا إلى الملفات التي ينشئها في OneDrive الخاص بك.",
   ". Make OneDrive the active cloud provider.": ". اجعل OneDrive مزود السحابة النشط.",
   "Sentence Pause": "وقفة بين الجمل",
-  "Paragraph Gap": "وقفة بين الفقرات"
+  "Paragraph Gap": "وقفة بين الفقرات",
+  "Time Remaining": "__STRING_NOT_TRANSLATED__",
```
chrox: You cannot leave untranslated keys in the translation files. This will make the English fallback fail when the strings are not translated timely. And the CI will fail this. Please just don't extract the keys if you don't translate them.

chrox (changes requested):

apps/readest-app/src/types/book.ts
```diff
@@ -487,6 +488,7 @@ export interface BookConfig {
   bookHash?: string;
   metaHash?: string;
   progress?: [number, number]; // [current pagenum, total pagenum], 1-based page number
+  timeRemainingMinutes?: number; // Time left until the end of the book
```
chrox: We need to remove this `timeRemainingMinutes` in favor of the `progress` as the source of truth.

chrox (changes requested):

chrox:
> So move `time = Math.round(page * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT` there instead of recalculating multiple times, right? In that case, should I also remove `time = Math.round(page * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT` from `ProgressBar.tsx`?

I meant showing the remaining reading time only in the book details dialog, alongside the book’s metadata, rather than displaying it for every book on the bookshelf, which would make the bookshelf feel cluttered. I wasn’t referring to where the value is calculated. I also think it’s premature optimization to worry about caching here, since the calculation is trivial.

chrox:
Or show the remaining time on the book items in the bookshelf only when sorting with the remaining time.

dastarruer:
> I meant showing the remaining reading time only in the book details dialog, alongside the book’s metadata, rather than displaying it for every book on the bookshelf, which would make the bookshelf feel cluttered. 

Would like to push back on that. This is only shown for books with progress made, not for finished or unread books:

image

Personally, I only have at most a few books that I'm currently reading. Hiding remaining reading time in the book details dialog would defeat the whole point of being to, at a glance, see how much time is left in the book (which fits the use case detailed in #5056). Of course, it's up to you in the end, but I don't think it's such a good idea.

Also, (even though I really don't feel like having to add this...) I can add this as a toggle to the Theme panel in settings.

chrox:
> I only have at most a few books that I'm currently reading. Hiding remaining reading time in the book details dialog would defeat the whole point of being to, at a glance, see how much time is left in the book (which fits the use case detailed in #5056).

It's reasonable, but please gate it behind the 'Remaining Time' sorting option. Users will be surprised to see a full screen of remaining hours which will defeat the look and feel of this product. Yes, there is no need to add yet another option to just show/dismiss this info.

dastarruer:
> Or show the remaining time on the book items in the bookshelf only when sorting with the remaining time.

Then users would be confused as to how to show time remaining. The only other solution I can think of is adding a toggle to the View Menu. 

Edit: maybe this could also auto-toggle on if sort by time remaining is also toggled?

chrox:
> Then users would be confused as to how to show time remaining.

No, they won't, if the remaining time is not shown by default. People won't get confused with things they are not aware of.

dastarruer:
Also, you removed the comment about how `TimeRemaining` and `Progress` are too similar. However, I agree, and I propose that we remove `Progress`, simply because `TimeRemaining` is a better indicator of how far a user from finishing a book.

chrox:
> Also, you removed the comment about how `TimeRemaining` and `Progress` are too similar. However, I agree, and I propose that we remove `Progress`, simply because `TimeRemaining` is a better indicator of how far a user from finishing a book.

I removed the comment because I realize they are actually different use cases, and they could both be in the sorting options. But it's definitely not necessary to save the `timeRemainingMinutes` in the book metadata, because it can be calculated from the `progress` which has the total pages count.

dastarruer:
> No, they won't, if the remaining time is not shown by default. People won't get confused with things they are not aware of.

While that's definitely correct, I think in general it's very difficult to discover all the features that Readest has to offer as a first-time user. New users should be able to easily find useful features such as these even without being aware of them beforehand, and so perhaps putting a toggle in the View Menu could help.

Also, me personally, while I probably won't use the sort by time remaining feature, I definitely want to see the time remaining. But, that's just my perspective. If you'd prefer it to be gated only by the sort option, then I'll implement that.

chrox:
> While that's definitely correct, I think in general it's very difficult to discover all the features that Readest has to offer as a first-time user. New users should be able to easily find useful features such as these even without being aware of them beforehand, and so perhaps putting a toggle in the View Menu could help.

Readest intentionally doesn’t expose every option and feature all at once. The goal is that a new user can simply open a book and enjoy a pleasant reading experience without needing to tweak any settings. And advanced users can still find the settings to tweak the reader to their preference.

Because of that, we’re very careful about the default UI and try to avoid adding controls or information that could distract from the reading experience. This is one of the key lessons we learned from KOReader, which is often regarded as overwhelming for new users because of the number of options available.

dastarruer:
> Readest intentionally doesn’t expose every option and feature all at once. The goal is that a new user can simply open a book and enjoy a pleasant reading experience without needing to tweak any settings. And advanced users can still find the settings to tweak the reader to their preference.
> 
> Because of that, we’re very careful about the default UI and try to avoid adding controls or information that could distract from the reading experience. This is one of the key lessons we learned from KOReader, which is often regarded as overwhelming for new users because of the number of options available.

Got it, thanks for explaining. I'll implement the rest of the requested changes.

dastarruer:
Getting this failing test that has nothing to do with my changes:

```
 FAIL  src/__tests__/app/reader/annotator/DictionarySheet.test.tsx > DictionarySheet — expand / collapse > toggles aria-expanded when a card is tapped
AssertionError: expected 'true' to be 'false' // Object.is equality

Expected: "false"
Received: "true"

 ❯ src/__tests__/app/reader/annotator/DictionarySheet.test.tsx:391:48
    389|
    390|     fireEvent.click(card);
    391|     expect(card.getAttribute('aria-expanded')).toBe('false');
       |                                                ^
    392|
    393|     fireEvent.click(card);
```

Likely just local to my machine.

apps/readest-app/src/store/readerStore.ts
```diff
@@ -434,6 +436,7 @@ export const useReaderStore = create<ReaderStore>((set, get) => ({
               config: {
                 ...existing.config,
                 progress,
+                timeRemainingMinutes,
```
chrox: Please remove this.

chrox (changes requested):

chrox:
Thanks for this! Nice feature. I reviewed it, ran it against my library (~688 books), and pushed a commit to your branch with fixes. Details below so nothing is a surprise.

### Sorting was producing a scrambled order

`compareBookByKey` did `aTime - bTime` where both sides are `Infinity` for books with no remaining time. `Infinity - Infinity` is `NaN`, and a comparator that returns `NaN` is inconsistent, so `Array.sort` scattered finished/unread books through the shelf instead of sinking them. That is why finished books and groups showed up *first* in my library rather than last.

Items that cannot show a remaining time now always sort last. This can't live inside the comparator: every comparison is multiplied by the sort-order multiplier, so an `Infinity` sentinel (even without the `NaN`) would flip those items to the *top* in descending order. It is applied as a bucket outside the multiplier instead (`withTimeRemainingLast`).

One subtlety worth calling out: **"has pages left" is not the same as "can show a time."** A book marked *finished* often still has pages left, so it has a remaining time, but `ReadingProgress` early-returns a status badge and never renders one. My first attempt keyed off `getTimeRemainingMinutes` and the finished books were still stuck at the top. Sorting and the label now share `getDisplayedTimeRemaining` so they can't drift apart. Groups sink too, since a group tile renders no progress at all.

### The label never showed for short books

The gate was `showTimeRemaining && hours && mins`. `hours` is `0` under 60 minutes and `mins` is `0` on an exact hour, so both are falsy and it silently fell back to a bare percentage. The effect: sorting ascending puts the shortest books first, and the entire first screen showed no time at all, i.e. exactly the books the feature exists for.

The format is now a single unit: minutes below an hour, hours with one decimal above it, whole hours past 10h (`8m left` / `1.6h left` / `24h left`). That also replaces values like `208h 37m left`, which is not something a reader can act on.

### The label wrapped and overflowed the tile

At the default 6 columns, `43% · 1h 38m left` wrapped to two lines and spilled into the row below. It's now clamped to one line (`truncate` + `min-w-0` on the flex item, which is required or the child won't shrink below its content width, plus `shrink-0` on the sibling actions). The compact format above means nothing actually gets ellipsized.

### Stale labels / the VirtuosoGrid `key`

`ReadingProgress` is `memo`'d with a custom comparator that only checks `hash`, `updatedAt` and `readingStatus`. It never learned about the new `showTimeRemaining` prop, so already-mounted items skipped the re-render and kept stale time labels when the sort changed.

`key={showTimeRemaining ? 'time' : 'not-time'}` was masking that by throwing the whole grid away and rebuilding it. That has a real cost (scroll position jumps back to the top; I measured 2200px to 0) and it only covered grid mode: list view has no key, so it kept showing stale labels. Adding the prop to the comparator fixes the cause, and the `key` is gone.

### i18n

`_(\`${progressPercentage}% · ${hours}h ${mins}m left\`)` builds the key at runtime. `i18next-scanner` can't extract it, and every value combination would be its own key, so it could never actually be translated. It's now interpolated keys (`{{minutes}}m left`, `{{hours}}h left`), and those plus `Time Remaining` are translated across all 33 locales (the PR only had `es`).

The extractor also surfaced `Theme`, an unrelated key that was missing from every locale on `main` already. I filled it in rather than leave a literal `__STRING_NOT_TRANSLATED__` in the settings UI, so ignore that part of the diff.

### Tests

Added regression tests for the bucket rule, including the descending case, which is the one the `Infinity` approach silently got wrong. Full suite passes (546 files / 7329 tests), along with biome and tsgo.

Happy to split any of this out if you'd rather land it separately.

chrox (approved):

dastarruer:
Thanks, that all looks great! I missed most of that, so thanks again for fixing it.

Though I’m not so sure about displaying time remaining as a decimal. I’d prefer being able to see minutes and hours left tbh. Perhaps for hour values under 2 digits, minutes can be displayed, and anything >= 10, minutes can be discarded (basically what you’re already doing). I’m saying this because 1.6 hours doesn’t read intuitively as 1 hour and 36 minutes.

dastarruer:
Dang as I was typing this got merged

If you agree with me on that last point can we revert this? @chrox

chrox:
> Though I’m not so sure about displaying time remaining as a decimal. I’d prefer being able to see minutes and hours left tbh. Perhaps for hour values under 2 digits, minutes can be displayed, and anything >= 10, minutes can be discarded (basically what you’re already doing). I’m saying this because 1.6 hours doesn’t read intuitively as 1 hour and 36 minutes.

The translated strings of '1 hour and 36 minutes' is too long in many of the languages Readest supports to fit on a single line in the grid tile. It would end up being truncated, which is arguably worse than displaying the time as a decimal.

PR #5085, opened 2026-07-13
feat(ios): CarPlay support and native TTS playout with Now Playing integration

chrox:
## Summary

Three related pieces of TTS media work, squashed from the development history:

### 1. CarPlay support for TTS (`feat(ios)`)
- `CarPlaySceneDelegate` presents a `CPListTemplate` root with the currently reading book and pushes the system Now Playing template; transport and metadata ride the existing `MPNowPlayingInfoCenter`/`MPRemoteCommandCenter` path.
- New iOS-only `update_carplay_state` command for the media bridge to report session start/stop; TTS autoplay deep links are consumed even when the target book is already mounted (cold restore into the reader).
- Registering the `UIApplicationSceneManifest` crashed the app at launch on real devices due to a use-after-free in tao's `configuration_for_connecting_scene_session` (returned a raw borrow of a `Retained` it dropped; UIKit then retained freed memory). This adds the tao fork as the `packages/tao` submodule (`readest/tao`, branch `readest-0.35`: tao 0.35.3 plus the callback returning an autoreleased strong reference), and bumps `@tauri-apps/api` to 2.11.1 / `@tauri-apps/cli` to 2.11.4 for the tao 0.35/wry 0.55 stack.
- The `packages/tauri` submodule advances to the fork branch `readest-2.11` (fork patches rebased onto upstream 2.11 plus a guard in the generated unlisten script against already-removed listeners).

### 2. Android audio focus (`fix(android)`)
- Modern `AudioFocusRequest` with `USAGE_MEDIA`/`CONTENT_TYPE_SPEECH` and pause-instead-of-duck, matching how spoken audio should yield to navigation prompts.
- Resume after transient loss only when the loss is what paused us; focus transitions forwarded to the web side as media-session play/pause; pause on `ACTION_AUDIO_BECOMING_NOISY`.

### 3. Native audio playout for Edge TTS on iOS (`feat(ios)`)
- WebAudio in WKWebView renders in WebKit's GPU process under an audio session the app cannot own, which made the lock-screen card, pause retention, AirPods routing, and mute-switch behavior unreliable on device.
- Edge TTS utterances now play through an in-process AVPlayer via a deliberately dumb playout engine in the native-tts plugin (`playout_enqueue`/`playout_control`/`playout_position`); the new `NativeAudioPlayer` mirrors the `WebAudioPlayer` session interface so `EdgeTTSClient` just picks the player per platform.
- Non-mixable `.playback` + `.spokenAudio` session with interruption/route-change handling; `MPNowPlayingSession` bound to the playout player reports paused state explicitly so the Now Playing slot survives a pause; word highlight reads the player's media clock; pitch-preserving live rate via the timeDomain algorithm; native inter-sentence gap timer immune to background JS throttling.
- Volume-key page flip no longer clobbers the session: the two plugins coordinate through in-process claim/release notifications, so volume keys keep flipping pages while TTS is paused without dropping the card.

## Test plan
- [x] Unit suite: 7286 tests passing; lint (Biome + tsgo), `fmt:check`, `clippy:check` clean
- [x] Device (iOS 18.7): Now Playing card with artwork on lock screen and control center; play/pause/skip from card and AirPods; card and controls survive pause; mute-switch immunity; background long-play; word highlight sync; live rate change; volume-key page flip before TTS, while paused, and after stop; interrupted music app receives its resume signal after TTS stops
- [x] Device: CarPlay Now Playing template and transport (earlier round on this branch history)
- [x] Android: `compileDebugKotlin` clean; Auto media controls verified in the prior Android Auto work

## Notes for reviewers
- `packages/tauri` submodule pointer requires the `readest-2.11` fork branch (pushed).
- `packages/tao` submodule pointer requires the `readest-0.35` fork branch (pushed); the one-line UAF fix is worth an upstream PR separately.


ElBalsero:
Wow!
May I just add a little comment here to remark how amazing you guys are?

PR #5088, opened 2026-07-13
Fix blank screen issue on Linux with Nvidia drivers

Aperture32GLaDOS:
A [known issue with webkit](https://gitlab.archlinux.org/archlinux/packaging/packages/webkit2gtk/-/work_items/1) on Linux with Nvidia causes a blank screen to show instead of the desired app. A simple fix is to set the WEBKIT_DISABLE_DMABUF_RENDERER environment variable to 1, which is done here automatically via the std::env methods.


chrox:
The best practice is letting the actual users who actually encounter the issue to set this `WEBKIT_DISABLE_DMABUF_RENDERER` in a wrapper script of the readest launcher to workaround the issue. It should not be a default from the upstream.

Issue #5065, opened 2026-07-11
iOS: pulled progress from custom KOReader sync server is never applied (fresh book, File Content checksum)


KhaosNL:
### Bug Description

Readest on iOS successfully connects to a self-hosted KOReader sync server, pushes its own progress, and retrieves the remote progress when a book is opened — but it never applies the retrieved position. The book always opens at the local position, with no prompt and no jump, regardless of the sync strategy setting.

Server logs confirm Readest performs the GET and the server returns the newer position from the other device (KOReader on Kindle) with HTTP 200. Readest then ignores it. In one session Readest subsequently PUT its own older position back to the server, overwriting the newer one.

Sync in the other direction works perfectly: KOReader on Kindle retrieves the position pushed by Readest and shows the "open at last position?" prompt as expected. So server, credentials, and document matching are all confirmed working.

### Steps to Reproduce

1. Self-hosted sync server: ghcr.io/nperez0111/koreader-sync (latest), behind Caddy reverse proxy with valid TLS.
2. Readest iOS: Settings → Integrations → KOReader Sync → connected ("Sync server connected"), checksum method = **File Content**, sync strategy = **Always keep latest** (also tested "Prompt on conflict" — same result).
3. KOReader 2026.03 on Kindle: same server, same account, document matching method = **Binary**.
4. Both devices download the **same EPUB** from the same Calibre-Web OPDS catalog.
5. Tested with a **completely fresh book** never opened in Readest before, to rule out stale local sync data:
   - Open the book on Kindle first, read a few pages, "Push progress from this device now".
   - Force-quit Readest, reopen, open the book.
   - Server log shows Readest's GET and the server responding 200 with the Kindle position.
   - Readest opens the book at its own local position. No prompt, no jump.

The document hashes match on both sides (confirmed: KOReader retrieves and offers Readest's position for the same hash).

### Expected Behavior

After retrieving a newer remote position, Readest should either jump to it ("Always keep latest") or prompt ("Prompt on conflict").

## Server logs

Domain redacted to `kosync.example.com`, username redacted to `<user>`. Timestamps are epoch ms.

**Test 1 — fresh book (hash `f248ce0f15105ff390e5292085e0622b`), never opened in Readest before:**

```
{"level":30,"time":1783772979275,"req":{"method":"GET","url":"http://kosync.example.com/syncs/progress/f248ce0f15105ff390e5292085e0622b"},"msg":"Incoming request"}
[...]

PR #5111, opened 2026-07-14
fix(kosync): resolve element-offset XPointers and isolate percentage drift anchor to KOReader

### LINKED ISSUE #4444: Progress synced from Kavita is not accurate
## Bug Description

Reading sync location down from Kavita incorrect but sends correct location.

Download book direct to android device location storage from machine running Kavita.
Import into Readest that has sync setup to Kavita.
Open book and go to page.
exit book
open book and get prompted for sync location conflict that is at least a few percent too far forward
exit the applciation and open book in Kavita in browser, and see location there is correct

## Expected behavior

Sync location match on read back

## Technical inputs

Operating System:

Android 16 Galaxy Samsung S25
Version 0.11.2
Kavita installed on Fedora 44 on version 0.9.0.2

adagues:
## Summary

Closes #4444.

Fixes a regression where KOReader/KOSync progress pulls could resolve to the wrong chapter, building on top of #5065 and the earlier DocFragment→spine drift fix.

Two distinct bugs were found and fixed:

**Bug 1 — XPointer parsing gap.** `XCFI.parseXPointer()` only recognized offset suffixes in the form `/text().N` or `/text()[K].N`. Real-world CREngine-produced XPointers can glue the numeric offset directly onto the last element segment with no `text()` step at all (e.g. `/body/DocFragment[16]/body/section/div[1].0`), which threw and caused the pull to be classified as `'unresolved'`. Added a new branch to `parseXPointer` to recognize this format; existing formats are unchanged (purely additive).

**Bug 2 — cross-server percentage contamination.** When resolution fails, `resolveRemoteLocalFraction` falls back to using the remote's raw `percentage` as a drift-anchor input to `resolveSpineSectionIndex`. Some KOSync-compatible servers (Kavita, Komga, Stump, calibre-web) emit CREngine-shaped XPointers but compute `percentage` from their own pagination, which is not comparable to foliate-js's byte-size-based spine fraction table. Feeding a non-CREngine percentage into the drift-anchor logic can shift the resolved chapter by several spine items even though the XPointer's nominal section is correct.

Added `isReportedByKOReader(remote)` (`kosyncProgress.ts`) — a heuristic based on the `device` field already present in KOSync responses — and used it to gate whether `percentage` is forwarded as the drift-anchor argument in both call sites that use it (`resolveRemoteLocalFraction` and `useKOSync.applyRemoteProgress`). Absence of a `device` string (the common real-KOReader case) is treated as trust; only a positively-identified look-alike disables the anchor.

## Why no regression in other sync paths

`useProgressSync.ts` (Readest cloud progress sync) and `useNotesSync.ts` (annotations sync) both call `getCFIFromXPointer` without the `percentage` argument, so `resolveSpineSectionIndex` already falls back to the clamped nominal index there — neither fix touches those paths. The XPointer parsing fix is purely additive (new branch, existing formats untouched).

## Testing

- Added regression tests:
  - `xcfi.spec.ts`: 3 new tests covering the `tag[idx].offset` (no `text()` step) XPointer format, including the exact real-world XPointer from the bug report.
  - `kosyncConflict.test.ts`: 6 new tests covering `isReportedByKOReader` and the conditional forwarding of `percentage` to `getCFIFromXPointer`.
- Targeted suite (`xcfi.spec.ts` + `xcfi.kosync-section-offset.test.ts` + `kosyncConflict.test.ts` + `useKOSync.test.tsx`): 72/72 passing.
- Manually verified against the real-world EPUB from the bug report ("Myron Bolitar T12"): before the fix, pulling a Kavita-reported KOSync position landed on the wrong chapter with a `Failed to convert XPointer` / `Failed to resolve remote progress` console error; after the fix, the correct chapter loads with no errors, confirmed across multiple reloads on web (`pnpm dev-web`) and on a physical iOS device (`tauri ios dev`).

## Notes

- `isReportedByKOReader` relies on the `device` string as the only available signal and is acknowledged as an imperfect heuristic — it trusts the common case (no device string) and only excludes positively-identified look-alikes.

### DIFF
--- a/apps/readest-app/.gitignore
+++ b/apps/readest-app/.gitignore
@@ -59,6 +59,9 @@ next-env.d.ts
 #generated
 src-tauri/gen
 
+# local dev overrides (personal signing, etc.)
+src-tauri/tauri.*.local.json
+
 # vendor
 /public/vendor
 
--- a/apps/readest-app/src/__tests__/hooks/kosyncConflict.test.ts
+++ b/apps/readest-app/src/__tests__/hooks/kosyncConflict.test.ts
@@ -3,6 +3,7 @@ import type { BookDoc } from '@/libs/document';
 import type { FoliateView } from '@/types/view';
 import {
   decideRemoteConflict,
+  isReportedByKOReader,
   resolveRemoteLocalFraction,
   type RemoteFractionResolution,
 } from '@/app/reader/hooks/kosyncProgress';
@@ -24,6 +25,28 @@ const makeView = (fraction: number | null | undefined): FoliateView =>
   }) as unknown as FoliateView;
 
 const bookDoc = {} as BookDoc;
+
+describe('isReportedByKOReader', () => {
+  it('trusts a report with no device string (the common KOReader case)', () => {
+    expect(isReportedByKOReader({})).toBe(true);
+  });
+
+  it('trusts a report whose device does not match a known look-alike', () => {
+    expect(isReportedByKOReader({ device: 'Boox Palma' })).toBe(true);
+    expect(isReportedByKOReader({ device: 'Kobo' })).toBe(true);
+  });
+
+  it('distrusts Kavita (#5109 — non-CREngine percentage on a CREngine-shaped XPointer)', () => {
+    expect(isReportedByKOReader({ device: 'Kavita' })).toBe(false);
+    expect(isReportedByKOReader({ device: 'kavita' })).toBe(false);
+  });
+
+  it('distrusts other known KOReader-sync look-alikes', () => {
+    expect(isReportedByKOReader({ device: 'Komga' })).toBe(false);
+    expect(isReportedByKOReader({ device: 'Stump' })).toBe(false);
+    expect(isReportedByKOReader({ device: 'calibre-web' })).toBe(false);
+  });
+});
 
 describe('resolveRemoteLocalFraction', () => {
   beforeEach(() => mockGetCFIFromXPointer.mockReset());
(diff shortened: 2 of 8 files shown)


adagues:
Closes #4444

This PR fixes the exact scenario described in #4444: Readest treats a KOSync-compatible server's own percentage calculation (e.g. Kavita) as directly comparable to Readest's byte-size-based spine fraction model, when it isn't. See the PR description above for the full root-cause breakdown (this turned out to be two compounding bugs: an XPointer parsing gap plus the percentage drift-anchor being fed a non-CREngine percentage).

chrox (approved):

chrox:
Looks good to me.

chrox:
Heads-up: the percentage drift anchor from this PR has been removed in #6014, which closes #5980. The two other fixes here are kept. Writing up what we found, since the reasoning matters more than the revert.

## What broke

A KOReader position for chapter 5 opened chapter 6, reproducibly, with Readest in Receive Only so no server overwrite was involved. The remote record was valid:

```
progress   = /body/DocFragment[14]/body/div/h3[17]/text().0
percentage = 0.3113
```

`DocFragment[14]` is spine index 13, chapter 5, and the sync server derived `epubcfi(/6/28!/4/2/410:0)` from it, the same index. `resolveSpineSectionIndex` moved it to 15. On that book (Nexus, Harari) Notes plus Index are 44% of all spine bytes, so CREngine's page-based percentage sits a whole chapter away from the byte-size table: chapter 5 spans 18.4189% to 27.8028%, and 31.13% falls inside chapter 6. `h3[17]` cannot exist there (chapter 6 has six `<h3>`, chapter 5 has 18), conversion threw, and the caller fell back to `goToFraction(0.3113)`, landing in chapter 6 anyway.

## Why the premise does not hold

**CREngine guarantees the mapping the anchor was reconstructing.** `crengine/src/epubfmt.cpp` walks the spine once and emits exactly one DocFragment per item, wrapping an SVG spine item in a `SpineSvgWrapper` and standing in a `SpineItemUnsupported` dummy for anything it cannot parse:

> We want to make sure all spineItems get a DocFragment made, even if we don't support or fail parsing them: [...] we won't be inserting a new DocFragment between existing ones and get all xpointers (highlights, last page) invalid because their DocFragment index has been shifted.

So `DocFragment[N]` is foliate section `N - 1` by design.

**The claimed drift runs in a direction CREngine cannot produce.** foliate builds one section per `<itemref>`, unfiltered (`epub.js:789`, mapped 1:1 at `:1209`). CREngine emits at most one fragment per spine item, and `ldomDocumentFragmentWriter::OnTagOpen` opens a DocFragment only when `!insideTag && baseTag==tagname`, so a second `<body>` in one file nests rather than starting a new fragment. The CREngine index is therefore always at or before the foliate index. This PR's reference fixture asserts `DocFragment[16]` (0-based 15) is section 14, and the code comment cites `DocFragment[326]` landing on section 274. Both run the other way, so neither is reachable under any DOM version.

**The reference case was circular.** The only evidence that `DocFragment[16]` should be chapter 9 was the reported percentage, which is the signal in question. The test then encoded that inference as ground truth, so it asserted the heuristic agrees with itself.

**#4444 had a different cause,** which you identified yourself in that thread: the `parseXPointer` gap, whose fallback to the raw percentage moved the reader. For Kavita, `isReportedByKOReader` returns false, so the anchor is skipped and the code uses the nominal index, which is what it did before this PR. The anchor's net effect on #4444 was zero; it stayed active only for real KOReader, which is where it corrupted valid locators.

There is one case where CREngine's mapping genuinely differs: a book pinned to a DOM version below 20240114, where `relaxed_spine` is false and only `application/xhtml+xml` spine items get fragments. That is exactly computable, it can only shift forward, and it needs a non-XHTML spine item to shift at all. No reports of it, and the EPUB in #5980 is 29/29 XHTML. The rule is recorded in the `getCFIFromXPointer` doc comment.

## What was kept

Both of the real fixes in this PR:

- `parseXPointer` handling CREngine's `tag[idx].N` offset format, with its tests.
- The `RemoteFractionResolution` discriminated union for #5065, so an unresolved XPointer forces a conflict instead of being assimilated as "no conflict" and overwritten by auto-push.

`isReportedByKOReader` went with the anchor, since gating the anchor was its only purpose.

Worth flagging for anyone reading this later: the anchor spread beyond its original two call sites, into `useProgressSync` via #5630 and #5866, where it was treated as established mechanism. Both are cleaned up in #6014.

Thanks for the two fixes that stuck. The `tag[idx].N` parse gap in particular was a real find and is still doing work.

PR #5121, opened 2026-07-15
fix(opds): escape malformed XML in proxy

yozlog:
## Summary

Escape unescaped ampersands in buffered XML responses from the OPDS proxy before returning them to the browser.

## Changed

- Added XML sanitization in `/api/opds/proxy`
- Replaced raw `&` with `&amp;` in buffered XML payloads
- Added regression coverage for malformed upstream XML

## Why

Some upstream OPDS servers return XML with unescaped ampersands in metadata URLs. That makes the XML not well-formed and causes browser-side parsing failures. Sanitizing the proxy response avoids the crash.

## Tests

- `opds-proxy-ssrf.test.ts`

### DIFF
--- a/apps/readest-app/src/__tests__/app/api/opds-proxy-ssrf.test.ts
+++ b/apps/readest-app/src/__tests__/app/api/opds-proxy-ssrf.test.ts
@@ -117,8 +117,21 @@ describe('OPDS proxy SSRF guard', () => {
 
     const res = await GET(proxyReq('https://feeds.example.com/catalog.atom'));
     expect(res.status).toBe(200);
-    expect(Buffer.from(await res.arrayBuffer()).toString('latin1')).toContain(
-      'José &amp; María',
+    expect(Buffer.from(await res.arrayBuffer()).toString('latin1')).toContain('José &amp; María');
+  });
+
+  it('leaves ampersands inside CDATA unchanged', async () => {
+    fetchSpy.mockResolvedValueOnce(
+      new Response('<feed><content><![CDATA[Tom & Jerry]]></content></feed>', {
+        status: 200,
+        headers: { 'Content-Type': 'application/xml' },
+      }),
     );
+
+    const res = await GET(proxyReq('https://feeds.example.com/catalog.atom'));
+    const body = await res.text();
+    expect(res.status).toBe(200);
+    expect(body).toContain('<![CDATA[Tom & Jerry]]>');
+    expect(body).not.toContain('<![CDATA[Tom &amp; Jerry]]>');
   });
 });
--- a/apps/readest-app/src/__tests__/app/api/opds-proxy-ssrf.test.ts
+++ b/apps/readest-app/src/__tests__/app/api/opds-proxy-ssrf.test.ts
@@ -93,4 +93,16 @@ describe('OPDS proxy SSRF guard', () => {
     const body = await res.text();
     expect(body).toContain('<feed');
   });
+
+  it('escapes stray ampersands in XML responses', async () => {
+    fetchSpy.mockResolvedValueOnce(
+      new Response('<feed><link href="BOOKS.zip&file=1"/></feed>', {
+        status: 200,
+        headers: { 'Content-Type': 'application/xml' },
+      }),
+    );
+    const res = await GET(proxyReq('https://feeds.example.com/catalog.atom'));
+    expect(res.status).toBe(200);
+    expect(await res.text()).toContain('BOOKS.zip&amp;file=1');
+  });
 });
--- a/apps/readest-app/src/__tests__/app/api/opds-proxy-ssrf.test.ts
+++ b/apps/readest-app/src/__tests__/app/api/opds-proxy-ssrf.test.ts
@@ -105,4 +105,20 @@ describe('OPDS proxy SSRF guard', () => {
     expect(res.status).toBe(200);
     expect(await res.text()).toContain('BOOKS.zip&amp;file=1');
   });
+
+  it('preserves non-utf8 XML bytes while escaping stray ampersands', async () => {
+    const latin1 = Buffer.from('<feed><title>José & María</title></feed>', 'latin1');
+    fetchSpy.mockResolvedValueOnce(
+      new Response(latin1, {
+        status: 200,
+        headers: { 'Content-Type': 'application/xml; charset=iso-8859-1' },
+      }),
+    );
+
+    const res = await GET(proxyReq('https://feeds.example.com/catalog.atom'));
+    expect(res.status).toBe(200);
+    expect(Buffer.from(await res.arrayBuffer()).toString('latin1')).toContain(
+      'José &amp; María',
+    );
+  });
 });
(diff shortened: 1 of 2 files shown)


chrox (commented):
Thanks for the fix — escaping stray `&` in the proxy is a reasonable place to keep non-well-formed feeds from breaking browser-side XML parsing, and the regex itself is sound: it correctly leaves `&amp;`, `&lt;`, `&#160;`, `&#x1F600;` untouched and only escapes genuine stray ampersands. For UTF-8 feeds (the common case) this does exactly what it claims.

My concern is the **decode-as-UTF-8 / re-encode-as-UTF-8 round trip applied to the whole document**, which introduces two behaviors the current happy-path test doesn't cover.

### 1. Corrupts non-UTF-8 XML feeds (regression) — main issue

The code hard-codes `new TextDecoder('utf-8')` regardless of the feed's actual charset. Any OPDS/Atom feed served as ISO-8859-1, Windows-1252, GBK, Shift-JIS, etc. (older Calibre/regional servers do this, and the `Content-Type` / XML prolog declares it) gets its non-ASCII bytes mangled. Repro with a Latin-1 `"José & María"`:

```
original bytes:   … 4a 6f 73 e9 20 26 …                       (0xe9 = é in Latin-1)
re-encoded bytes: … 4a 6f 73 ef bf bd 20 26 61 6d 70 3b …     (é → U+FFFD replacement char)
```

Before this PR those bytes passed through untouched and the browser decoded them per the declared charset; after it, `é`/`í` become replacement characters. So the fix trades one class of broken feeds (stray `&`) for silent data corruption in a different class. Web-only and non-UTF-8-only, but a real regression for those users.

**Suggestion:** do the replacement without changing the byte encoding. Since `&` (0x26) and every entity character (`a-zA-Z0-9 # x ;`) are single-byte ASCII in every ASCII-superset encoding, a Latin-1 round-trip (or a direct byte scan) preserves all original bytes and only inserts the ASCII `amp;`:

```ts
if (contentType.toLowerCase().includes('xml')) {
  const bytes = new Uint8Array(buf);
  let text = '';
  for (const b of bytes) text += String.fromCharCode(b);   // lossless byte↔char
  text = text.replace(/&(?!([a-zA-Z0-9]+|#[0-9]+|#x[0-9a-fA-F]+);)/g, '&amp;');
  const out = new Uint8Array(text.length);
  for (let i = 0; i < text.length; i++) out[i] = text.charCodeAt(i) & 0xff;
  buf = out.buffer;
}
```

### 2. Over-escapes inside CDATA sections — edge case

The regex has no notion of CDATA, so literal ampersands inside `<![CDATA[…]]>` get escaped even though they shouldn't:

```
<![CDATA[Tom & Jerry & Co]]>  →  <![CDATA[Tom &amp; Jerry &amp; Co]]>
```

OPDS feeds commonly wrap HTML `<summary>`/`<content>` descriptions in CDATA, so a description like "Tom & Jerry" would then render the literal text "Tom &amp; Jerry". Lower severity than #1 and harder to fix perfectly, but worth a comment noting the limitation (skipping CDATA regions would need the replace to run only outside `<![CDATA[ … ]]>` spans).

### Minor

The existing test only covers the ASCII stray-`&` happy path. If you keep this approach, adding a non-UTF-8 (Latin-1) round-trip assertion and a CDATA assertion would lock in the intended behavior.

Directionally correct and it fixes a genuine crash, but I'd suggest replacing the whole-document UTF-8 re-encode with a byte-preserving replacement (#1) before merging — otherwise it swaps a visible parse failure for silent metadata corruption on non-UTF-8 feeds.

yozlog:
> Thanks for the fix — escaping stray `&` in the proxy is a reasonable place to keep non-well-formed feeds from breaking browser-side XML parsing, and the regex itself is sound: it correctly leaves `&amp;`, `&lt;`, `&#160;`, `&#x1F600;` untouched and only escapes genuine stray ampersands. For UTF-8 feeds (the common case) this does exactly what it claims.
> 
> My concern is the **decode-as-UTF-8 / re-encode-as-UTF-8 round trip applied to the whole document**, which introduces two behaviors the current happy-path test doesn't cover.
> ### 1. Corrupts non-UTF-8 XML feeds (regression) — main issue
> 
> The code hard-codes `new TextDecoder('utf-8')` regardless of the feed's actual charset. Any OPDS/Atom feed served as ISO-8859-1, Windows-1252, GBK, Shift-JIS, etc. (older Calibre/regional servers do this, and the `Content-Type` / XML prolog declares it) gets its non-ASCII bytes mangled. Repro with a Latin-1 `"José & María"`:
> 
> ```
> original bytes:   … 4a 6f 73 e9 20 26 …                       (0xe9 = é in Latin-1)
> re-encoded bytes: … 4a 6f 73 ef bf bd 20 26 61 6d 70 3b …     (é → U+FFFD replacement char)
> ```
> 
> Before this PR those bytes passed through untouched and the browser decoded them per the declared charset; after it, `é`/`í` become replacement characters. So the fix trades one class of broken feeds (stray `&`) for silent data corruption in a different class. Web-only and non-UTF-8-only, but a real regression for those users.
> 
> **Suggestion:** do the replacement without changing the byte encoding. Since `&` (0x26) and every entity character (`a-zA-Z0-9 # x ;`) are single-byte ASCII in every ASCII-superset encoding, a Latin-1 round-trip (or a direct byte scan) preserves all original bytes and only inserts the ASCII `amp;`:
> 
> ```ts
> if (contentType.toLowerCase().includes('xml')) {
>   const bytes = new Uint8Array(buf);
>   let text = '';
>   for (const b of bytes) text += String.fromCharCode(b);   // lossless byte↔char
>   text = text.replace(/&(?!([a-zA-Z0-9]+|#[0-9]+|#x[0-9a-fA-F]+);)/g, '&amp;');
>   const out = new Uint8Array(text.length);
>   for (let i = 0; i < text.length; i++) out[i] = text.charCodeAt(i) & 0xff;
>   buf = out.buffer;
> }
> ```
> 
> ### 2. Over-escapes inside CDATA sections — edge case
> 
> The regex has no notion of CDATA, so literal ampersands inside `<![CDATA[…]]>` get escaped even though they shouldn't:
> 
> ```
> <![CDATA[Tom & Jerry & Co]]>  →  <![CDATA[Tom &amp; Jerry &amp; Co]]>
> ```
> 
> OPDS feeds commonly wrap HTML `<summary>`/`<content>` descriptions in CDATA, so a description like "Tom & Jerry" would then render the literal text "Tom & Jerry". Lower severity than #1 and harder to fix perfectly, but worth a comment noting the limitation (skipping CDATA regions would need the replace to run only outside `<![CDATA[ … ]]>` spans).
> ### Minor
> 
> The existing test only covers the ASCII stray-`&` happy path. If you keep this approach, adding a non-UTF-8 (Latin-1) round-trip assertion and a CDATA assertion would lock in the intended behavior.
> 
> Directionally correct and it fixes a genuine crash, but I'd suggest replacing the whole-document UTF-8 re-encode with a byte-preserving replacement (#1) before merging — otherwise it swaps a visible parse failure for silent metadata corruption on non-UTF-8 feeds.

Thanks for the review. I’ve updated the patch to address the issues you pointed out and split the fix into two commits.

chrox:
Thanks.

PR #5128, opened 2026-07-15
chore: replace outdated `react-color` package with `react-colorful`

dastarruer:
The old color picker was completely unstyled:

image

So I replaced it with `react-colorful`'s `HexColorPicker`, which is smaller, and doesn't look outdated:

https://github.com/user-attachments/assets/6630d1c7-147d-43c4-b1ce-40deafe5053e

### DIFF
--- a/apps/readest-app/package.json
+++ b/apps/readest-app/package.json
@@ -192,7 +192,7 @@
     "overlayscrollbars-react": "^0.5.6",
     "posthog-js": "^1.246.0",
     "react": "19.2.5",
-    "react-color": "^2.19.3",
+    "react-colorful": "^5.8.0",
     "react-dom": "19.2.5",
     "react-i18next": "^15.2.0",
     "react-icons": "^5.4.0",
--- a/apps/readest-app/src/__tests__/components/HighlightColorsEditor.test.tsx
+++ b/apps/readest-app/src/__tests__/components/HighlightColorsEditor.test.tsx
@@ -40,6 +40,7 @@ vi.mock('react-colorful', () => ({
       />
     );
   },
+  HexColorInput: () => <input data-testid='mock-hex-input' />,
 }));
 
 afterEach(() => {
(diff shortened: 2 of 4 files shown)


dastarruer:
You'll probably notice that if editing an already-added color with the color picker, it becomes very laggy. I'm not sure why this is; was hoping you could help?

chrox:
It seems the new color input doesn’t support entering HEX values. That would be a feature regression. I’m afraid I don’t have time to look into this right now.

dastarruer:
> It seems the new color input doesn’t support entering HEX values. That would be a feature regression. I’m afraid I don’t have time to look into this right now.

That’s fine, it should be a simple fix anyways. Will get on it

dastarruer:
I've added an input box + styling:

image

It'd be nice to prepend the hex string with a `#` but `HexColorInput` doesn't seem to have a way to do this.

I also have not tested the `if (compact)` branch, since after looking through the code, I can't seem to find where `compact` is ever set to `true`.

Also the same lag issue is still present. Will look into it myself.

Edit: turns out the `compact` flag was unused, so I've removed it.

dastarruer:
Also added eink support:

image

dastarruer:
image

Would you support redesigning this to look similar to the highlight color page? Just to a) maintain visual consistency with the rest of the settings and b) simplify the `ColorInput` logic.

chrox:
Sure, please feel free to redesign/refactor it.

dastarruer:
> Sure, please feel free to redesign/refactor it.

This will be a decently-sized change, so I'm opening a separate PR for this.

dastarruer:
Given the DESIGN.md file, do I remove the box shadow from the picker?

chrox:
> Given the DESIGN.md file, do I remove the box shadow from the picker?

It only enforces that no box shadow will be there in e-ink mode. In non-eink mode box shadow is allowed.

dastarruer:
> > Given the DESIGN.md file, do I remove the box shadow from the picker?
> 
> It only enforces that no box shadow will be there in e-ink mode. In non-eink mode box shadow is allowed.

Just fixed this.

PR #5135, opened 2026-07-16
fix: change dictionary icon

dastarruer:
The old dictionary icon was too ambiguous, and I constantly mistook the find icon for the dictionary icon because of it. This changes the dictionary icon to a book:

image

### DIFF
--- a/apps/readest-app/src/app/reader/components/annotator/AnnotationTools.tsx
+++ b/apps/readest-app/src/app/reader/components/annotator/AnnotationTools.tsx
@@ -2,7 +2,8 @@ import { IconType } from 'react-icons';
 import { FiSearch } from 'react-icons/fi';
 import { FiCopy } from 'react-icons/fi';
 import { FiShare } from 'react-icons/fi';
-import { PiTextAa, PiHighlighterFill } from 'react-icons/pi';
+import { PiHighlighterFill } from 'react-icons/pi';
+import { LuBookA } from 'react-icons/lu';
 import { BsPencilSquare } from 'react-icons/bs';
 import { BsTranslate } from 'react-icons/bs';
 import { FaHeadphones } from 'react-icons/fa6';
@@ -66,7 +67,7 @@ export const annotationToolButtons = createAnnotationToolButtons([
     type: 'dictionary',
     label: _('Dictionary'),
     tooltip: _('Look up text in dictionary after selection'),
-    Icon: PiTextAa,
+    Icon: LuBookA,
     quickAction: true,
   },
   {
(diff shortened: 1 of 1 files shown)


chrox:
The new icon is too prominent and doesn’t match the style of the other icons.

chrox:
Specifically, the new icon is filled, while the other icons are outlined. It also looks more like a regular book than a dictionary, which could make it even more confusing.

dastarruer:
> The new icon is too prominent and doesn’t match the style of the other icons.

Actually, while I'm at it, can I ask if you'd be willing to switch the icon library from `react-icons` to something like lucide? I'm asking because currently Readest is using icons from completely different libraries (ie. this toolbar alone uses Pi, Fi, Bs, etc.) and it already looks quite disjointed. I was going to create an issue and post a proof-of-concept screenshot of this, but thought I could ask here first.

Also yes, you're right, I'll change the icon.

chrox:
The ideal approach would be to design a complete icon set for Readest. I’m afraid Lucide doesn’t cover all the icons we already use, but you’re welcome to create a proof of concept first.

dastarruer:
> I’m afraid Lucide doesn’t cover all the icons we already use, but you’re welcome to create a proof of concept first.

Perhaps lucide + another library if necessary. Though I would like to know before I start which icons lucide doesnt provide?

Also I'll try making a few proof of concepts with different libraries to see which one looks best.

dastarruer:
Basically a toss-up between one of these icons: https://react-icons.github.io/react-icons/search/#q=open%20book%20text

This is what I've gone with for now:

image

chrox:
It’s not an improvement over the previous icon. It still looks like a book rather than a dictionary.

Readest’s UI is localized into more than 30 languages and runs on a wide range of devices, including smartphones, e-readers, tablets, laptops/desktops, and even TVs. To ensure the interface works well across all languages and screen sizes, we’ve intentionally omitted text labels for most icons. As a result, each icon needs to clearly convey its function.

For that reason, we don’t choose icons solely for visual consistency. An icon must also be semantically appropriate, which is a different design philosophy from some other projects.

In the long term, we plan to design a complete custom icon set for Readest to address this issue properly.

dastarruer:
> It’s not an improvement over the previous icon. It still looks like a book rather than a dictionary.

How about this?

image

chrox:
Definitely no. The Aa icon is already used for the Font & Layout settings in Readest for a very long time. Even if it were the perfect icon for a dictionary, we still couldn’t use it here because the cost of breaking existing users’ mental model is simply too high.

dastarruer:
Ok, then I'll just keep pasting icons until we get one that looks good.

First candidate:

image

I keep using books because to me, somehow it looks like a dictionary.

dastarruer:
image

This one could be good?

dastarruer:
image

Last one

chrox:
> image
> This one could be good?

Could you try this and share a screenshot?

chrox:
Could you also take a screenshot of the original annotation tool bar, so that we can compare it side by side?

dastarruer:
> Could you try this and share a screenshot?

image

Also the other icon:

image

> Could you also take a screenshot of the original annotation tool bar, so that we can compare it side by side?

image

chrox:
> > Could you try this and share a screenshot?
> 
> image

I think this one is better.

dastarruer:
> > image
> 
> I think this one is better.

Just pushed the change.

PR #5152, opened 2026-07-16
feat: redesign custom theme creation menu

dastarruer:
Gives the custom theme creation menu a much-needed face-lift, as discussed in #5128. 

# Screenshot

image

Likely needs more work, but thought I'd get feedback first before proceeding.


dastarruer:
While writing this, it seems to me that the primary color is vastly underutilized throughout the app. Perhaps it should be more interspersed throughout the UI to give the app a pop of color?

Edit: For now I can only think of using it for the progress bar in reading mode and toggles in settings panel.

chrox:
> While writing this, it seems to me that the primary color is vastly underutilized throughout the app. Perhaps it should be more interspersed throughout the UI to give the app a pop of color?
> 
> Edit: For now I can only think of using it for the progress bar in reading mode and toggles in settings panel.

Please don’t. The UI is intentionally designed around background and foreground colors. The primary (accent) color is reserved for places where it serves a clear purpose, not as a decorative element. I’d rather not overuse it.

chrox:
Please checkout this first for the design system: https://github.com/readest/readest/blob/main/apps/readest-app/DESIGN.md

dastarruer:
> Please don’t. The UI is intentionally designed around background and foreground colors. The primary (accent) color is reserved for places where it serves a clear purpose, not as a decorative element. I’d rather not overuse it.

Not planning on overusing it. However:

image

It's a little difficult to tell at a glance whether this option is enabled or not. Here might be a good place to use the primary color.

This button should probably use it:

image

And then I also mentioned the progress bar. However, this just seems to be a visual bug on ipad only. Will open an issue.

chrox:
Please also verify the app’s visual appearance against the latest version of Readest running on Chrome, which should be treated as the reference implementation and ground truth. Outdated browser engines, such as WebKitGTK, may not accurately render the intended design.

chrox:
> # Screenshot
> image
> Likely needs more work, but thought I'd get feedback first before proceeding.

It's much better now. Could you also use the `btn-contrast` for the 'Save' button? For most primary button we should use `btn-contrast`, only for those Call-To-Action button we will use the `btn-primary`. We will update the `DESIGN.md` for this.

chrox:
DESIGN.md is updated in https://github.com/readest/readest/pull/5155.

apps/readest-app/src/components/settings/color/ThemeEditor.tsx
```diff
@@ -33,7 +33,7 @@ const ThemePreview: React.FC<{
       >
         <p className='mb-2 whitespace-pre-line text-xs'>
           {_(
-            "All the world's a stage,\nAnd all the men and women merely players;\nThey have their exits and their entrances,\nAnd one man in his time plays many parts,\nHis acts being seven ages.\n\n— William Shakespeare",
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
```
chrox: Why this?

dastarruer: I did this because the Shakespeare text was too long, and causes users to have to scroll down to see the full preview, which is inconvenient:

image

You'll see that fully scrolled down, part of the theme selection is actually obscured, making the whole panel feel claustrophobic.

 Lorem ipsum text seemed like a better placeholder. I explained this in the commit message, but I likely should've added it as a comment. Can this be reverted?

chrox: I’d prefer not to revert this. A shorter placeholder does not guarantee that the panel will fit without scrolling, since the result depends largely on screen size and the length of the text.

Lorem ipsum also tends to give me the impression that the interface is still a unfinished design mockup rather than from a real product. And it's weird to translate it into the 30 localized languages, or even more weird if we choose not to translate it.

If we decide to use a shorter placeholder, I’d prefer a passage from classic literature written in Early Modern English, ideally Shakespeare. Since “Readest” itself is from an Early Modern English word, it could serve as a small Easter egg while still feeling like part of the product.

dastarruer: It’s a little out there for an Easter egg tbh… though I see what you’re saying. You want a literary passage instead of Lorem ipsum. Problem is, even without the scrolling issue I brought up, I still think it’s too lengthy. Will see if I can come up with something that’s both short yet well-known.

chrox: AI will do a much better job than human beings for this. They are built for this. Please try it.

dastarruer: Good idea, here’s a few I got that I think would fit:

- To be or not to be
- All the world’s a stage (since this is the most well known part of the quote anyways)
- All that glitters is not gold (didn’t know this was Shakespeare til now but hey! you learn something new everyday)

PR #5157, opened 2026-07-17
feat(tts): add 0.8x and 0.85x tts speech speed presets

### LINKED ISSUE #5078: FR: More TTS Speeds

Listening to books via TTS at a 'better' speed

There used to be a more granular slider with TTS which has since been replaced with predertmined speeds. I would like something between 0.75X and 0.9X.

Using 0.9X but it's too fast.

### Additional Context

Image

gojodennis:
Closes #5078


dastarruer:
A better fix would likely to be just reimplementing the speed slider that was removed for some reason, rather than just adding more options.

gojodennis:
> A better fix would likely to be just reimplementing the speed slider that was removed for some reason, rather than just adding more options.

ay ay captain

Issue #5006, opened 2026-07-07
Koreader Plugin: Recent v0.11.18 Update - Increased slowness when opening book when auto-sync enabled


anon3792737:
## Bug Description

With the most recent plugin update, Koreader is noticeably slower when opening a book. This doesn't happen when wifi and /or auto sync is disabled. "Application Not Responding" frequenlty appears.

## Expected behavior

No slowness when opening book.

## Technical inputs

Operating System: Android 15. Boox Palma Pro 2

Readest Version:

V0.11.18

# Notes

I often get the "application not responding message" when switching books. Usually the book will then load. Web service call related?

I suspect so because if I turn off "Auto Sync" in the readest plugin setting, everything is very fast .

I did some code exploration with my harness, and the bots suggestions _seem_ in line with what I see, in that rapid swtiching between books appears to overload the UI thread.

**Short Bot explanation:**
Rapid book switching appears to amplify the issue
because each switch triggers AutoSync work for
both the old book and the new book. The old book
close path may push config/notes/stats and sync
the library, while the new book open path pulls
config/notes/stats. On Android these calls appear
to run synchronously on the UI thread, so rapid
switching creates a backlog/burst of blocking sync
work and KOReader hangs until the calls complete
or time out. 

**Longer Bot explanation:**
[Gpt 5.5 XHIGH + GLM 5.2 High generated suggested.html](https://github.com/user-attachments/files/29769974/readest-autosync-android-anr-github-issue.html)

anon3792737:
changed title to better reflect issue. Unable to reproduce crashing on my side.

Issue #4138, opened 2026-05-12
Readest - Koreader Sync - Newest update results in Readest book losing categorization + need reupload to cloud on readest device 


anon3792737:
## Bug Description

As of the most recent update when I read a book on koreader and then I pick up the book on Readest. I notice that in the Readest app the book now appears without the former categorization and now appears in the root of the Readest directory. It also appears like it needs to be re-uploaded to the cloud. This happens on Redest for Android and Readest for iOS. I have not checked on a regular computer. 

> **Example:**
I can get screenshots if necessary.

## Expected behavior

I would expect the categorization in Readest to stay the same and for the book not to require an upload to the cloud. 

## Technical inputs

Operating System:

Readest Version:

> **Example:**
> Operating System: ipados 26.3.1 ; Android 13
> Readest Version: 0.11.1

chrox:
What's the Koreader version?

chrox:
Please also upgrade the KOReader to at least version 2026.03.

anon3792737:
2026.03 - Snowflake

PR #5187, opened 2026-07-18
fix(koplugin): keep auto-sync off the book open/close UI thread (#5006)

chrox:
## Problem

With auto-sync on, opening or closing a book froze the KOReader UI for seconds, and rapid book switching triggered "Application Not Responding" (#5006).

Root cause: the sync HTTP runs synchronously on the UI thread. The Turbo async path is dead on shipping builds (`DUSE_TURBO_LIB` defaults to `false`, so `UIManager.looper` is nil and Spore's `AsyncHTTP` middleware falls back to blocking socket HTTP; KOReader's own `KOSyncClient` has the same vestigial code). The block was amplified by doing far more work on the critical path than KOSync does:

- **Close** (`onCloseDocument`) ran `pushBookConfig` + `pushBookNotes` + `pushBookStats` + `syncBooksLibrary("both")`, and `"both"` did a **full library pull of every row (~3s) then a push**, inline via `goOnlineToRun`.
- **Open** (`onReaderReady`) fired three blocking per-book pulls on `nextTick`, and every rapid switch stacked another three.

## Fix

Keep the heavy work off the open/close critical path (the approach KOReader's own KOSync uses: tiny payload on the hot path, defer the rest).

**Close path, push-only + targeted** (`pushOpenBook`): push only the open book's library row instead of the full library pull+push. Closing a book only needs to persist this book's changes. `touchOpenBook` returns the row with the cloud-side `uploaded_at` / `metadata` / `group_id` still in place from the last pull, so the single-book push preserves them with no explicit-null wipe (#4138), same as the existing interactive "push" path. Cross-device library reconciliation still runs when the Library widget opens (it already did `"both"` there).

**Open path, deferred + cancellable pull**: the per-book pull is scheduled (reader paints and becomes interactive first) and stored so a rapid book switch cancels it in `onCloseWidget` instead of stacking blocking round-trips.

Net: per-book progress/notes/stats still sync on open and close; only the heavy library-wide reconciliation moves off the reading hot path.

## Tests

- `spec/main_close_spec.lua`: close path dispatches push-only + `pushOpenBook` with no library pull; `pushOpenBook` pushes only the open book's row.
- `spec/main_open_spec.lua`: open pull is deferred, cancelled on close (rapid switching), and does not stack duplicates.
- `pnpm lint:lua` clean, `pnpm test:lua` green.

Note: depends on the JSON-null push crash fix in #5186 (the other half of #5006); the two touch disjoint files and can merge independently.

### DIFF
--- a/apps/readest.koplugin/main.lua
+++ b/apps/readest.koplugin/main.lua
@@ -22,6 +22,10 @@ local ReadestSync = WidgetContainer:new{
 }
 
 local API_CALL_DEBOUNCE_DELAY = 30
+-- Delay before the on-open pull runs, so the reader paints and becomes
+-- interactive first and rapid book switching coalesces to a single pull for
+-- the book you settle on, instead of stacking blocking round-trips (#5006).
+local READER_READY_PULL_DELAY = 1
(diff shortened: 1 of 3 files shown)


chrox:
Consolidated into #5186 as requested (single PR for both halves of #5006).

PR #5189, opened 2026-07-18
fix(opds): restore Calibre pipe-escaped commas in author names and join authors with &

### LINKED ISSUE #5183: FR: different names separators

In the OPDS Catalogs, author names are split into two parts: family name followed by given names, separated by `|`. For example: `Doe| John Walter`.

This becomes confusing when there are multiple authors, as the result looks like: `Doe| John Walter, Smith| James Richard`.

In Western typography, the `|` character is generally perceived as a stronger separator than `,`. As a result, the example above can be read as `Doe / John Walter Smith / James Richard`, making it unclear where one author's name ends and the next begins.

For comparison, Calibre uses `&` to separate multiple authors: `Doe, John Walter & Smith, James Richard`, which is much easier to read.

Suggestion: follow the Calibre convention, `&` between authors, `,` between the family name and the first names.

_No response_

### Additional Context

_No response_

chrox:
### What

Fixes #5183

OPDS catalogs backed by Calibre libraries (Calibre-Web, Calibre-Web-NextGen) showed author names like `Doe| John Walter, Smith| James Richard`, which is hard to parse when several authors each carry a family-name/given-names split.

### Why

The `|` comes from the server side: Calibre stores commas in author names escaped as pipes (`Doe, John` is stored as `Doe| John` in `authors.name`), and Calibre-Web's OPDS template emits `{{author.name}}` raw without the `replace('|', ',')` filter its HTML templates apply. Readest displayed the escaped form verbatim and joined multiple authors with `, `, compounding the ambiguity.

### How

- New `formatContributorName()` in `opdsUtils.ts` restores the escaped commas (`|` to `,`).
- `PublicationCard` (catalog view) and `PublicationView` (book view) normalize each author name and join multiple authors with `&`, following the Calibre convention suggested in the issue: `Doe, John Walter & Smith, James Richard`.
- Author links and the streamed-book author string pick up the normalized names as well.

### Tests

- Unit tests for `formatContributorName`.
- `PublicationView` test: normalized names, `&` separator, linked author stays clickable.
- `FeedView` card test: normalized names with `&` join (the react-virtuoso mock now also provides `VirtuosoGrid`, which the publications grid uses).

### DIFF
--- a/apps/readest-app/src/__tests__/app/opds/feed-view.test.tsx
+++ b/apps/readest-app/src/__tests__/app/opds/feed-view.test.tsx
@@ -15,22 +15,22 @@ vi.mock('@/components/CachedImage', () => ({
 // synchronously via itemContent, matching the TOCView/BooknoteView test mocks.
 vi.mock('react-virtuoso', async () => {
   const React = await import('react');
+  const renderAll = (
+    {
+      totalCount,
+      itemContent,
+    }: { totalCount: number; itemContent: (index: number) => React.ReactNode },
+    _ref: React.Ref<unknown>,
+  ) => (
+    <div>
+      {Array.from({ length: totalCount }, (_, index) => (
+        <div key={index}>{itemContent(index)}</div>
+      ))}
+    </div>
+  );
   return {
-    Virtuoso: React.forwardRef(
-      (
-        {
-          totalCount,
-          itemContent,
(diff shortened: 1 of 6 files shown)


chloeroform:
Thanks! Looking forward to testing this.

PR #5194, opened 2026-07-18
feat: improve accuracy of time remaining calculation

### LINKED ISSUE #5116: Adaptive remaining-time estimation based on actual reading/TTS speed
### Bug Description

The time remaining in a chapter is completely out of sync with what the tts player reports:

Image

### Steps to Reproduce

1. Enable 'Show Remaining Time' in Layout settings
2. Go to the beginning of a chapter in an epub
3. Start tts player to compare expected vs actual time left

### Expected Behavior

The time remaining in the footer should at least be the same as that of the tts player.

### Reproduction File

_No response_

### Operating System

iOS/iPadOS

### OS Version

iPadOS 26.5

### Readest Version

Version 0.11.18 (WebView 605.1.15)

### Custom CSS

_No response_

dastarruer:
Closes #5116.

Previously, time left calculations did not take into account the user's natural reading speed. This PR fixes that by using the user's seconds-per-page rate.

However, it would probably be better to use words per minute instead. This is just a hacked together version for now.

### DIFF
--- a/apps/readest-app/src/app/library/utils/libraryUtils.ts
+++ b/apps/readest-app/src/app/library/utils/libraryUtils.ts
@@ -237,9 +237,11 @@ export const convertPagesToTimeRemainingMinutes = (
   pagesLeft: number,
   medianPageDurationSecs?: number,
 ): number => {
-  return medianPageDurationSecs
+  const minutes = medianPageDurationSecs
     ? Math.round((pagesLeft * medianPageDurationSecs) / 60)
     : Math.round((pagesLeft * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT); // Fall back to less precise calculation
+
+  return Math.max(1, minutes);
 };
 
 /**
--- a/apps/readest-app/src/app/library/utils/libraryUtils.ts
+++ b/apps/readest-app/src/app/library/utils/libraryUtils.ts
@@ -212,11 +212,11 @@ export const convertPagesToTimeRemainingMinutes = (
   pagesLeft: number,
   medianPageDurationSecs?: number,
 ): number => {
-  const minutes = medianPageDurationSecs
-    ? Math.round((pagesLeft * medianPageDurationSecs) / 60)
-    : Math.round((pagesLeft * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT); // Fall back to less precise calculation
-
-  return Math.max(1, minutes);
+  // Prefer the reader's own pace; fall back to the coarse global estimate.
+  const minutesPerPage = medianPageDurationSecs
+    ? medianPageDurationSecs / 60
+    : SIZE_PER_LOC / SIZE_PER_TIME_UNIT;
+  return Math.max(1, Math.round(pagesLeft * minutesPerPage));
 };
 
 /**
@@ -240,11 +240,8 @@ export const getDisplayedTimeRemaining = (
  * Remaining minutes for a shelf item, or `undefined` when its tile can show no
  * time at all — that includes every group, since a group tile renders no progress.
  */
-const getShelfItemTimeRemaining = (
-  item: Book | BooksGroup,
-  medianPageDurationSecs?: number,
-): number | undefined =>
-  'books' in item ? undefined : getDisplayedTimeRemaining(item, medianPageDurationSecs);
+const getShelfItemTimeRemaining = (item: Book | BooksGroup): number | undefined =>
+  'books' in item ? undefined : getDisplayedTimeRemaining(item);
 
 /**
  * Wrap a comparator that has *already* had the sort direction applied, so items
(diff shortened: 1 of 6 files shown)


apps/readest-app/src/app/library/utils/libraryUtils.ts
```diff
@@ -199,9 +202,46 @@ const getBookReadRatio = (book: Book): number => {
...
+      setMedianPageDurationSecs(medianPageDurationSecs);
+    };
+
+    void load();
+  });
+
+  return medianPageDurationSecs;
+};
+
+export const getTimeRemainingMinutes = (
```
dastarruer, line 202: Should I remove this in favor of `convertPagesToTimeRemainingMinutes`?

apps/readest-app/src/app/library/utils/libraryUtils.ts
```diff
@@ -199,9 +202,46 @@ const getBookReadRatio = (book: Book): number => {
...
+  book: Book,
+  medianPageDurationSecs?: number,
+): number | undefined => {
   const pagesLeft = book.progress ? book.progress[1] - book.progress[0] : undefined;
-  return pagesLeft ? Math.round((pagesLeft * SIZE_PER_LOC) / SIZE_PER_TIME_UNIT) : undefined;
+  if (!pagesLeft) return undefined;
+  return convertPagesToTimeRemainingMinutes(pagesLeft, medianPageDurationSecs);
+};
+
+export const convertPagesToTimeRemainingMinutes = (
```
dastarruer, line 211: There's definitely a better name I could use here.

dastarruer:
I'm debating whether taking the time to implement words per minute rather than seconds per page would actually be worth it. Not sure if it'd be that much more accurate?

apps/readest-app/src/app/library/utils/libraryUtils.ts
```diff
@@ -199,9 +202,46 @@ const getBookReadRatio = (book: Book): number => {
   return current / total;
 };
 
-export const getTimeRemainingMinutes = (book: Book): number | undefined => {
+export const useMedianPageDurationSecs = (bookMd5?: string): number | null => {
```
chrox: This is a hook and it's better to be inside a hook directory.

chrox (commented):
Thanks for tackling this! The plumbing is reasonable, but the core median calculation is broken two ways and the loading hook will hit the DB much harder than intended. A couple of small unit tests would have caught the median bugs (this repo is test-first).

### 1. `getMedianPageDurationSecs` never sorts by duration, so it isn't a median

`src/services/statistics/statisticsDb.ts`

```sql
SELECT duration FROM page_stat_data WHERE id_book = ? ORDER BY start_time DESC LIMIT 50
```
```ts
const pageDurations = rows.map((d) => d['duration']);
const mid = Math.floor(pageDurations.length / 2);
```

The rows come back ordered by `start_time` (the most-recent 50 pages), and the JS then takes the middle element **without sorting by value**. `pageDurations[mid]` is just whichever page happens to sit in the middle of the time ordering, not the median. Since the whole estimate rests on this number, the result is effectively arbitrary.

Fix: sort by value after fetching the recent window.

```ts
const pageDurations = rows.map((r) => r.duration).sort((a, b) => a - b);
```

### 2. Operator-precedence bug in the even-count branch

```ts
return pageDurations.length % 2 !== 0
  ? (pageDurations[mid] ?? 0)
  : (pageDurations[mid - 1] ?? 0 + (pageDurations[mid] ?? 0)) / 2;
```

`+` binds tighter than `??`, so the even branch parses as `(pageDurations[mid - 1] ?? (0 + (pageDurations[mid] ?? 0))) / 2`. For any non-empty even-length array `pageDurations[mid - 1]` is defined, so this returns `pageDurations[mid - 1] / 2` — half of one element, not the average of the two middle values. e.g. `[10, 30]` yields `5` instead of `20`.

Fix:

```ts
: ((pageDurations[mid - 1] ?? 0) + (pageDurations[mid] ?? 0)) / 2;
```

### 3. `useMedianPageDurationSecs` effect has no dependency array

`src/app/library/utils/libraryUtils.ts`

```ts
useEffect(() => {
  if (!appService || !bookMd5) return;
  const load = async () => {
    const db = await StatisticsDb.open(appService);
    const book = await db.getBookByMd5(bookMd5);
    ...
  };
  void load();
});   // <- no deps array
```

Without a dependency array the effect runs after **every render**, each time opening `StatisticsDb` and running two queries. In `ProgressBar` (which re-renders on every progress tick while reading) and in `ReadingProgress` (one instance per book in the library grid), this fires DB work continuously instead of once per book.

Fix: `}, [appService, bookMd5]);`

### Minor / non-blocking

- **Dead sort plumbing.** `compareBookByKey`, `withTimeRemainingLast`, and `getShelfItemTimeRemaining` all gained a `medianPageDurationSecs` parameter, but no caller passes it, so the library sort still uses the fallback formula. A single scalar also can't represent per-book medians inside a synchronous comparator — sorting by accurate per-book time would need prefetched values. As written these params are unused; consider dropping them (or wiring them up properly) so the signatures don't imply behavior that isn't there.
- **No tests.** `getMedianPageDurationSecs` and `convertPagesToTimeRemainingMinutes` are pure and easy to unit-test; both bugs above are exactly what `[10,30] -> 20` / `[10,20,30] -> 20` cases would surface.
- **Shadowing.** Inside `load`, `const medianPageDurationSecs = await db.getMedianPageDurationSecs(id)` shadows the state variable of the same name. Harmless but confusing.
- **Minor edge case.** `getTimeRemainingMinutes` guards `if (!pagesLeft) return undefined;`, so a negative `pagesLeft` slips through to `convertPagesToTimeRemainingMinutes`, where `Math.max(1, negative)` clamps to `1`. Unlikely, low impact.

Bottom line: items 1 and 2 mean the estimate is incorrect, so they need fixing before this does what the title claims, and item 3 is a real perf regression on the library grid and reader. All three are small changes.

chrox:
I don't know if any coding agent is used in your workflow, but Claude Fable 5 / Opus 4.8 or OpenAI GPT 5.6/5.5 are very likely able to review and find the issues in this PR and I would recommend integrating one of the models in your workflow.

dastarruer:
> I don't know if any coding agent is used in your workflow, but Claude Fable 5 / Opus 4.8 or OpenAI GPT 5.6/5.5 are very likely able to review and find the issues in this PR and I would recommend integrating one of the models in your workflow.

As much as I’d like to, they cost a lot of money in tokens. Perhaps this repo could integrate coderabbit for PR reviews, seeing as not everyone can access frontier models? I use coderabbit for my personal projects, though the free tier may be a little stingy (1 review per hour per contributor if I recall correctly) for this project.

dastarruer:
> ```ts
> const pageDurations = rows.map((d) => d['duration']);
> const mid = Math.floor(pageDurations.length / 2);
> ```

Can I not just add ORDER BY duration to the query?

chrox:
I will setup a code review workflow on GitHub soon. Thanks for the info.

dastarruer:
> > ```ts
> > const pageDurations = rows.map((d) => d['duration']);
> > const mid = Math.floor(pageDurations.length / 2);
> > ```
> 
> Can I not just add ORDER BY duration to the query?

@chrox I see you already fixed this, but wouldn’t it be more efficient to sort in the query? Or am I wrong?

chrox:
> > > ```ts
> > > const pageDurations = rows.map((d) => d['duration']);
> > > const mid = Math.floor(pageDurations.length / 2);
> > > ```
> > 
> > 
> > Can I not just add ORDER BY duration to the query?
> 
> @chrox I see you already fixed this, but wouldn’t it be more efficient to sort in the query? Or am I wrong?

You can calculate the median in SQL using a more complex expression, but that would make the query less simple and harder to maintain. It would also be more vulnerable to schema changes, and the SQL is not covered by linting or a test harness.

PR #5197, opened 2026-07-18
feat: use shorter quote in theme preview

dastarruer:
Use shorter quote for theme preview, as discussed in #5152.

image


dastarruer:
Can either use 'glitters' or 'glisters.' Not sure if we're going for accuracy or not, so I used 'glisters.'

Issue #5196, opened 2026-07-18
FR: Remove buttons with duplicate purpose


dastarruer:
### What problem would this feature solve?

Currently, there are multiple buttons that do the exact same thing:

- Exiting a book to go to the library has two buttons: <img width="39" height="31" alt="Image" src="https://github.com/user-attachments/assets/c31f32ab-616d-4697-ac3d-8f15e93c781a" /> and <img width="30" height="34" alt="Image" src="https://github.com/user-attachments/assets/084391fd-c777-4175-a531-128a768b4855" />
- Accessing the notebook has two buttons: <img width="60" height="39" alt="Image" src="https://github.com/user-attachments/assets/fb854c7c-a4d3-4b04-b6bb-97ae2bd74736" /> and <img width="27" height="28" alt="Image" src="https://github.com/user-attachments/assets/0b69076c-ac52-4c2c-8655-dfae18450728" />
- Font and layout technically has just one button, but if you count the hamburger menu it's two: <img width="43" height="38" alt="Image" src="https://github.com/user-attachments/assets/d310b498-d7c3-4e9f-b555-dedb720739c0" /> and <img width="29" height="31" alt="Image" src="https://github.com/user-attachments/assets/4e441313-e825-487a-8347-19b55333ee9b" />

All these buttons only serve to clutter the UI and confuse first-time users.

### Describe the solution you'd like.

1. Remove the X button to go back to library in favor of the aptly-named 'Go to Library' button
2. Remove the 'Notebook' button in the top bar in favor for the 'Annotate' button at the bottom of the sidebar.
3. Remove the 'Aa' button and keep the button in the hamburder menu (though this I'm not so sure about; likely a better solution)

chrox:
> Remove the X button to go back to library in favor of the aptly-named 'Go to Library' button

The X button will also handle closing one book in parallel reading. It cannot be deleted.

> Remove the 'Notebook' button in the top bar in favor for the 'Annotate' button at the bottom of the sidebar.

They actually navigate to different pages, the feather pen icon is for writing notes and the bottom notes icon is for browsing notes. Although we plan to unify the two panels (see https://github.com/readest/readest/issues/3870), but before that we should not break existing UI, right?

> Remove the 'Aa' button and keep the button in the hamburger menu (though this I'm not so sure about; likely a better solution)

This is intentional for better discoverability of the settings.

dastarruer:
[...]

PR #5229, opened 2026-07-20
WIP: merge notebook with sidebar annotations tab 

### LINKED ISSUE #3870: Logic/UI Suggestion: Unifying "Notes" and "Annotations / highlights" for a smoother experience
First, I want to say thank you for building such an incredible program. Readest is truly amazing, and the KOReader sync is a game-changer for my workflow. The app is already so polished!

I have a suggestion to streamline the "Annotate" view logic to make it even more intuitive and efficient.

**The Logic**
Currently, the "Annotate" pane treats Highlights and Annotations (Highlight + Note) as distinct behaviors, but they could be unified for a faster UI:

The Definition: Typically, an Annotation is simply a Highlight + a Note. They are part of the same object.
The Current Issue: When clicking a pure Highlight, the app takes me to the text (Great!). But when clicking an Annotation, it opens a separate note viewer on the right. To actually get to the book text, I have to click a second time in that new pane.

**The Suggestion: A "Unified" Annotate Pane**
To save system resources and reduce UI "jumping" (especially in large books with hundreds of notes), we could treat them the same in the left pane:

**1. Keep it in the Sidebar:** Allow us to view or edit the Note text directly within the left "Annotate" pane, right under the highlight.
**2. One-Click Navigation:** Clicking any item in the left list (Highlight or Annotation) should take the user directly to the reading location in the book immediately.
**3. Eliminate or allow disabling the notebook from the Right Pane:** This would remove the need for the right-side note viewer to pop open and closed, making the experience feel much tighter and more stable.

**Why this helps:**

Efficiency: For those of us with 800+ highlights (like my current ePub!), it prevents the right pane from constantly "flashing" open and closed while scrolling.
Consistency: It treats "Annotations" as what they are—highlights with text attached—keeping everything in one place.
Thank you again for all your hard work on this amazing tool. I hope this suggestion helps make the Readest UI feel even more "pro" and streamlined!

dastarruer:
Closes #3870.

The idea is to hopefully get rid of the right sidebar entirely, as its function can just as easily be replaced with the existing left one. Unifying the right sidebar with the left would also reduce UI clutter in the reader ribbon.

### DIFF
--- a/apps/readest-app/src/app/reader/components/sidebar/BooknoteView.tsx
+++ b/apps/readest-app/src/app/reader/components/sidebar/BooknoteView.tsx
@@ -317,20 +317,14 @@ const BooknoteView: React.FC<{
           ref={osRootRef}
           data-overlayscrollbars-initialize=''
           style={{ height: containerHeight }}
-          className='flex flex-col'
+          className='flex !flex-col' // TODO: Get rid of !flex-col (which is the only solution that's worked so far)
         >
-          <div
-            className={clsx('search-bar', {
-              'search-bar-visible': true,
-            })}
-          >
-            <SearchBar
-              isVisible={true}
-              bookKey={sideBarBookKey}
-              searchTerm=''
-              onSearchResultChange={setSearchResults}
-            />
-          </div>
+          <SearchBar
+            isVisible={true}
+            bookKey={sideBarBookKey}
+            searchTerm=''
+            onSearchResultChange={setSearchResults}
+          />
           <div className='min-h-0 flex-1'>
             <Virtuoso
               ref={virtuosoRef}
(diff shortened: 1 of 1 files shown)


chrox:
> The idea is to hopefully get rid of the right sidebar entirely, as its function can just as easily be replaced with the existing left one.

This does not align with Readest’s design direction, the reader will still be a 3-column layout. Please leave #3870 unchanged.

dastarruer:
> > The idea is to hopefully get rid of the right sidebar entirely, as its function can just as easily be replaced with the existing left one.
> 
> This does not align with Readest’s design direction, the reader will still be a 3-column layout. Please leave #3870 unchanged.

Still, like I said in #5196, the notebook and annotations tab are almost exactly the same for no real reason. Perhaps annotations in the left sidebar should be removed entirely, as a read-only view of all highlights likely isn’t something a user really needs?

dastarruer:
I'll close this since I don't have much interest in working on it right now. Maybe I'll revisit it in the future, idk.

PR #5242, opened 2026-07-21
feat: style checked toggle

dastarruer:
Before:

image

After:

image

Just makes it more obvious at a glance when a toggle is checked.

### DIFF
--- a/apps/readest-app/src/components/primitives/toggle.tsx
+++ b/apps/readest-app/src/components/primitives/toggle.tsx
@@ -4,7 +4,12 @@ const Toggle = React.forwardRef<
   React.ComponentRef<'input'>,
   React.ComponentPropsWithoutRef<'input'>
 >(({ className, ...props }, ref) => (
-  <input ref={ref} type='checkbox' className='toggle' {...props} />
+  <input
+    ref={ref}
+    type='checkbox'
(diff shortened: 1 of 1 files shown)


chrox:
Please also test it in the e-ink mode.

dastarruer:
> Please also test it in the e-ink mode.

image

Is this expected behavior?

chrox:
You will need to switch to the 'Contrast' theme color and light theme mode and enable the 'Eink Mode' in Settings - Behavior - Device to actually test the eink mode.

dastarruer:
> You will need to switch to the 'Contrast' theme color and light theme mode and enable the 'Eink Mode' in Settings - Behavior - Device to actually test the eink mode.

image

Still not sure if this is intended behavior. Is the toggle supposed to lose its bg color when eink is enabled?

chrox:
> Is the toggle supposed to lose its bg color when eink is enabled?

Yes, that is intentional. In E Ink mode, we should ideally use only pure black and white, avoiding intermediate grayscale values. E Ink devices typically support just 16 levels of gray, and some can be difficult to distinguish. A pure black-and-white palette provides the clearest and safest experience.

dastarruer:
Then this should be ready to go.

PR #5245, opened 2026-07-21
feat: add setting to enable skeuomorphic book covers

### LINKED ISSUE #4905: FR: Make book covers look like actual book covers

dastarruer:
Closes #4905.

Changes old book spine gradient to look like a hardcover book, adds a `Library` section to the theme panel with a new 'Skeuomorphic Covers' toggle. 

Also renames `apps/readest-app/src/components/settings/color` to `apps/readest-app/src/components/settings/theme` to match #5042.

# Screenshots

image

image


dastarruer:
It's not as clean looking as the apple books version. In fact it looks quite cheap, so if you could generate a better looking gradient somehow that would be very much appreciated.

Also not sure why the book covers are rounded.

dastarruer:
image

After testing on my own library, it looks good for some books but not really for all (specifically talking about the first two; the gradient just looks off on them).

Issue #5222, opened 2026-07-19
FR: Show all import options from the Bookshelf add button


bcrave:
### What problem would this feature solve?

The “+” button in the Library header opens a menu containing the import methods available on the current platform, such as importing from a local file or browsing the Online Library.

However, the “+” tile displayed at the end of the Bookshelf grid opens the local file picker directly. The “Import Books” button shown when the library is empty behaves the same way.

This inconsistency makes the Online Library and other supported import methods less discoverable. Two controls that appear to perform the same “add books” action currently produce different behavior.

### Describe the solution you'd like.

I would like the Bookshelf “+” tile and the empty-state “Import Books” button to open an “Import Books” modal.

The modal would:

- Display the same platform-dependent import options as the Library header.

- Include “From Local File” and “Online Library”/“OPDS Catalogs.”

- Include “From Directory” and “From Web URL” on platforms where those actions are supported.

- Present the options as large, fully clickable cards based on the visual design of the “Popular Catalogs” cards in the OPDS dialog.

- Include a short description beneath each option.

- Close before launching the selected file picker, dialog, or Online Library.

The existing Library header dropdown would remain visually unchanged. Both presentations would share the same option definitions so their available actions remain consistent.

### Describe workarounds you may have used to acheive the same results.

There aren't really any workarounds for this sort of problem. There are simply two UI components that appear to have the same purpose but function differently.

### Additional Context

I have a working implementation available and would be happy to open a pull request if this approach is approved.

PR #5247, opened 2026-07-22
Show import options when clicking the bookshelf add button

bcrave:
## Summary

The Bookshelf "+" tile and the empty-library "Import Books" button opened the local file picker directly, so the Online Library, feeds and the other platform-specific import methods were only discoverable from the Library header. Both entry points now open the same `ImportMenu` that the header "+" button uses.

This addresses #5222.

### Reuse the header import menu as an anchored popup

- Added `ImportMenuPopup`, which renders the existing `ImportMenu` as a fixed overlay anchored to the button that opened it.
- The "+" tile lives inside the virtualized bookshelf scroller, whose `overflow-y: scroll` would clip an absolutely positioned `dropdown-content`, so the menu is positioned from the anchor's bounding rect instead.
- The menu is centered below the anchor, flips above it when there is no room below, and is clamped to the viewport minus safe-area insets.
- `getMenuPosition` is a pure function so the placement rules are unit tested.
- Dismisses on backdrop click and on Escape, like the header dropdown.

### Keep a single source of import options

- `ImportMenu` is unchanged apart from an optional `menuClassName`, which follows the existing `Dropdown` convention and lets the popup drop the header-specific triangle and top margin.
- The header dropdown and both bookshelf entry points now render the same component, so the available options and their platform gates cannot drift apart.

### Update the Bookshelf import entry points

- The Bookshelf "+" tile and the empty-library "Import Books" button report the clicked element as the popup anchor.
- `page.tsx` holds the anchor state and renders one popup, reusing the file, directory, URL, feed and catalog handlers with the same platform gates the Library header already uses.

### Tests

- Added `import-menu-popup.test.tsx`: the always-available options, the platform-dependent Directory and Web URL options, the Online Library / OPDS label, action + dismiss, backdrop dismiss, and the placement math including safe-area insets.
- Updated the Playwright `LibraryPage.importBook()` helper, since the empty-state button now opens the menu instead of the file picker directly.
- Added a web e2e test that the empty-library button opens the import menu.

## Testing

- [x] `pnpm test` - 7815 passed, 7 skipped
- [x] `pnpm lint` (tsgo + Biome) and `pnpm format:check`
- [x] `pnpm test:e2e:web` - 15 passed
- [x] Manual web check: the menu opens under the "+" tile, flips above and clamps to the edge when the tile sits at the bottom of the scroller without being clipped, the backdrop dismisses it, and "Online Library" opens the catalog dialog.

### DIFF
--- a/apps/readest-app/e2e/pages/LibraryPage.ts
+++ b/apps/readest-app/e2e/pages/LibraryPage.ts
@@ -11,6 +11,9 @@ export class LibraryPage extends BasePage {
   readonly searchInput: Locator;
   readonly clearSearchButton: Locator;
   readonly emptyState: Locator;
+  readonly emptyStateImportButton: Locator;
+  readonly importMenu: Locator;
+  readonly localFileImportItem: Locator;
 
   constructor(page: Page) {
     super(page);
@@ -20,6 +23,11 @@ export class LibraryPage extends BasePage {
     this.searchInput = page.locator('.search-input');
     this.clearSearchButton = page.locator('[aria-label="Clear Search"]');
     this.emptyState = page.getByRole('heading', { name: 'Start your library' });
+    this.emptyStateImportButton = page.locator('.hero').getByRole('button', {
+      name: 'Import Books',
+    });
+    this.importMenu = page.locator('.menu-container');
+    this.localFileImportItem = this.importMenu.getByRole('menuitem', { name: 'From Local File' });
   }
 
   async goto(): Promise<void> {
@@ -37,16 +45,19 @@ export class LibraryPage extends BasePage {
   }
 
   /**
-   * Import a book file via the empty-state "Import Books" button.
+   * Import a book file via the empty-state "Import Books" button, which opens
+   * the same import menu as the library header "+" button.
    *
    * The file `<input>` is created off-DOM (see `useFileSelector.selectFileWeb`),
    * so a `filechooser` event must be awaited rather than locating an
    * `<input type="file">`.
    */
   async importBook(filePath: string): Promise<void> {
-    const importButton = this.page.locator('.hero').getByRole('button', { name: 'Import Books' });
+    await this.emptyStateImportButton.click();
+    await this.localFileImportItem.waitFor({ state: 'visible' });
+
     const chooserPromise = this.page.waitForEvent('filechooser');
-    await importButton.click();
+    await this.localFileImportItem.click();
     const chooser = await chooserPromise;
     await chooser.setFiles(filePath);
   }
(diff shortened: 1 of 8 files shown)


chrox:
Please also attach some screenshots for the UI part.

bcrave:
image

chrox:
I suggest reusing the existing import menu by displaying it as a popup from the + button rather than adding a new modal dialog.

bcrave:
Ok, so you want me to update the bookshelf + tile to render the existing ImportMenu inside the same Dropdown component used by LibraryHeader, instead of maintaining a separate modal? Should the empty-library “Import Books” button use the same popup as well, or should this change apply only to the bookshelf tile?

chloeroform:
In terms of UX, if we expose the same options with the `+` tile and the `+` menu, do we really need the `+` tile? It is functionally redundant for the user and takes visual space. 🙂

dastarruer:
> In terms of UX, if we expose the same options with the `+` tile and the `+` menu, do we really need the `+` tile? It is functionally redundant for the user and takes visual space. 🙂

Agreed, the `+` tile also requires the user to scroll all the way down to the bottom of their library, which, in a big library, is easy to miss. 

But also, the `+` menu is located in the search bar, which makes its function ambiguous unless you hover over/click on it. It could either be moved to the right of the search bar next to the `...` menu, or as a floating button in the bottom right corner (though this would look out of place on desktop; better to keep this mobile-only). Though I think this would be out of the scope of this PR.

bcrave:
I also agree about the redundancy, and also that it probably needs to be its own issue. I think in this case the redundancy is ok so long as the behavior is consistent. It gives the user multiple options to add a book depending on how large the library is.  Another option could be to put the + tile at the beginning of the Bookshelf?  So the user doesn't have to scroll to find it with a large library.

chrox:
> Another option could be to put the + tile at the beginning of the Bookshelf? So the user doesn't have to scroll to find it with a large library.

Not really an option. To be honest, having the + tile permanently at the beginning of the bookshelf would make the bookshelf look so ugly that I would never open it again. Isn't it?

dastarruer:
> > Another option could be to put the + tile at the beginning of the Bookshelf? So the user doesn't have to scroll to find it with a large library.
> 
> Not really an option. To be honest, having the + tile permanently at the beginning of the bookshelf would make the bookshelf look so ugly that I would never open it again. Isn't it?

Well it wouldn’t be that ugly… but I agree, it would at least look out of place. Either way, this discussion should probably be in a separate issue, not here.

chloeroform:
In terms of pure UX, I think the best solution would be to:

1. Remove the `+` tile.
2. Make sure the top bar is always visible when you scroll.

chrox:
> Well it wouldn’t be that ugly… but I agree, it would at least look out of place.

This is merely an example of how aesthetic tastes vary from person to person. Taste is subjective rather than factual, and persuading someone to adopt a different aesthetic preference can be difficult, if not impossible. In fact, convincing someone to accept even an objective fact can be challenging. That is why people often rely on belief and trust.

PR #5252, opened 2026-07-22
fix: apply percentage-only KOSync progress

masonc15:
## Summary

- accept remote KOSync progress when it contains either a non-empty position string or a usable percentage
- let percentage-only responses reach the existing `goToFraction` fallback
- preserve precise XPointer navigation whenever a position string is available
- add helper-level and rendered-hook regression coverage

## Root cause

Some KOSync-compatible servers cannot translate a source reader locator into a KOReader XPointer and return a valid percentage with a null progress string. Grimmory is one such KOSync-compatible server. `KOSyncClient` already accepted this response shape, and `applyRemoteProgress` already had a `goToFraction` fallback, but `pullProgress` returned early whenever the progress string was absent, so that fallback was unreachable.

The implementation remains server-agnostic: it checks only whether the response has a usable position or fraction. Existing upload behavior, timestamp comparison, conflict strategies, and XPointer conversion are unchanged.

## User impact

Readers can now apply newer percentage-only remote progress instead of silently remaining at the current position. Invalid fractions such as zero, negative values, `NaN`, and infinity remain unusable.

## Validation

- Full unit suite: `pnpm exec dotenv -e .env -e .env.test.local -- vitest run` — 589 files passed, 3 skipped; 7,722 tests passed, 7 skipped.
- Focused KOSync suite: direct one-shot Vitest run over `kosyncProgress`, `kosyncConflict`, `kosyncRemoteFraction`, `useKOSync`, and `KOSyncClient` — 5 files passed, 37 tests passed.
- `pnpm exec biome lint` on the four touched files — passed.
- `pnpm exec tsgo --noEmit` — passed.
- `git diff --check origin/main...HEAD` — passed.

### DIFF
--- a/apps/readest-app/src/__tests__/hooks/kosyncProgress.test.ts
+++ b/apps/readest-app/src/__tests__/hooks/kosyncProgress.test.ts
@@ -1,5 +1,9 @@
 import { describe, it, expect } from 'vitest';
-import { isXPointerProgress, getRemoteFraction } from '@/app/reader/hooks/kosyncProgress';
+import {
+  getRemoteFraction,
+  hasUsableRemoteProgress,
+  isXPointerProgress,
+} from '@/app/reader/hooks/kosyncProgress';
 
 describe('isXPointerProgress', () => {
   it('accepts CREngine XPointers', () => {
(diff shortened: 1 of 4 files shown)


chrox:
> Some KOSync-compatible servers cannot translate a source reader locator into a KOReader XPointer and return a valid percentage with a null progress string. Grimmory is one such KOSync-compatible server.

If a sync server cannot return a KOReader XPointer as the progress location, it does not fully implement the KOSync protocol. Readest uses [the xcfi module](https://github.com/readest/readest/blob/main/apps/readest-app/src/utils/xcfi.ts) to convert between standard EPUB CFI locations and KOReader XPointers. Since the module is licensed under AGPL-3.0, it should be compatible with Grimmory.
I would suggest implementing this conversion on Grimmory’s side, or otherwise clarifying that Grimmory is not fully KOSync-compatible so users are not given misleading expectations. If Grimmory’s built-in sync service has a stable API, Readest could also support that API directly instead of relying on Grimmory’s incomplete KOSync endpoint.

Percentage-only progress is not sufficient and not a good option because there is no shared standard for mapping a percentage to a paragraph, sentence, page, or section. Different readers calculate progress differently. For example, Readest treats an EPUB as a continuous character stream and calculates paragraph positions and percentages from character offsets. This makes the percentage independent of the device and font size. KOReader, by contrast, renders the book on the current device, assigns content to pages, and then calculates progress from those rendered pages. Its percentage therefore depends on factors such as screen size and font size. As a result, the same percentage may point to different paragraphs, or even different chapters, in KOReader and Readest.

chrox:
This will not work as expected, so I’m closing this for now. I’d be happy to help investigate further once an issue has been filed with Grimmory.

PR #5256, opened 2026-07-22
fix(translate): restore Yandex Translate provider

Columpio:
## Summary

I originally added the Yandex Translate provider to Readest. The relay service it depended on is no longer available, so this PR repairs my original integration and brings the provider back without relying on that relay.

Yandex Translate is especially useful for Slavic languages, where in my experience it often produces better translations than the other available providers. I would like to restore it as an option for Readest users.

## What changed

- Replaced the unavailable `translate.toil.cc` relay with Yandex's translation endpoints.
- Added session creation and reuse, including one retry when Yandex reports an invalid session.
- Added a same-origin Next.js proxy for web builds because browsers cannot set the `Referer` header required by Yandex. Tauri builds call Yandex directly.
- Split long text into requests below Yandex's actual size limit while preserving sentence, whitespace, and grapheme boundaries where possible.
- Kept automatic source-language detection and normalized Chinese locale variants to the language code accepted by the endpoint.
- Added request concurrency limits, timeouts, cancellation support, response validation, and basic abuse protection for the web proxy.
- Re-enabled Yandex in the provider registry.

## Testing

- Yandex provider and translator registry tests: 43 passed.
- Yandex web proxy tests: 18 passed.

### DIFF
--- a/apps/readest-app/src/app/api/yandex-translate/route.ts
+++ b/apps/readest-app/src/app/api/yandex-translate/route.ts
@@ -1,0 +1,129 @@
+﻿import { NextRequest, NextResponse } from 'next/server';
+import {
+  YANDEX_ORIGIN,
+  YANDEX_SESSION_URL,
+  YANDEX_TRANSLATE_URL,
+  YANDEX_USER_AGENT,
+} from '@/services/translators/providers/yandexShared';
+
+/**
+ * Same-origin proxy for the Yandex Translate web API, used by the `yandex`
+ * translation provider in web builds. The Yandex session endpoint requires a
+ * translate.yandex.ru Referer, which a browser cannot send cross-origin, so
+ * the client calls this route and the headers are attached server-side.
+ * Only two fixed upstream endpoints are exposed — no arbitrary URL is ever
+ * fetched, so there is no SSRF surface.
+ */
+const ENDPOINTS: Record<string, string> = {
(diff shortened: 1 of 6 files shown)


apps/readest-app/src/app/api/yandex-translate/route.ts
```diff
@@ -0,0 +1,129 @@
...
+// Response constructor throws when these statuses carry a body.
+const NULL_BODY_STATUSES = new Set([204, 205, 304]);
+
+const getClientId = (request: NextRequest) =>
+  request.headers.get('cf-connecting-ip') ??
+  request.headers.get('x-real-ip') ??
+  request.headers.get('x-forwarded-for')?.split(',')[0]?.trim() ??
+  'unknown';
+
+export async function POST(request: NextRequest) {
```
chrox, line 41: Please refer to readest/apps/readest-app/src/app/api/tts/edge/route.ts and add this
```
const { user, token } = await validateUserAndToken(request.headers.get('authorization'));
  if (!user || !token) {
    return NextResponse.json({ error: 'Not authenticated' }, { status: 403 });
  }
```

so that this proxy api won't be misused and we can add rate limit to specific users.

chrox (changes requested):

chrox (approved):

PR #5280, opened 2026-07-23
Hide upload/download buttons when signed out

chloeroform:
Simplify the library view for signed-out users by hiding the upload and download cloud icons displayed in the bottom-right corner of each book.

These actions are unavailable without signing in, and currently only redirect users to the sign-in page. Since sign-in is already accessible via the swipe-down gesture and the Settings menu, the extra buttons add unnecessary clutter.

### DIFF
--- a/apps/readest-app/src/app/library/components/BookItem.tsx
+++ b/apps/readest-app/src/app/library/components/BookItem.tsx
@@ -10,12 +10,10 @@ import {
 import { Book } from '@/types/book';
 import { useEnv } from '@/context/EnvContext';
 import { useAuth } from '@/context/AuthContext';
-import { useRouter } from 'next/navigation';
 import { useTranslation } from '@/hooks/useTranslation';
 import { useSettingsStore } from '@/store/settingsStore';
 import { useResponsiveSize } from '@/hooks/useResponsiveSize';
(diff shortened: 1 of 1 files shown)


54wedge:
Custom css is all you need. https://github.com/readest/readest/pull/2095#issuecomment-3315633204

chloeroform:
Thanks for the context. We'll see what @chrox says.

I understand the

> This is meant to encourage users to log in with a Readest account thus I cannot merge this PR.

But do we need to remind the user for _every_ book on the home library?

I hope we can revisit this decision.

chrox:
The target users are expected to sign in with an account. If the cloud icons are hidden while signed out, they may not be able to tell whether a book has been uploaded or whether their account is not signed in properly.

That said, users can still override this behavior with custom UI CSS to hide the cloud-sync icons from the bookshelf:

```
.show-cloud-button {
  opacity: 0;
}

.group:hover .show-cloud-button {
  opacity: 1;
}
```

See https://discord.com/channels/1314226120886976544/1418764497807609856

chloeroform:
There has to be a better way to do this... I just can't think of one right now though, so I agree to reject this PR.

I may come back with a different approach later, though. Brace yourself. 🙂

PR #5282, opened 2026-07-23
feat(reader): add collapsible chapter sections to search results

WantenMN:
## Summary

Add collapsible chapter sections to the search results sidebar, making it easier to navigate through long search result lists. Each chapter group now has a sticky header with a toggle, so readers can collapse irrelevant chapters and focus on the ones they care about.

### Collapsible chapter sections

Extracted a `ChapterSection` component from the previously flat inline JSX. Each section header:

- Shows the **chapter label** and **match count** badge
- Uses a **sticky header** (`position: sticky; top: 0`) so it remains visible while scrolling through results
- Uses the same **SVG triangle expander icon** as the table of contents (TOC) chapter list for visual consistency

demo

### DIFF
--- a/apps/readest-app/src/__tests__/components/SearchBar.test.tsx
+++ b/apps/readest-app/src/__tests__/components/SearchBar.test.tsx
@@ -1,0 +1,97 @@
+import { act, cleanup, render } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import SearchBar from '@/app/reader/components/sidebar/SearchBar';
+
+const mocks = vi.hoisted(() => ({
+  // getProgress returns null until the book emits its first relocate event.
+  progress: null as { section: { current: number } } | null,
+  search: vi.fn(),
+}));
+
+vi.mock('@/context/EnvContext', () => ({
+  useEnv: () => ({ envConfig: {}, appService: null }),
+}));
+
+vi.mock('@/store/settingsStore', () => ({
+  useSettingsStore: () => ({ settings: {} }),
+}));
+
+vi.mock('@/store/bookDataStore', () => ({
+  useBookDataStore: () => ({
+    getBookData: () => ({ book: { primaryLanguage: 'en' } }),
+    getConfig: () => ({ searchConfig: { scope: 'section', mode: 'text' } }),
+    setConfig: vi.fn(),
+    saveConfig: vi.fn(),
+  }),
+}));
+
+vi.mock('@/store/readerStore', () => ({
+  useReaderStore: () => ({
+    getView: () => ({ search: mocks.search, clearSearch: vi.fn() }),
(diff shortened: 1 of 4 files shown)


dastarruer:
The scroll bar seems to slightly overlap the chapter dropdown. Otherwise, nice feature!

chrox:
Thanks for this! Nice feature. I reviewed it, rebased onto current `main`, and pushed two follow-up commits directly to the branch (hope you don't mind) rather than leaving a long list of review nits. Summary of what changed and why:

### `fix(reader): correct RTL, e-ink and collapse-state issues in search sections`

- **Collapse state leaked across searches.** `handleSearch` doesn't clear results up front, so the previous results stay mounted until the first new hit arrives. Sections are keyed by `index + label`, so the first chapter of the next query reused the key and came back already collapsed for results the reader had never seen. Collapse state now lives in `SearchResults` and resets when `searchProgress` drops to 0.
- **RTL.** `ml-2` on the match count violates DESIGN.md 2.8 (the sidebar sets `dir='rtl'` for RTL books). Now `ms-2`.
- **E-ink.** The sidebar renders `bg-base-100` under e-ink (`SideBar.tsx:190`), so the hardcoded `bg-base-200` header showed up as a grey band. Now `not-eink:bg-base-200 eink:bg-base-100`, matching the convention used in `NavigationBar`/`ColorPanel`. Hover and the expander transition are suppressed on e-ink too.
- **Scroll container lookup.** `closest('.overflow-y-auto')` couples the logic to a utility class; scoped it to `.search-results`. The rest of the sidebar scrolls via OverlayScrollbars, so if search results ever adopt it, the utility-class lookup would silently no-op.
- **A11y.** `role='button'` on the `<h3>` strips the heading role that readers had before. It's now a real `<button>` inside the `<h3>`, which keeps heading semantics and gets Enter/Space plus a `focus-visible` ring for free, so the hand-rolled `onKeyDown` could go.
- **`line-clamp-1`** on the header was inert: Tailwind emits `display` after `line-clamp`, so `flex` wins the cascade, and the inner span already truncates.

### `fix(reader): stop search from crashing before the first relocate event`

Pre-existing, unrelated to your change, but easy to hit while testing this one. `getProgress` returns `BookProgress | null`, but `SearchBar` asserted it non-null and destructured `section` off it, so searching right after opening a book threw:

```
TypeError: Cannot destructure property 'section' of 'progress' as it is null.
```

The value is only needed to scope a section-scoped search, so it's now read optionally and falls back to searching the whole book.

### Tests

Added `SearchResults.test.tsx` and `SearchBar.test.tsx` (5 tests), each written against the unfixed code first to confirm they reproduce. Full suite green: 7735 passed, plus lint and format.

Your sticky-header scroll restoration works nicely, by the way. I verified in the browser that collapsing a stuck header keeps it pinned with no scroll jump.

PR #5283, opened 2026-07-23
Load OPDS catalogs when opening the Integrations panel

chloeroform:
Opening the Integrations panel for the first time always reports 0 OPDS catalogs, even when catalogs are already configured, because the catalog list has not yet been loaded.

This change makes sure we load the catalog list before computing the count so the displayed value reflects the actual number of registered OPDS catalogs, and not zero.


chloeroform:
Any issue with this PR @chrox ?

(No hurry, just cleaning my todo list!)

chrox:
Please hold on. We usually process pull requests in FIFO order and issues in LIFO order.

Issue #3183, opened 2026-02-06
FR: option to disable internal brightness override in mobile


Yoav6:
**Does your feature request involve difficulty for you to complete a task? Please describe.**

The mobile app has its own brightness slider which overrides the phone brightness. What inevitably happens is that either I turn on the app at night and get blinded because last time I used it was in daylight and full brightness, or I turn on the app at day and can't see anything because last time I used it was at night on minimum brightness. I already change the brightness of my phone throughout the day, I don't need another brightness slider to change. 

Personally I don't see the point of this feature, but assuming others do find it useful, I would at least like the ability to disable it.

chrox:
Yes, there is already one option to always use system screen brightness in Settings - Behavior - Device - Auto Screen Brightness. But if you slide the brightness bar in the bottom navigation bar in Readest this option will be unchecked and you need to check it to use system screen brightness.

Yoav6:
Ok, nice, found it and turned it on. But this isn't very intuitive. I would change the setting's name to something like "Use system screen brightness", and also add a toggle or a notice in the slider UI.

forteller:
I agree with @Yoav6: I saw that setting, but had no idea before finding this issue that "auto screen brightness" meant use the system brightness. I thought it meant that it would automatically adjust brightness based on your surroundings, like you can select in the system also. Would be very helpful if this feature was renamed!

Thank you!

PR #5292, opened 2026-07-23
fix(reader): keep system brightness on after swipe

alexander-pecheny:
> Disclaimer: the code is fully agent-written. The PR description is fully human-written (`→`s courtesy of [Ilya Birman's typography layout](https://ilyabirman.net/typography-layout/)).

I had the same problem as described in #3183:

- turn on "follow system brightness"
- adjust brightness in readest during the day, system brightness toggle silently turns off
- pick up a phone to read at night → unpleasantly surprised by bright screen

Workaround as proposed in #3183:

- keep "follow system brightness" off, only modify brightness via system controls

That works, but is not very convenient. This PR makes it so swipe-to-adjust-brightness still works, but if you close the app and open it again, you get current system brightness.

# Review

Agent self-reviewed via mattpocock's `code-review` skill. I then read the whole PR myself, changed several things. In its current state it looks reasonable to me. (But I don't know neither Swift nor Kotlin, so take that with a grain of salt.)

# Testing

My agent tested in Simulator via `axe` CLI and in android emulator. Then I booted dev app on my physical iPhone and confirmed the flow:

1. Set system brightness low, open Readest → brightness is low.
2. Swipe brightness up to medium, close Readest → brightness is back to system low.
3. Reopen Readest → back at system low.

I didn't test on a physical Android.

Also, this PR doesn't break any tests (apparently there are some pre-existing broken tests).

### DIFF
--- a/apps/readest-app/src-tauri/plugins/tauri-plugin-native-bridge/android/src/main/java/NativeBridgePlugin.kt
+++ b/apps/readest-app/src-tauri/plugins/tauri-plugin-native-bridge/android/src/main/java/NativeBridgePlugin.kt
@@ -2,7 +2,9 @@ package com.readest.native_bridge
 
 import android.Manifest
 import android.app.Activity
+import android.app.Application
 import android.app.PendingIntent
+import android.os.Bundle
 import android.content.ComponentName
 import android.content.ContentValues
 import android.content.Context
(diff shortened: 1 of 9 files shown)


dastarruer:
Why is there a threshold? It might make more sense to use system brightness whenever system brightness differs from readest brightness, as system brightness should override readest.

Also thanks for implementing this, this issue’s been plaguing me for the past few weeks (too lazy to open an issue), so it’s nice to see a fix for it.

alexander-pecheny:
> Why is there a threshold? It might make more sense to use system brightness whenever system brightness differs from readest brightness, as system brightness should override readest.

I was thinking about a case where e.g. user is reading in the dark using the dark theme, and concurrently (switching in and out) uses an app with lighter background – it would make sense for them to have slightly more brightness in Readest than in the other app. But maybe that is an overcomplication and we can just drop it

alexander-pecheny:
@dastarruer dropped it, now system brightness always wins

PR #5302, opened 2026-07-24
fix(windows): enforce LF line endings and normalize test path separators

LukashevychAndrii:
1. .gitattributes - enforce LF line endings

On Windows, Git's core.autocrlf converts LF to CRLF on checkout.
This broke serve.mjs: the shebang line became #!/usr/bin/env node\r, and esbuild
choked on the trailing \r. Prevents similar issues across the entire project.

2. useScreenWakeLock-scope.test.ts - normalize path separators

path.relative() returns backslashes on Windows, but the test asserted forward
slashes. Normalized on collection.

### DIFF
--- a/apps/readest-app/scripts/nightly-verify-harness/serve.mjs
+++ b/apps/readest-app/scripts/nightly-verify-harness/serve.mjs
@@ -1,4 +1,3 @@
-#!/usr/bin/env node
 // Local nightly-updater verification harness (Tier 2 detection + Tier 4 invoke).
 //
 // `pnpm verify:nightly` serves crafted nightly/stable manifests + a test artifact
--- a/apps/readest-app/src/__tests__/hooks/useScreenWakeLock-scope.test.ts
+++ b/apps/readest-app/src/__tests__/hooks/useScreenWakeLock-scope.test.ts
@@ -25,7 +25,7 @@ describe('screen wake lock scope', () => {
   const callSites = walk(SRC)
     .filter((path) => !path.endsWith('hooks/useScreenWakeLock.ts'))
     .filter((path) => /useScreenWakeLock\(/.test(readFileSync(path, 'utf8')))
-    .map((path) => relative(SRC, path));
+    .map((path) => relative(SRC, path).replace(/\\/g, '/'));
 
   it('is acquired while reading', () => {
     expect(callSites).toContain('app/reader/components/Reader.tsx');


.gitattributes
```diff
@@ -0,0 +1,44 @@
+* text=auto eol=lf
```
chrox: Why is this actually needed? I have never seen this .gitattributes in any popular open-source projects.

Also the binary file types seem not exhaustive at all. This does not look right.

chrox (changes requested):

chrox:
It might be more proper to enforces LF line endings in your editor as the default for this project since this is not a Windows project for sure.

.gitattributes
LukashevychAndrii: Not gonna do the "but *some big project* does it" thing - that's not a real argument.  text=auto without eol=lf still lets git convert to CRLF on windows checkout. That's exactly how serve.mjs got the stray \r that broke esbuild. The eol=lf is the part that actually fixes it. Went with a global rule + binary exceptions instead of listing every text format - just cleaner. Lmk if I missed any binary format actually used in the project, pretty sure i got them all

LukashevychAndrii:
> It might be more proper to enforces LF line endings in your editor as the default for this project since this is not a Windows project for sure.

how is editor settings "more proper"? I didn't open an editor. I cloned, changed nothing, ran pnpm test, and got failures. That's git corrupting files on checkout - editor has nothing to do with it. You're shipping tests that fail out of the box on windows

LukashevychAndrii:
> It might be more proper to enforces LF line endings in your editor as the default for this project since this is not a Windows project for sure.

Also, been chasing that TTS freeze bug where it randomly locks up. If you don't want that fix either, I'd rather know now than keep sinking time into it

chrox:
Is it possible to configure your git to use LF line endings? I also build and test on Windows 10 and never encounter this issue so I believe it's a problem of the environment settings.

chrox:
If we have to update the configuration every time we introduce a new file type, the workflow becomes fragile. There should be a more robust solution.

LukashevychAndrii:
> If we have to update the configuration every time we introduce a new file type, the workflow becomes fragile. There should be a more robust solution.

That's why I went with * text=auto eol=lf - it covers all text files automatically, nothing to update when new file types are added. The binary list only changes when you add a new binary format to the project. If the global rule bothers you *.mjs text eol=lf is the minimal fix, but that leaves the footgun

chrox:
I'm fine with configuration to the text files, but it still surprises me when the git changes every LF line endings by default for even binary files. I'm not familiar with Windows, but is this true behavior?

LukashevychAndrii:
> I'm fine with configuration to the text files, but it still surprises me when the git changes every LF line endings by default for even binary files. I'm not familiar with Windows, but is this true behavior?

core.autocrlf=true is meant to normalize line endings, but it's breaking the build: git adds \r on checkout, esbuild strips only the shebang (leaving \r behind), and the parse fails.

about binaries: git detects binary fine on its own, the list is just good practice, i collected all formats that are being used currently

chrox:
Then we should fix the serve.mjs. The shebang is not cross-platform so we should remove it.

LukashevychAndrii:
> Then we should fix the serve.mjs. The shebang is not cross-platform so we should remove it.

Agreed, that's cleaner

chrox (approved):

PR #5305, opened 2026-07-24
feat(theme): unify theme selector and bg texture selector styling

### LINKED ISSUE #5289: Inconsistent styling between theme picker and bg image picker
Image

While the background image places its radio button in the top right corner and uses a blue button, the theme picker places it in the center and uses a white button. Personally, I think both sections should use white buttons, but the theme picker should be styled more in line with the minimal look of the background picker.

dastarruer:
Closes #5289.

# Screenshots

image

image

image

Inspiration taken from apple books


chrox:
It's much better now. Thanks.

PR #5328, opened 2026-07-25
[WIP] feat: add mangadex integration

### LINKED ISSUE #5221: FR: Download manga from mangadex

Currently in order to read manga from mangadex in readest, this has to be done with a third-party tool, and then imported into readest. In addition, so far, the tools that I've found do not allow you to download by volume, and instead require you to download by chapter.

A feature that downloads manga using the [official mangadex api](https://api.mangadex.org/docs/) in the CBZ format, and saves it to the user's library.

_No response_

### Additional Context

This is not against their terms of service, as the api exposes downloads directly. However, they may not organize chapters by volume, so if this is the case, perhaps the user can select multiple chapters to be downloaded and merged into one file.

dastarruer:
Closes #5221.

The idea here is to add a `/mangadex` route that lets the user download manga to their library in the `.cbz` format via the [MangaDex API](https://api.mangadex.org/docs/).

Here is a list of features I hope to implement:

# Roadmap

## Core features

- [ ] Search MangaDex for manga
- [ ] View manga details and available chapters
- [ ] Download chapters as `.cbz` files
  - [ ] Choose between original and reduced quality (supported natively by the mangadex api)
  - [ ] Download a single chapter
  - [ ] Download multiple chapters in bulk
    - [ ] Save chapters as separate books or a single book
  - [ ] Download chapters by volume when possible

## Possible future improvements

While these features would be nice, whether I actually go through with adding them is another thing. These features would basically turn readest into a mangadex client, which is *way* out of the scope of my abilities. That being said:

- [ ] MangaDex account integration (prerequisite for the next two features)
- [ ] Sync manga reading status with MangaDex
- [ ] Manga updates feed (perhaps with added push notifications as well)
- [ ] Resize manga to user's device screen resolution (would reduce file sizes)
  - [ ] If sync is enabled, the option to choose which device to resize/optimize for

(Personally feel that resizing manga would be nice, just because ~40 chapters can result in a 400mb file in my experience, which, given the 500mb sync limit would be very limiting.)

### DIFF
--- a/apps/readest-app/src/__tests__/services/mangadex/search.test.ts
+++ b/apps/readest-app/src/__tests__/services/mangadex/search.test.ts
@@ -1,0 +1,115 @@
+import { searchManga } from '@/services/mangadex/search';
+import { describe, it, expect, vi } from 'vitest';
+
+describe('search', () => {
+  const fakeResponse = {
+    result: 'ok',
+    response: 'collection',
+    data: [
+      {
+        id: 'fcb5111f-be88-4f5a-b456-22135f3eda49',
+        type: 'manga',
+        attributes: {
+          title: { en: 'ONE PIECE STRONG WORLD' },
+          altTitles: [{ en: 'One Piece: Strong World' }, { en: 'One PIece Volume 0' }],
+          description: {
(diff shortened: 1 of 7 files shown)


dastarruer:
Given mangadex's dubious legality, supporting mangadex integration may not be the best idea in the long term, so I'll be closing this.

PR #5350, opened 2026-07-26
Fix/OpenAI tts playback gap

nikitastaf1996:


nikitastaf1996:
Oops. Ai agent messed up. I was going for my fork. Messing around

PR #5351, opened 2026-07-26
feat(popup): restyle popups

### LINKED ISSUE #5267: Popup backgrounds aren't themed with the rest of the app
Self explanatory, actually:

Image
<br>
Image

dastarruer:
Closes #5267.

# Screenshots

## Dark mode

image

image

image

## Light mode

image

image

image


dastarruer:
More screenshots!

# Dark mode

image

image

image

# Light mode

(Just realized I need to style the highlight type buttons, but here's a screenshot nonetheless):
image

image

image

dastarruer:
How do I update the tests so checks pass? (Pre-push hook passed on my machine so I didn't realize my mistake)

chrox:
Remove the PNG files that serve as the reference images for the highlight rendering tests. Running pnpm test:browser on your local machine will regenerate them. Once you commit the regenerated images, the CI will use them as the new baseline for rendering comparisons.

chrox:
> # Light mode
> 
> 
> 
> (Just realized I need to style the highlight type buttons, but here's a screenshot nonetheless):
> 
> image
> 

Yes, it should be restyled with theme colors too.

dastarruer:
image

Also realized that these toggles look reversed, so I'll need to fix that as well before this can be merged.

apps/readest-app/src/components/Popup.tsx
```diff
@@ -128,22 +128,10 @@ const Popup = ({
   const outerTriangleStyles = getTriangleStyles(trianglePosition, 7, 0);
   const innerTriangleStyles = getTriangleStyles(trianglePosition, 7, -1);
 
-  const popupHeight = height ?? childrenHeight;
-  const triangleHidden = !!(
```
chrox, line 138: If this is removed, triangle might be drawn at a place where no text is selected, especially when the popup is higher than half of the screen and the selected text is at the center of the screen, it could be reproduced with the dictionary popup on smaller screens.

chrox (changes requested):

apps/readest-app/src/components/Popup.tsx
dastarruer: https://github.com/user-attachments/assets/bf075b17-84c6-459b-95e6-417156ed33d7

Can't seem to replicate it. Is this what you meant?

chrox: This can be reproduced with a wider but lower window so that the popup is a modal instead of an action sheet.

dastarruer: I can't reproduce it at all:

image

If you can, please send a screenshot so I know what you're talking about.

chrox: image


image

image


Note that for the 'floated' word, the triangle is hidden otherwise it will be drawn at a wrong place.

dastarruer: Still can't reproduce it. But the reason I removed it was to fix a visual bug:

image

I'll see if I can't work around it, but I don't really know why this is happening.

chrox: The triangle color might be hardcoded to `gray-500`.

dastarruer: Doesn't seem to be the case anywhere. The effect I'm trying to get (which you can see in the screenshots) only seems to work once both inner and outer triangles are set to `z-50`. 

Again, can't seem to reproduce the bug. Maybe instead of hiding the triangle all popups could get a fixed height, so the triangle never needs to be hidden? Either way will keep working at it.

chrox: > Again, can't seem to reproduce the bug. Maybe instead of hiding the triangle all popups could get a fixed height, so the triangle never needs to be hidden? Either way will keep working at it.

There are still cases where the popup is more than half the height of the window. When the selected text is near the center of the window, the popup’s pointer triangle can therefore be misplaced. It will be reproducible with a higher popup.

dastarruer:
> Yes, it should be restyled with theme colors too.

I took the liberty of redoing the buttons instead:

image

Mostly taken from gdocs:

image

chrox:
image

Please keep the original highlight style, it was much better than this.

image

This may be an existing issue, but the shape should be a perfect circle.

dastarruer:
> image
> Please keep the original highlight style, it was much better than this.

The original style had poor contrast:

image

Since the user can choose literally any color, choosing a text color that properly contrasts is difficult. Hence, I took the approach that gdocs does, and just made it a colored underline.

> image
> This may be an existing issue, but the shape should be a perfect circle.

Yes, I’ll fix this. It was an issue introduced by this pr.

chrox:
For users who speak languages other than English, the U and A icons may not clearly communicate “underline” and “highlight,” so they are not entirely self-explanatory.

Since the highlighter icon also does not necessarily need to preview the currently selected font and highlight colors, we could instead use a more recognizable combination: the most common yellow marker color as the background, paired with a high-contrast font color such as black or dark gray.

dastarruer:
> For users who speak languages other than English, the U and A icons may not clearly communicate “underline” and “highlight,” so they are not entirely self-explanatory.

The 'A' icon is just an icon; it doesn't mean anything in English, and so doesn't need to be translated. Though if this is still an issue, perhaps using a highlighter pen icon instead would be more universal. 'U' on the other hand stands for underline, so perhaps translating it into other languages (e.g. 'S' for 'subrayar' in Spanish) is necessary. I also chose 'U' since it seems to be standard across all text editors (see word, gdocs, libreoffice, etc.)

> Since the highlighter icon also does not necessarily need to preview the currently selected font and highlight colors, we could instead use a more recognizable combination: the most common yellow marker color as the background, paired with a high-contrast font color such as black or dark gray.

Fair, yes. Though if you'd like me to use a highlighter pen icon instead, the colored underline should stay. Let me know and I'll make the necessary changes.

chrox:
> 'U' on the other hand stands for underline, so perhaps translating it into other languages (e.g. 'S' for 'subrayar' in Spanish) is necessary.

It's practically impossible to translate it for most languages.

For the highlight options it will be better to keep all the original icon and color design including the color and style preview and only fix the contrast issue that is really necessary. I find the original design is much better and intuitive than your redesign. And Readest is not an editor and there is no reason to follow UI design from editors.

apps/readest-app/src/components/Popup.tsx
dastarruer: The issue seems to be that `triangleHidden` incorrectly evaluates to `true` whenever direction is `'down'`. So the issue lies with the faulty check.

dastarruer:
Highlight options now look like this:

image

chrox:
Looks much better now.

dastarruer:
Thanks very much!

> image
> Also realized that these toggles look reversed, so I'll need to fix that as well before this can be merged.

I just need to fix this first and then this should be ready to be merged.

dastarruer:
> Remove the PNG files that serve as the reference images for the highlight rendering tests. Running pnpm test:browser on your local machine will regenerate them. Once you commit the regenerated images, the CI will use them as the new baseline for rendering comparisons.

I've removed all the screenshots, but running `pnpm test:browser` and `pnpm test:browser -- --update` does not update anything. The tests simply keep failing.

chrox (approved):

chrox:
I pushed a follow-up commit and rebased the branch onto current `main` (your changes came across unmodified). CI is green now. Thanks for sticking with this — the direction is good, these were mostly last-mile issues.

**On the screenshots: you were right, and my earlier advice was wrong.** Deleting the PNGs and re-running genuinely could not work. `resolveScreenshotPath` in `vitest.browser.config.mts` returned a *relative* path — reads happened to resolve it against the cwd, but writes go through Vite's `server.fs` access check, which rejects relative paths and reports it as the unhelpful `Couldn't write file to fs`. `--update` hit the same wall and failed silently. I made the path absolute, so `pnpm test:browser <file> --update` now regenerates baselines, and I've committed the regenerated references.

The substantive fixes:

- **Removed the `not-eink:drop-shadow-xl` wrapper.** This was the one real bug. A non-none `filter` makes that element a containing block *and* a stacking context for absolutely positioned descendants, so the popup was being re-anchored away from the viewport coordinates it is given, and its `z-50` was demoted below reader chrome rendered after `<Annotator>` (FooterBar, BooknotesNav, SearchResultsNav, FootnotePopup). The triangle shadow now lives on the triangle element itself, which keeps the look without the side effects. Added `popup-ancestor-filter.browser.test.tsx` to lock it down.
- **Reverted the `isPointInRect` padding to the default.** This is the `triangleHidden` behaviour you were chasing: the real cause was the popup being measured by content box instead of border box, fixed in #5431, which is now in `main` under this branch. With that in place the 7px padding only weakens the guard.
- **Restored e-ink affordances** — `eink-bordered` on the style buttons, the global toggle and the proofread input, plus the B&W e-ink branch for the marker swatch, which otherwise renders yellow as a mid-grey blob.
- **Gave the marker glyph a fixed dark ink.** The swatch is always yellow now, but the `A` was inheriting `base-content`, so it was white-on-yellow in dark themes — the same contrast problem you originally raised, just moved.
- `text-foreground` maps to `var(--foreground)`, which is not defined anywhere in the repo, so it silently resolved to `inherit`; switched to `text-base-content`.
- Translator footer moved into flow but the grid still declared three rows, and the padding reserved for the old absolutely positioned footer was left behind.
- Dropped a leftover `console.log` in `Popup.tsx`, and de-duplicated the repeated toggle classes.

Two changes worth pushing back on if you disagree:

1. I replaced the `isDarkMode ? … : …` class ternaries with a new `theme-dark:` Tailwind variant (`html[data-theme$="-dark"] &`), so the surfaces resolve in CSS rather than from the theme store. Behaviour is identical; `popup-theme-surface.browser.test.tsx` asserts the resolved colours so a broken variant can't pass silently.
2. I toned `shadow-2xl` down to `shadow-sm` on the small round highlight buttons and the colour strip. That is a visual change from the screenshots above and is baked into the new baselines — happy to put it back.

PR #5354, opened 2026-07-26
Allow Tauri app to connect to a custom self-hosted instance

jeankhawand:
Tauri desktop/mobile users can now point the app at their own Readest
   docker stack without rebuilding. The self-hosted Supabase URL, anon key,
   and site URL are saved to both the settings store and localStorage so
   the supabase client (initialized at module-eval time) picks them up on
   the next app start.

   Key changes:
   - runtimeConfig.ts: add getSelfHostedLocalConfig() + SELF_HOSTED_STORAGE_KEY
     for reading the localStorage bridge at module-eval time
   - supabase.ts: read localStorage self-hosted config as a fallback before
     env-var defaults, enabling runtime URL switching without a rebuild
   - environment.ts: getBaseUrl() falls back to selfHosted.apiBaseUrl so API
     calls route to the custom instance
   - settings.ts: add SelfHostedConfig interface and selfHosted? on SystemSettings
   - SelfHostedForm.tsx: new settings sub-page (Supabase URL, Anon Key, Site URL)
     with Save / Clear actions; Clear removes the localStorage key so the next
     restart reconnects to Readest Cloud
   - IntegrationsPanel.tsx: Advanced section with Self-hosted Instance
     on Tauri or whenever selfHosted is already configured (so it can always be cleared)
   - user/page.tsx: hide UsageStats, PlansComparison, and cloud account actions
     when self-hosted is active; show "Self-hosted" plan badge instead
   - SettingsMenu.tsx: hide the storage/translation quota row when self-hosted
   - library/page.tsx: force redirect to /auth when self-hosted is configured and
     no session exists, using the existing keepLogin path
   - runtimeConfig.test.ts: unit tests for getSelfHostedLocalConfig covering
     happy path, missing key, malformed JSON, and thrown storage errors

### DIFF
--- a/apps/readest-app/src/__tests__/services/runtimeConfig.test.ts
+++ b/apps/readest-app/src/__tests__/services/runtimeConfig.test.ts
@@ -1,0 +1,69 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
+const localStorageMock = (() => {
+  let store: Record<string, string> = {};
+  return {
+    getItem: (key: string) => store[key] ?? null,
+    setItem: (key: string, value: string) => {
+      store[key] = value;
+    },
+    removeItem: (key: string) => {
+      delete store[key];
+    },
+    clear: () => {
+      store = {};
+    },
+  };
+})();
+
+Object.defineProperty(globalThis, 'localStorage', { value: localStorageMock });
+
+import {
+  getSelfHostedLocalConfig,
+  SELF_HOSTED_STORAGE_KEY,
+  type SelfHostedLocalConfig,
+} from '@/services/runtimeConfig';
+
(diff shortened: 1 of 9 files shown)


chrox:
> Tauri desktop/mobile users can now point the app at their own Readest
> docker stack without rebuilding.

We currently assume that users who self-host the Readest Docker stack will build the desktop or mobile applications themselves. This ensures they understand how the service API is configured and used.

Exposing the self-hosted service interface to all users is premature. It is not aligned with the current product roadmap, and the Readest team does not yet have the resources to handle the support issues that would arise once this feature is broadly exposed. Supporting those deployments would become our responsibility, so we should defer this until we are prepared to maintain it properly.

PR #5355, opened 2026-07-26
feat(tts): add End of Chapter option to sleep timer

nurumubu:
## Problem

Read Aloud (TTS) currently has no way to stop at a chapter boundary - when the current chapter/section ends, playback silently continues into the next one. For long books, that means there's no natural checkpoint to pause at (as opposed to only being able to pause mid-sentence anywhere).

This was raised before as a side note in #4408 ("it would be nice to have a settings toggle... to stop at the end of a chapter if you were going to fall asleep and didn't want to backtrack too far"), but that issue was about a different bug and got closed once that bug was fixed - the chapter-pause idea itself was never implemented.

## Solution

Adds an "End of Chapter" option to the existing Sleep Timer picker (in the TTS player sheet), alongside the numeric duration options. Selecting it:

- Lets the current chapter finish normally.
- Stops right at the chapter boundary without tearing down the player (unlike the numeric timer firing, or a normal full stop) - the panel stays visible, paused, with the position already advanced to the start of the next chapter.
- Is a standing preference, not a one-shot: pressing Play again reads the next chapter, and it stops again at the end of that one, and so on - until the user picks a different Sleep Timer option. This matches the actual use case (chapter-by-chapter listening, e.g. falling asleep) better than a one-shot timer would.
- Persists correctly across the reader's stop/resume cycle (each chapter's playback uses a freshly-created TTSController, so the preference is re-applied via TTSSessionManager on every claim).

## Implementation
- TTSSessionManager.ts: new TTS_STOP_AT_CHAPTER_END sentinel value, setStopAtChapterEnd() / getStopAtChapterEnd(), mutually exclusive with the numeric sleep timer.
- TTSController.ts: new stopAtChapterEnd flag, checked at both places where the controller silently auto-advances into the next section (the SSML-streaming path in #speak(), and the mark-by-mark path in forward()/#handleNavigationWithoutSSML used by e.g. Web Speech) - so it isn't engine-specific. When set, advances into the next section (via the existing #initTTSForNextSection()) but leaves it unspoken and sets forward-paused (not plain paused - the play/pause toggle treats those two differently: paused calls the lightweight ttsClient.resume() meant for a genuinely mid-utterance-suspended engine, while forward-paused correctly calls start(), which is what's needed here since nothing was spoken yet).
- useTTSControl.ts / TTSPlayerSheet.tsx: wire the new option into the existing timeout picker and button caption.

## Testing

Manually tested via pnpm dev-web, reading through a multi-chapter test EPUB chapter by chapter with "End of Chapter" enabled:

- Web Speech (browser TTS): confirmed stopping precisely at each chapter boundary, panel staying visible, and Play correctly starting the next chapter with audio (not just an inert "playing" icon).
- Edge TTS: confirmed working the same way when testing in an actual Edge browser (Chrome/plain-web testing hits an unrelated 403 on the Edge TTS websocket handshake, since the Origin-spoofing edge TTS relies on isn't available to page JS in a normal browser context - not something this PR touches).

Have not been able to build/test the Tauri iOS/desktop targets directly (no Mac/Xcode available), but the change is entirely within the shared TS/React TTSController/TTSSessionManager layer used across all platforms, not anything platform-specific.

## Known limitation / open question for review

Manually pressing the "skip to next chapter" button (>>, i.e. forward() with byMark=false) while "End of Chapter" is enabled and actively playing will also stop instead of skipping ahead, since it shares the same auto-advance code path as natural end-of-chapter continuation. Happy to special-case that distinction (e.g. threading through an "explicit user navigation" flag) if that's considered worth the extra complexity - wanted to flag it rather than silently leave it.


nurumubu:
One more caveat worth flagging while this is in review: for PDFs, "End of Chapter" actually stops per-page rather than per-chapter. `book.sections` in `pdf.js` is generated one section per PDF page (`Array.from({ length: pdf.numPages })`), not grouped by any chapter/outline structure - so `stopAtChapterEnd`, which triggers on section boundaries, ends up pausing after every single page for PDFs instead of at real chapter breaks. It still "works" in the sense that it stops and resumes correctly, just much more frequently than intended for that format. Happy to look into scoping this to EPUB specifically (or deriving real chapter boundaries from the PDF outline/bookmarks if present) if that's considered worth doing before merge - wanted to flag it either way.

test_mock_fix.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/apps/readest-app/src/__tests__/hooks/useTTSControl.test.tsx b/apps/readest-app/src/__tests__/hooks/useTTSControl.test.tsx
```
chrox: Why patching this instead of modifying it?

chrox (changes requested):

chrox:
For now it's fine for the PDFs. We will handle the TOC hierarchy for PDFs in a dedicated PR.

chrox:
Pushed three commits to this branch (and rebased it onto current `main`, so the earlier SHAs were rewritten — sorry for the force-push, `git fetch && git reset --hard nurumubu/feat/tts-stop-at-chapter-end` if you have it checked out locally).

**`chore: remove stray test_mock_fix.patch`** — the leftover I asked about; its contents were already applied in `test(tts): mock`.

**`fix(tts): let user skips cross the chapter-end stop`** — this is the limitation you flagged, and it turned out to be wider than the `>>` button. The gate in `#handleNavigationWithoutSSML` sees the speak loop's own continuation and a deliberate user skip identically (both arrive with `isForward && isPlaying`), so with the mode armed it also swallowed next-sentence, and the media session's `nexttrack` / `seekforward` at `ttsMediaBridge.ts:290,294`. Reaching for skip on the lock screen and getting playback stopped instead is the case I'd most want to avoid.

The fix is smaller than the flag-threading you were braced for: gate in `forward()` behind an `isAutoAdvance` argument, passed only by the four calls `#speak()` makes to continue on its own. `#handleNavigationWithoutSSML` goes back to its original two-argument signature, which also reverts the `backward()` churn. Your `#stopAtChapterBoundary` reasoning was right and is unchanged (I condensed the comment, keeping both load-bearing points).

**`chore(i18n): translate "End of Chapter"`** — the option had no locale entries, so every non-English user saw a bare English string next to translated siblings. Added by hand across all 33 locales rather than via `i18n:extract`, which would have swept in unrelated drift already sitting on `main`.

Tests: `tts-controller.test.ts` and `tts-session-manager.test.ts` already had harnesses for exactly this, so the behaviour is now pinned on both sides — auto-advance parks on the next section in `forward-paused`, user skips and `backward` still cross, the last chapter still terminates, the speak loop really does pass the flag, and the manager both propagates the mode to later claims and stays mutually exclusive with the numeric timer. Full suite, lint, and format all pass.

Four things I left alone because they're your call, not defects:

1. **The `#speak()` gate fires on "no SSML", which isn't strictly "chapter ended."** A paragraph emptied by `filterSSMLWithLang` (when `ttsTargetLang` is set) or by `preprocessCallback` lands in the same branch mid-chapter. The base code already jumps a section there, so it isn't newly introduced, but the mode makes it visible as a bogus chapter boundary. I kept the gate — post-fix it legitimately covers pressing Play on an already-exhausted section.
2. **Three lifetimes from one picker.** `stopActive()` clears the numeric timer but deliberately not `#stopAtChapterEnd`, which also carries into a different book — yet neither survives an app restart. So "30 minutes" is forgotten after a stop while "End of Chapter" isn't. Either clear it next to `#clearSleepTimer()`, or persist it as a real preference.
3. **Invisible outside the sheet.** `TTSMiniPlayer` renders its timer chip from `timeoutTimestamp`, always `0` here, so an armed chapter-end stop shows no indicator at all.
4. **The position advances at the stop.** `#initTTSForNextSection()` fires `onSectionChange`, so the reader visibly lands on the start of the next chapter — for the fall-asleep case that means waking up already scrolled past where the audio stopped. Reads intentional from the description; worth confirming against deferring the advance until Play.

chrox (approved):

test_mock_fix.patch
nurumubu: Whoops, my mistake - still getting the hang of the git workflow here, that patch file wasn't meant to be committed. Removed it, the actual changes are directly in `useTTSControl.test.tsx` now. Thanks for catching it!

nurumubu:
> Pushed three commits to this branch (and rebased it onto current `main`, so the earlier SHAs were rewritten — sorry for the force-push, `git fetch && git reset --hard nurumubu/feat/tts-stop-at-chapter-end` if you have it checked out locally).
> 
> **`chore: remove stray test_mock_fix.patch`** — the leftover I asked about; its contents were already applied in `test(tts): mock`.
> 
> **`fix(tts): let user skips cross the chapter-end stop`** — this is the limitation you flagged, and it turned out to be wider than the `>>` button. The gate in `#handleNavigationWithoutSSML` sees the speak loop's own continuation and a deliberate user skip identically (both arrive with `isForward && isPlaying`), so with the mode armed it also swallowed next-sentence, and the media session's `nexttrack` / `seekforward` at `ttsMediaBridge.ts:290,294`. Reaching for skip on the lock screen and getting playback stopped instead is the case I'd most want to avoid.
> 
> The fix is smaller than the flag-threading you were braced for: gate in `forward()` behind an `isAutoAdvance` argument, passed only by the four calls `#speak()` makes to continue on its own. `#handleNavigationWithoutSSML` goes back to its original two-argument signature, which also reverts the `backward()` churn. Your `#stopAtChapterBoundary` reasoning was right and is unchanged (I condensed the comment, keeping both load-bearing points).
> 
> **`chore(i18n): translate "End of Chapter"`** — the option had no locale entries, so every non-English user saw a bare English string next to translated siblings. Added by hand across all 33 locales rather than via `i18n:extract`, which would have swept in unrelated drift already sitting on `main`.
> 
> Tests: `tts-controller.test.ts` and `tts-session-manager.test.ts` already had harnesses for exactly this, so the behaviour is now pinned on both sides — auto-advance parks on the next section in `forward-paused`, user skips and `backward` still cross, the last chapter still terminates, the speak loop really does pass the flag, and the manager both propagates the mode to later claims and stays mutually exclusive with the numeric timer. Full suite, lint, and format all pass.
> 
> Four things I left alone because they're your call, not defects:
> 
>     1. **The `#speak()` gate fires on "no SSML", which isn't strictly "chapter ended."** A paragraph emptied by `filterSSMLWithLang` (when `ttsTargetLang` is set) or by `preprocessCallback` lands in the same branch mid-chapter. The base code already jumps a section there, so it isn't newly introduced, but the mode makes it visible as a bogus chapter boundary. I kept the gate — post-fix it legitimately covers pressing Play on an already-exhausted section.
> 
>     2. **Three lifetimes from one picker.** `stopActive()` clears the numeric timer but deliberately not `#stopAtChapterEnd`, which also carries into a different book — yet neither survives an app restart. So "30 minutes" is forgotten after a stop while "End of Chapter" isn't. Either clear it next to `#clearSleepTimer()`, or persist it as a real preference.
> 
>     3. **Invisible outside the sheet.** `TTSMiniPlayer` renders its timer chip from `timeoutTimestamp`, always `0` here, so an armed chapter-end stop shows no indicator at all.
> 
>     4. **The position advances at the stop.** `#initTTSForNextSection()` fires `onSectionChange`, so the reader visibly lands on the start of the next chapter — for the fall-asleep case that means waking up already scrolled past where the audio stopped. Reads intentional from the description; worth confirming against deferring the advance until Play.

Thanks for the thorough review and for fixing the manual-skip bug + handling the translations - really appreciate the depth here.

On #2 (timer asymmetry): I'd actually lean toward keeping this as-is rather than making it symmetric with the numeric timer. The numeric timer is a "stop in N minutes, I'm done for now" intent - it makes sense for it to reset after firing, since that specific countdown is spent. "End of Chapter" is a different kind of intent though: it's not "read one chapter and stop," it's "let me keep reading indefinitely and pull me back to a clean chapter boundary whenever I happen to doze off" - which could be chapter 3 tonight and chapter 11 two nights from now, possibly spanning into whatever I open next if I fall asleep mid-session and pick a different book the next day. Resetting it after every single chapter would mean re-enabling it constantly, which defeats the "set it and forget it before bed" use case it's meant for. So I think the current behavior (persists until explicitly changed, even across books) actually matches the intent better than symmetry would.

On #4 (position advancing immediately at stop): I think deferring the advance until Play is pressed would be better. Landing on the next chapter's start the moment it stops means if you check your phone before falling back asleep, the reading position already looks like it's ahead of what you actually heard - which is exactly the kind of "where was I" confusion this feature is trying to avoid in the first place. Advancing only on Play would keep the visible position matched to what was actually read.

chrox:
Thanks, the PR is already merged.

PR #5374, opened 2026-07-28
fix(translate): restore Yandex auto-detection and translated TTS

Columpio:
﻿## Summary

- fix Yandex Translate auto-detection by omitting `source_lang` instead of sending the unsupported `source_lang=auto`
- preserve the first translated sentence when TTS is configured to read translated text only
- add regression coverage for translated paragraphs with and without later sentence marks

### Yandex Translate auto-detection

The classic Yandex `tr.json` endpoint rejects `source_lang=auto` with HTTP 400:

```json
{"code":502,"message":"Invalid parameter: source_lang"}
```

Auto-detection works when `source_lang` is omitted entirely. The provider now leaves the parameter out for `AUTO` while continuing to send normalized explicit source languages.

### Translated-text TTS

Foliate can place the mark for the first translated sentence immediately before the translated `<lang>` block, while marks for later sentences remain inside the block. Filtering SSML to the target language removed the first mark, leaving initial translated text without an active mark. `parseSSMLMarks()` then omitted that first sentence, so playback started from the next one.

`filterSSMLWithLang()` now preserves the nearest preceding Foliate mark whenever pronounceable translated text appears before the first internal mark. Existing mark names are retained so playback highlighting continues to map back to Foliate ranges.

## TTS videos

Both videos use the **Read translated text only** setting.

### Before the fix

When playback moves to the next paragraph, TTS skips some of its sentences. The skipped text is visible on screen and can also be heard missing from the narration.

https://github.com/user-attachments/assets/2710fe71-7a95-4811-837a-578cbf95a6a7

### After the fix

With the same book and TTS settings, playback reads the next paragraph from its first sentence without skipping any translated text.

https://github.com/user-attachments/assets/4f1e37f2-1a8c-477c-9b91-e33a7773cc91

## Testing

- `apps/readest-app/src/__tests__/document/tts.test.ts`: 34 passed
- `apps/readest-app/src/__tests__/services/translators/providers.test.ts`: 44 passed
- reproduced the TTS regression under the IDE debugger before the fix and confirmed all three translated sentence marks after the fix
- verified Yandex AUTO translation against the live web endpoint: requests without `source_lang` returned HTTP 200 and rendered translated text

## Notes

`pnpm format:check` from the pre-push hook could not complete on Windows because Biome reported repository-wide CRLF differences in unrelated files. The changed files passed the staged Biome formatter and `git diff --check`.

### DIFF
--- a/apps/readest-app/src/__tests__/document/tts.test.ts
+++ b/apps/readest-app/src/__tests__/document/tts.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi } from 'vitest';
 import { textWalker } from 'foliate-js/text-walker.js';
 import { TTS } from 'foliate-js/tts.js';
 import { createRejectFilter } from '@/utils/node';
+import { filterSSMLWithLang, parseSSMLMarks } from '@/utils/ssml';
 
 const createHTMLDoc = (bodyHTML: string, attrs: Record<string, string> = {}): Document => {
   const parser = new DOMParser();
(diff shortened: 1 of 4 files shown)


Columpio:
@chrox you merge faster than I add PR description :D

I mean PoC before-after videos

PR #5392, opened 2026-07-29
fix(layout): keep dropdown menus within viewport

### LINKED ISSUE #5259: The drop-down menu isn't fully displayed
### Bug Description

The drop-down menu isn't fully displayed.
Ru language.
Samsung Galaxy A57.
The app version is 0.11.20.

Image
Image

### Steps to Reproduce

In main three dots menu
Three dots in Fonts menu

### Expected Behavior

Menu fully displayed

### Reproduction File

_No response_

### Operating System

Android

### OS Version

Android 16 UI 8.5

### Readest Version

0.11.20

### Custom CSS

_No response_

ChuwuYo:
Closes #5259

## Summary

- Keep the original CSS-anchored dropdown placement so the menu stays DOM-adjacent to its toggle and screen readers (TalkBack, VoiceOver) traverse it right after the button. An earlier revision of this PR repositioned menus with Floating UI through a portal at the end of `<body>`; on-device testing with TalkBack showed the portaled menu was effectively unreachable (entries at the end of the traversal order behind an unlabeled dismiss overlay), so the branch was updated in place with a CSS-anchored fix.
- Let long menu labels wrap with start alignment and dynamic row height instead of truncating.
- Give menus their natural `max-content` width capped to the viewport (`:where(.dropdown-content)` keeps explicit utility widths like `w-44` winning). Without this, enabling wrapping would let anchor-positioned menus collapse toward their longest word.
- Measure the open menu in `Dropdown` and shift it back into the viewport with a `translateX` clamp, re-applied on window resize and content resize. RTL is unaffected since the clamp is rect-based.
- Drop the per-menu inline `maxWidth`/`marginRight` viewport hacks in the reader view menu and quick action menu.
- Hide the pointer-only dismiss overlay from assistive tech and remove the `aria-hidden` `<summary>` from the Tab order.
- Suppress hydration warnings caused by browser extensions injecting attributes into the root `<html>` element before hydration.

## Verification

- New browser test `dropdown-viewport.browser.test.tsx` (real Chromium at 360px): long Russian labels wrap and stay within the viewport for start, center, and end placements near both screen edges; CJK menus keep natural width without wrapping; explicit consumer widths are preserved; RTL end alignment stays inside the viewport; and a regression guard asserts the menu stays DOM-adjacent to its toggle for assistive tech.
- Verified on a Xiaomi 13 (392px viewport) with TalkBack enabled: the first menu entry immediately follows the toggle in the accessibility tree, no unlabeled overlay node in the traversal path, no spontaneous dismissal, and both the library and reader view menus render fully inside the viewport.

## Before / After

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      Dropdown before the fix
    </td>
    <td>
      Dropdown after the fix
    </td>
  </tr>
</table>

### DIFF
--- a/apps/readest-app/src/__tests__/components/dropdown-viewport.browser.test.tsx
+++ b/apps/readest-app/src/__tests__/components/dropdown-viewport.browser.test.tsx
@@ -1,0 +1,175 @@
+/**
+ * Layout regression test for issue #5259 (drop-down menu not fully displayed).
+ *
+ * Renders the real Dropdown + Menu + MenuItem with Tailwind loaded and checks,
+ * in a real Chromium layout engine, that the CSS-anchored menus stay inside
+ * the viewport on a narrow phone screen:
+ *
+ * 1. Long translations (Russian): the menu is capped to the viewport width and
+ *    labels wrap to multiple lines so the full text stays readable.
+ * 2. Short labels (CJK): the menu keeps its natural content width and labels
+ *    do NOT wrap — guards the shrink-to-fit collapse where removing `nowrap`
+ *    let menus shrink to their longest word/character.
+ * 3. Every placement (start, center, end) is shifted back into the viewport
+ *    when its anchor sits near a screen edge, without portaling the menu out
+ *    of its DOM position next to the toggle (screen-reader traversal order).
+ * 4. Explicit consumer widths (w-44) and RTL alignment are preserved.
+ */
+
+import { describe, it, expect, vi, beforeAll, afterEach } from 'vitest';
+import { render, screen, cleanup, fireEvent, waitFor } from '@testing-library/react';
+import { page } from 'vitest/browser';
+
+import '@/styles/globals.css';
+
+vi.mock('@/hooks/useTranslation', () => ({
+  useTranslation: () => (s: string) => s,
+}));
+
+vi.mock('@/context/EnvContext', () => ({
+  useEnv: () => ({ envConfig: {}, appService: null }),
+}));
+
+import Dropdown from '@/components/Dropdown';
+import Menu from '@/components/Menu';
+import MenuItem from '@/components/MenuItem';
+import { DropdownProvider } from '@/context/DropdownContext';
+
+const PHONE_WIDTH = 360;
+const VIEWPORT_PADDING = 16;
+const RIGHT_ANCHOR_LEFT = 280;
+const LEFT_ANCHOR_LEFT = 8;
+
+const renderMenu = (
+  labels: string[],
+  {
+    placement = 'dropdown-end',
+    anchorLeft = RIGHT_ANCHOR_LEFT,
+    menuClassName = '',
+    direction = 'ltr',
+  }: {
+    placement?: 'dropdown-bottom' | 'dropdown-center' | 'dropdown-end';
+    anchorLeft?: number;
+    menuClassName?: string;
+    direction?: 'ltr' | 'rtl';
+  } = {},
+) => {
+  document.documentElement.classList.toggle('ui-rtl', direction === 'rtl');
+  const utils = render(
+    <DropdownProvider>
+      <div data-testid='dropdown-anchor' style={{ position: 'fixed', top: 8, left: anchorLeft }}>
+        <Dropdown
+          label='Test Menu'
+          className={`dropdown-bottom ${placement}`}
+          buttonClassName='btn btn-ghost h-8 min-h-8 w-8 p-0'
+          toggleButton={<span>⋯</span>}
+          showTooltip={false}
+        >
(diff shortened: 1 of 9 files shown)


chrox:
Thanks for this PR! The diagnosis of #5259 is spot on, and the Before/After looks great visually. I reviewed and tested it thoroughly, including on real devices, and unfortunately found that the portal approach breaks screen reader navigation, so I could not merge it as is.

Static checks all pass (ARIA attributes, axe scans, and the Blink accessibility tree are identical to main), but with TalkBack actually enabled on an Android device the menu becomes effectively unusable:

- Portaling the menu to the end of `<body>` moves its entries to the very end of the accessibility traversal order. From the toggle, a TalkBack user has to swipe past the settings button, every book on the shelf, and more before reaching the first entry. In the current CSS-anchored DOM they are one swipe away.
- The dismiss `Overlay` is exposed as an unlabeled, full-screen, clickable node sitting in the traversal path right before the entries; activating it closes the menu the user was trying to reach.
- With TalkBack running, the menu intermittently dismissed itself right after opening, and explore-by-touch taps on entries sometimes dismissed it instead of focusing them.

VoiceOver fails the same way for the same DOM-order reason. This class of issue is invisible to axe and to the CDP accessibility tree; it only shows up with a screen reader connected, which is why CI and the browser tests could not catch it.

I have opened #5410, which fixes #5259 while keeping the menu DOM-adjacent to its toggle: it adopts the label wrapping and dynamic row heights from this PR (thank you, credited there), caps the menu box to the viewport with `width: max-content` plus a `max-width`, and adds a small measure-and-shift clamp in `Dropdown` for placements near screen edges. It also ports a variant of your browser layout test, plus a regression guard asserting the menu stays DOM-adjacent for assistive tech.

The `suppressHydrationWarning` change is independent of the dropdown work and still worth landing. If you rescope this PR to just that commit, I would be happy to merge it.

chrox:
Following up on the review above: I updated this branch directly (maintainer edit) rather than splitting the work into a separate PR.

The branch now contains two commits on top of current main:

1. Your `suppressHydrationWarning` fix, unchanged (thank you!).
2. A CSS-anchored viewport fix for #5259 that replaces the Floating UI portal approach. It adopts this PR's label wrapping and dynamic row heights, caps menus to `width: max-content` / `max-width: calc(100vw - 32px)`, and adds a small measure-and-shift clamp in `Dropdown` for placements near screen edges. The menu stays DOM-adjacent to its toggle, which restores TalkBack and VoiceOver navigation, and the browser layout tests from this PR were ported to the new approach along with a DOM-adjacency regression guard.

Verified on a Xiaomi 13 with TalkBack enabled: the first entry is one swipe after the toggle, no unlabeled overlay in the traversal path, no spontaneous dismissal, and menus fit a 392px viewport. Full suites are green (8009 unit tests, 330 browser tests, lint).

Note: your fork's `main` was force-pushed as part of this update; if you have local work there, please rebase onto the new head before pushing.

ChuwuYo:
> Following up on the review above: I updated this branch directly (maintainer edit) rather than splitting the work into a separate PR.
> 
> The branch now contains two commits on top of current main:
> 
> 1. Your `suppressHydrationWarning` fix, unchanged (thank you!).
> 2. A CSS-anchored viewport fix for [The drop-down menu isn't fully displayed #5259](https://github.com/readest/readest/issues/5259) that replaces the Floating UI portal approach. It adopts this PR's label wrapping and dynamic row heights, caps menus to `width: max-content` / `max-width: calc(100vw - 32px)`, and adds a small measure-and-shift clamp in `Dropdown` for placements near screen edges. The menu stays DOM-adjacent to its toggle, which restores TalkBack and VoiceOver navigation, and the browser layout tests from this PR were ported to the new approach along with a DOM-adjacency regression guard.
> 
> Verified on a Xiaomi 13 with TalkBack enabled: the first entry is one swipe after the toggle, no unlabeled overlay in the traversal path, no spontaneous dismissal, and menus fit a 392px viewport. Full suites are green (8009 unit tests, 330 browser tests, lint).
> 
> Note: your fork's `main` was force-pushed as part of this update; if you have local work there, please rebase onto the new head before pushing.

Thank you for the thorough review, real-device accessibility testing, and for updating the branch directly. I had not considered how portaling the menu would affect real-world screen reader navigation with TalkBack and VoiceOver, especially since the static accessibility checks did not reveal the problem. Your detailed findings were very helpful.

The accessibility rationale for keeping the menu DOM-adjacent to its toggle makes complete sense. I also appreciate that you preserved and built on the label-wrapping and browser-test work from this PR.

I’ll sync my local branch with the updated remote head and make sure not to push the old portal-based implementation back. Thanks again for your work!

PR #5394, opened 2026-07-29
feat(theme): add Ambient Mode on Android

Jesusm1229:
## Summary

- Add a fourth theme mode, **Match Surroundings**, that switches light/dark from the Android ambient light sensor (same idea as Apple Books).
- Keep Auto / Light / Dark unchanged. The new mode only appears when a light sensor is available.
- Use lux hysteresis so the theme does not flicker near the threshold.
- Fall back to Auto if the sensor is missing at runtime.

Android-only for now. iOS / web / desktop have no usable public ambient-light API.

## Test plan

- [x] On an Android phone with a light sensor, open Settings → Theme Mode and confirm **Match Surroundings** appears as a fourth option
- [x] Select Match Surroundings, cover the front light sensor → page theme turns dark
- [x] Uncover / add light → page theme turns light
- [x] Near dim mid-light, theme should hold without rapid flicker
- [x] On web / desktop / iOS, confirm only Auto / Light / Dark are shown
- [x] `pnpm test` passes


dastarruer:
For other platforms, perhaps accessing the camera and taking its exposure value could emulate a light sensor. Otherwise, I like the feature. Good job!

PR #5410, opened 2026-07-31
fix(layout): keep CSS-anchored dropdown menus within viewport

### LINKED ISSUE #5259: The drop-down menu isn't fully displayed

chrox:
Closes #5259

## Summary

Fixes dropdown menus overflowing the viewport while keeping the original CSS-anchored placement, so the menu stays DOM-adjacent to its toggle and screen readers can navigate it.

An earlier approach (#5392) repositioned menus with Floating UI through a portal at the end of `<body>`. Static checks (ARIA attributes, axe scans, Blink accessibility tree) all passed, but on-device testing with TalkBack enabled showed real screen readers could not navigate the portaled menu:

- Menu entries landed at the very end of the accessibility traversal order, dozens of swipes away from the toggle, instead of right next to it.
- The dismiss overlay was exposed as an unlabeled full-screen clickable node sitting in the traversal path right before the entries, and activating it closed the menu.
- The menu intermittently dismissed itself right after opening when TalkBack's accessibility focus was on background content.

VoiceOver failed the same way for the same DOM-order reason.

## Changes

- Let menu labels wrap (`break-words`) and size library view-menu rows with `min-h` so long translations stay readable instead of truncating (adopted from #5392, credit to @ChuwuYo).
- Give `.dropdown-content` its natural `max-content` width capped to the viewport, with a zero-specificity `:where()` rule so explicit utility widths (`w-44`, `max-w-[90vw]`) still win. Without this, enabling wrapping would let anchor-positioned menus collapse toward their longest word.
- Measure the open menu in `Dropdown` and shift it back inside the viewport with a `translateX` clamp, re-clamped on window resize and content resize. RTL works unchanged since the clamp is rect-based.
- Drop the per-menu inline `maxWidth`/`marginRight` viewport hacks in the reader view menu and quick action menu.
- Hide the pointer-only dismiss `Overlay` from assistive tech (`aria-hidden`), and remove the `aria-hidden` `<summary>` from the Tab order (`tabIndex={-1}`).

## Verification

- New browser test `dropdown-viewport.browser.test.tsx` (real Chromium layout at 360px): long Russian labels wrap and stay within the viewport for start/center/end placements near both screen edges, CJK menus keep natural width without wrapping, explicit consumer widths are preserved, RTL end alignment stays inside the viewport, and the menu stays DOM-adjacent to its toggle as an accessibility regression guard.
- Verified on a Xiaomi 13 (392px viewport, TalkBack enabled): the accessibility tree now lists the first menu entry immediately after the toggle, the unlabeled overlay node is gone, the menu no longer auto-dismisses, and both the library view menu and the reader view menu render fully inside the viewport.
- `pnpm test` (8009 passed), `pnpm test:browser` (330 passed), `pnpm lint` all green.

### DIFF
--- a/apps/readest-app/src/__tests__/components/dropdown-viewport.browser.test.tsx
+++ b/apps/readest-app/src/__tests__/components/dropdown-viewport.browser.test.tsx
@@ -1,0 +1,175 @@
+/**
+ * Layout regression test for issue #5259 (drop-down menu not fully displayed).
+ *
+ * Renders the real Dropdown + Menu + MenuItem with Tailwind loaded and checks,
+ * in a real Chromium layout engine, that the CSS-anchored menus stay inside
+ * the viewport on a narrow phone screen:
+ *
+ * 1. Long translations (Russian): the menu is capped to the viewport width and
+ *    labels wrap to multiple lines so the full text stays readable.
+ * 2. Short labels (CJK): the menu keeps its natural content width and labels
+ *    do NOT wrap — guards the shrink-to-fit collapse where removing `nowrap`
+ *    let menus shrink to their longest word/character.
+ * 3. Every placement (start, center, end) is shifted back into the viewport
+ *    when its anchor sits near a screen edge, without portaling the menu out
+ *    of its DOM position next to the toggle (screen-reader traversal order).
+ * 4. Explicit consumer widths (w-44) and RTL alignment are preserved.
+ */
+
+import { describe, it, expect, vi, beforeAll, afterEach } from 'vitest';
+import { render, screen, cleanup, fireEvent, waitFor } from '@testing-library/react';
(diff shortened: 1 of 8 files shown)


chrox:
Superseded: #5392 was updated in place with this fix (maintainer edit), keeping the contributor's PR as the single change. Closing.

PR #5421, opened 2026-08-01
fix(proofread): match Unicode punctuation next to letters

### LINKED ISSUE #5257: Proofread rules cannot match Unicode punctuation adjacent to letters
### Bug Description

Proofread replacement rules do not seem to work for Unicode punctuation (e.g. Russian quotation marks `«»`) when those characters are adjacent to letters.
    
I originally wanted to remove these quotation marks before TTS because Apple Speech on recent iOS versions appears to pronounce them incorrectly in some voices.
    
However, this also affects normal text rendering. Even a simple replacement like:
    
```
Find: «
Replace: TEST
```
    
does not change the displayed book content.
    
The rules are saved correctly and remain enabled, but they are never applied.

### Steps to Reproduce

1. Open any EPUB containing Russian text such as:

`«Привет»`
    
2. Open **Proofread Replacement Rules**.
    
3. Create a rule:

```
Find: «  
Replace: TEST  
Regex: Off
```
    
4. Save the rule.
    
5. Reopen the book (or restart the reader).
    
6. Observe that the replacement is never applied.
    
The same happens for:

```
Find: »  
Replace: TEST
```

and for the regex:

`[«»]`

### Expected Behavior

The quotation mark should be replaced.
    
For example:

`«Привет»`

should become

`TESTПривет»`

or, if replacing both quotation marks,

`Привет`

### Reproduction File

_No response_

### Operating System

iOS/iPadOS

### OS Version

26.5.2

### Readest Version

0.11.20

### Custom CSS

_No response_

ChuwuYo:
Closes #5257

## Summary

Fixes proofread replacement rules that could not match Unicode punctuation when it appeared directly next to letters.

For example, literal rules for `«` or `»`, and regex rules such as `[«»]`, can now transform text such as `«Привет»`.

## Changes

- Determine the punctuation exemption from the actual matched text rather than only from the rule pattern.
- Skip Unicode word-boundary validation when the matched text consists entirely of punctuation.
- Preserve the existing boundary behavior for letters, numbers, underscores, combining marks, and symbols.
- Preserve the existing handling for scripts that do not normally use whitespace between words.
- Keep document structure and attributes such as `lang`, `dir`, and `aria-labelledby` unchanged.

## Behavior

The following cases are supported:

- Literal Unicode punctuation next to letters, such as `«Привет»`.
- Regex punctuation classes such as `[«»]`.
- Punctuation adjacent to Cyrillic, Arabic, and CJK text.
- Punctuation-only replacements in the TTS transformation pipeline.
- Mixed regex patterns containing connector punctuation such as `_`.

Existing behavior remains unchanged for:

- Whole-word matching of Unicode words.
- Combining marks.
- Currency and other Unicode symbols.
- Word-like connector punctuation.

## Scope

This change is limited to the shared proofread transformer. It does not modify rule storage, the proofread rule editor, document parsing, or TTS playback.

Visible book-content replacements were manually verified with Cyrillic, Arabic, and CJK text, including punctuation-only regex rules and boundary cases involving `_` and `€`.

The TTS-only transformer path is covered by an automated test using SSML content, but a complete manual end-to-end TTS-only verification was not performed.

The standard Proofread Replacement Rules settings do not provide a `TTS only` option when creating or editing a rule. Existing rules display their `TTS only` state, but that state cannot be changed from the regular rule editor.

A `TTS only` control is available through the EPUB text-selection popup, but it is disabled for the `Current selection` scope and becomes available only after switching to the book or library scope. The selection-based proofread action is also unavailable for non-EPUB content.

These are pre-existing UI limitations and are not changed by this PR. Therefore, this PR verifies the TTS-only transformation behavior at the transformer level, but does not claim manual validation of final speech output across platforms, voices, or TTS engines.

### DIFF
--- a/apps/readest-app/src/__tests__/utils/replacement.test.ts
+++ b/apps/readest-app/src/__tests__/utils/replacement.test.ts
@@ -1027,6 +1027,179 @@ describe('proofreadTransformer', () => {
       expect(result).not.toContain('café');
     });
 
+    test('should replace Unicode punctuation adjacent to letters', async () => {
+      const rules: ProofreadRule[] = [
+        {
+          id: 'opening-quote',
+          scope: 'book',
+          pattern: '«',
+          replacement: '',
+          enabled: true,
+          isRegex: false,
+          caseSensitive: true,
+          order: 1,
+          wholeWord: true,
+        },
+        {
+          id: 'closing-quote',
+          scope: 'book',
+          pattern: '»',
+          replacement: '',
+          enabled: true,
+          isRegex: false,
+          caseSensitive: true,
+          order: 2,
+          wholeWord: true,
+        },
+      ];
+      const ctx = createMockContext(
+        rules,
+        '<section aria-labelledby="greeting"><h2 id="greeting" lang="ru" dir="auto">«Привет»</h2></section>',
+      );
+      const result = await proofreadTransformer.transform(ctx, {
+        docType: 'application/xhtml+xml',
+      });
+      const doc = new DOMParser().parseFromString(result, 'application/xhtml+xml');
+      const section = doc.querySelector('section');
+      const heading = doc.querySelector('#greeting');
+
+      expect(result).toContain('Привет');
+      expect(result).not.toMatch(/[«»]/u);
+      expect(section?.getAttribute('aria-labelledby')).toBe('greeting');
+      expect(heading?.getAttribute('lang')).toBe('ru');
+      expect(heading?.getAttribute('dir')).toBe('auto');
+    });
+
+    test('should replace punctuation adjacent to letters across scripts', async () => {
+      const rules: ProofreadRule[] = [
+        {
+          id: 'punctuation-regex',
+          scope: 'book',
+          pattern: '[«»،「」]',
+          replacement: '',
+          enabled: true,
+          isRegex: true,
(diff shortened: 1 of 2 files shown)


ChuwuYo:
| Before | Proofread Settings | After |
|:---:|:---:|:---:|
| Before proofreading | Proofread settings | After proofreading |

ybbluesky:
被替换的文本希望能替换为空，即是删除部分文字，原程序好像不能替换为空值 @ChuwuYo

ChuwuYo:
> 被替换的文本希望能替换为空，即是删除部分文字，原程序好像不能替换为空值 @ChuwuYo

Thanks for the clarification. There are currently two different behaviors:

1. In the standard Proofread Replacement Rules manager, an empty replacement is already supported. The replacement field can be left blank to delete the matched text. This is how the empty replacement rules shown in the screenshots were created.

2. In the text-selection proofread popup, the Apply button is disabled when the replacement field is empty, so deletion cannot currently be submitted through that interface.

This PR only fixes Unicode punctuation matching next to letters and does not change either UI behavior.

Whether the text-selection popup should also support empty replacements, and whether that change belongs in this PR or should be handled separately, needs to be confirmed by the maintainer @chrox .

Issue #267, opened 2025-01-29
Windows font size


andrewchenk:
## Bug Description

Font size settings don't seem to do anything

**Example:**

![Image](https://github.com/user-attachments/assets/06ed8578-60f8-44f4-9db7-acbec72ae0f7)

## Expected behavior

The font size should change when I change its setting. I have override set to on 

## Technical inputs

Operating System:

```
Operating System: Windows 10
Readest Version: 0.9.7

```

chrox:
I cannot reproduce this on macOS Readest 0.9.7 and the web version. Does this issue occur for this book or also for other books?

andrewchenk:
I just tested it and it doesn't occur with another book I tried. I could maybe send it to you somehow? Interestingly enough though using the same epub Thorium (another epub reader) was able to change the font size.

chrox:
I will be very kind of you to send it to my email which is attached in my commit message.

chrox:
> EDIT: my guess is it isn't able to override font size if it's set in the epub itself? Changing the font family works.

Some epubs use weird css to hardcode the font size which might be not handled properly by readest to override it.

chrox:
Now with #270 it's easier to set the font size which might be the most frequently used setting. I made it override font-size by default so it should be effective to all books now without checking the override fonts option.

PR #5422, opened 2026-08-01
fix(reader): normalize body text size in reflowable books

### LINKED ISSUE #5420: Font size is not overriden in some books
### Bug Description

In some books (at least one) the font (throughout the entire book) is smaller than in other books

### Steps to Reproduce

1. Enable “Override Book Font”
2. Open the book from the link below

### Expected Behavior

Regular paragraphs have the font size defined in the  settings.

### Reproduction File

https://limewire.com/d/yQCZh#5vjYbAJJlU

### Operating System

Android

### OS Version

Android 16

### Readest Version

0.11.20

### Custom CSS

**Content:**

```css
::selection
{
	background-color: #0073e5;
	color: #f7f7f7;
}

a
{
	color: transparent !important;
	background-image: linear-gradient(to right, #6c7bdf, #56a4ff);
	background-clip: text;
}
```

**Reader UI:**

```css
button[aria-label="Translation Disabled"],
button[aria-label="Notebook"],
button[aria-label="Speak"],
.footer-bar button[aria-label="Font & Layout"]
{
	display: none;
}

#titlebar-close { background-color: #f7f7f7; }
#titlebar-close > svg > path { fill: #222; }

button[aria-label="Color"] { order: 3; }

.selection-buttons { justify-content: space-around; }

.lg\:tooltip[title="Highlight"],
.lg\:tooltip[title="Annotate"],
.lg\:tooltip[title="Search"],
.lg\:tooltip[title="Speak"],
.lg\:tooltip[title="Proofread"]
{
	display: none;
}

ChuwuYo:
Closes #5420

## Summary

Fixes reflowable books whose regular body text remains smaller than the configured default font size because the publication assigns explicit font sizes to its paragraphs.

The reproduction EPUB declares its regular paragraphs with `font-size: small`. Readest converts this publisher value to `0.875rem`, so changing the root font size still leaves the entire body text at 87.5% of the configured size.

## Design Rationale

In #267 and #270, Readest established that font size should always be controlled by the reader setting, without requiring “Override Book Font”. The existing implementation applies the configured size to `html` and `body`, while `overrideFont` only controls whether the publisher’s font family is replaced.

This PR preserves that design. It does not move font-size behavior behind the font-family override option. Instead, it completes the existing implementation by normalizing publisher-defined sizes on body-copy descendants that do not inherit the configured root size.

The implementation follows the body-copy element set used by Readium CSS font-size normalization:

- `p`
- `li`
- `div`
- `pre`
- `dd`

This is also consistent with the EPUB Reading Systems specification, which permits reading systems to override publication styling in response to user preferences, including more accessible reading settings.

References:

- Existing Readest behavior: #267 and #270
- [Readium CSS font-size normalization](https://github.com/readium/readium-css/blob/develop/css/src/modules/ReadiumCSS-fs_normalize.css)
- [CSS Fonts Level 4 absolute-size keywords](https://www.w3.org/TR/css-fonts-4/#absolute-size-mapping)
- [EPUB Reading Systems 3.4 styling overrides](https://www.w3.org/TR/epub-rs-34/#sec-css)

## Changes

- Mark rendered documents as reflowable or fixed-layout through the existing document-style lifecycle.
- Normalize body-copy elements in reflowable documents to the configured root font size.
- Continue enforcing the configured minimum font size through `--min-font-size`.
- Keep font-size normalization independent from the “Override Book Font” setting.
- Preserve the existing font-family override behavior.
- Exclude fixed-layout EPUBs, PDFs, and CBZ files from the new normalization.
- Avoid reintroducing document-level `zoom`, which was previously removed because of WebKit rendering and TTS synchronization issues.

## Behavior

For reflowable reading content:

- Regular body text follows the configured default font size.
- Publisher styles such as `p.indent { font-size: small; }` no longer make the entire book consistently smaller.
- The minimum font-size setting remains effective.
- Heading sizes and inline relative elements such as `<small>` retain their relative hierarchy.
- The behavior is independent of language, text direction, and writing mode.

For fixed-layout content:

- Publisher layout and font sizes remain unchanged.
- PDF, CBZ, and fixed-layout EPUB rendering are not affected.

## Scope

The change is limited to generated reader CSS and the existing document layout marker applied by `FoliateViewer`.

It does not modify:

- EPUB source files or document parsing
- Text nodes or DOM reading order
- Selection and annotation behavior
- CFI generation
- TTS extraction or playback
- Focus handling or accessibility semantics

The computed font-size behavior is covered in real Chromium for publisher-defined body sizes, disabled font-family override, minimum font size, fixed-layout isolation, RTL text, and vertical writing.


ChuwuYo:
Verified locally using the EPUB provided in #5420.

The body text now follows the configured font size with “Override Book Font” both enabled and disabled. The setting also remains effective after reopening the book, while the part and chapter headings retain their relative sizing.

chrox:
Thanks for the thorough analysis and the fix. I pushed a simplified version of this branch with two changes.

First, the `readest-reflowable` / `readest-fixed-layout` marker classes are gone. The CSS from `getFontStyles` is delivered through `renderer.setStyles`, which only the reflowable paginator implements; the fixed-layout renderer has no `setStyles` at all. So this stylesheet can never reach fixed-layout EPUBs, PDFs, or CBZ files, and the per-document gating was redundant.

Second, the normalization now applies only when "Override Book Font" is enabled. The rule is a blanket `!important` flatten of every `p, li, div, pre, dd`, and many books size those elements on purpose. The reproduction book itself marks up its chapter titles as `<p class="ct">` and `<p class="chn">` at `x-large`: with the rule always on, those titles would collapse to body size while books using real `h1` headings would keep theirs. Footnotes, captions, and poetry styled smaller would likewise get bumped to full body size. This is the same reason Readium CSS ships fs-normalize as an explicit opt-in rather than a default; its own comments note the normalize is never applied automatically.

The #267 behavior is unchanged for the base size: the configured font size still applies to `html`/`body` unconditionally. What sits behind the toggle is the stronger step of discarding publisher sizing on body-copy elements, and "Override Book Font" is the existing switch for preferring your settings over the publisher's typography. It also matches the repro steps in #5420, where the reporter had that setting enabled.

I dropped the tests to keep the change minimal. Verified with the repro book: override off leaves `p.indent` at 0.875x the setting (publisher styling intact), override on renders body copy at exactly the configured size with the minimum font size still enforced.

ChuwuYo:
> Thanks for the thorough analysis and the fix. I pushed a simplified version of this branch with two changes.
> 
> First, the `readest-reflowable` / `readest-fixed-layout` marker classes are gone. The CSS from `getFontStyles` is delivered through `renderer.setStyles`, which only the reflowable paginator implements; the fixed-layout renderer has no `setStyles` at all. So this stylesheet can never reach fixed-layout EPUBs, PDFs, or CBZ files, and the per-document gating was redundant.
> 
> Second, the normalization now applies only when "Override Book Font" is enabled. The rule is a blanket `!important` flatten of every `p, li, div, pre, dd`, and many books size those elements on purpose. The reproduction book itself marks up its chapter titles as `<p class="ct">` and `<p class="chn">` at `x-large`: with the rule always on, those titles would collapse to body size while books using real `h1` headings would keep theirs. Footnotes, captions, and poetry styled smaller would likewise get bumped to full body size. This is the same reason Readium CSS ships fs-normalize as an explicit opt-in rather than a default; its own comments note the normalize is never applied automatically.
> 
> The #267 behavior is unchanged for the base size: the configured font size still applies to `html`/`body` unconditionally. What sits behind the toggle is the stronger step of discarding publisher sizing on body-copy elements, and "Override Book Font" is the existing switch for preferring your settings over the publisher's typography. It also matches the repro steps in #5420, where the reporter had that setting enabled.
> 
> I dropped the tests to keep the change minimal. Verified with the repro book: override off leaves `p.indent` at 0.875x the setting (publisher styling intact), override on renders body copy at exactly the configured size with the minimum font size still enforced.

Thanks for your reply.

Issue #480, opened 2025-03-03
FR: Essential EPUB3 support


chrox:
**Does your feature request involve difficulty for you to complete a task? Please describe.**
Currently, Readest does not fully support EPUB3, which results in rendering issues for modern EPUB files. Many books that use EPUB3 features, such as advanced typography, multimedia, and interactive content, do not function correctly. This makes it difficult to read or interact with EPUB3-based books.

**Describe the solution you'd like**
	

- [x] Support for multimedia elements (audio, video).
- [x]  Media Overlay for audio and video
- [x] Support for fixed layout documents.
- [x] Rendering of MathML for mathematical content.
- [x] Interactive content support, including JavaScript-based features.
- [x] Improved metadata and navigation support.

At the very least, Readest should correctly render the EPUB3 sample books available here: [EPUB3 Samples](https://idpf.github.io/epub3-samples/30/samples.html).

**Additional context**
Adding support for EPUB3 would greatly enhance Readest’s compatibility with modern books and ensure a better reading experience. Priority should be given to ensuring Readest can properly render the EPUB3 sample books from the IDPF collection: [EPUB3 Samples](https://idpf.github.io/epub3-samples/30/samples.html).

Svetloslav:
A great implementation. Readest should be able to handle any digital publication, especially all epub standards. Thank you for your constant work on Readest. This is the best modern digital publications application on all platforms.

kvkhav:
For your reference:

One of the options at https://standardebooks.org/  is an "advanced epub" format, which according to [this answer](https://news.ycombinator.com/item?id=32215850) from them is epub 3.2.

The cover image of these books does not render in the readest bookshelves.

God-damnit-all:
@chrox I'm using `Readest_0.9.95_x64-portable.exe` and I imported a fixed layout epub. It renders fine, the problem is attempting to turn the book pages one at a time (even with the keyboard shortcuts <kbd>h</kbd> & <kbd>j</kbd>) has it instead jump between TOC sections instead; so it'll jump from page 3 to page 20 for example.

It's a shame, because I like the UI far better than Kobo Reader, and the portable implementation is very nice.

Would you please look into this?

chrox:
Could you share a link of the epub file or send it to readestapp@gmail.com?

God-damnit-all:
> Could you share a link of the epub file or send it to [readestapp@gmail.com](mailto:readestapp@gmail.com)?

[...]

PR #5480, opened 2026-08-03
feat(tts): play a book's own recorded narration (EPUB 3 Media Overlays)

### LINKED ISSUE #3924: FR: Support for Storyteller readalouds

> I would like Storyteller books, with Media Overlay, to be able to read in Readest. As of now, they need to be downloaded into my phone and it's cumbersome to get the web reader for it working with my current use case. Taking advantage of Readest's cloud storage would be the best case scenario, bonus points if they can also synced with Koreader so I can pick up the same book in any format and with any device.

> The current TTS option is good, but of course the robotic voice doesn't have the same feeling as a book narrated by a professional narrator. 

**Additional context**

> Here is a sample file, as mentioned in Discord: https://nextcloud.friedmans.us/index.php/s/tH2cPynNHgedaLe It's Peter and Wendy, a DRM free book

cubicruler:
closes #3924
refs #480 (ticks the "Media Overlay for audio and video" box)

When a book ships with a recorded human narration, Readest now plays that
recording instead of synthesizing speech, while keeping everything else Read
Aloud already does: the moving highlight, page-following, sentence/paragraph
skip, the scrubber and seek, speed, sleep timer, lock-screen and CarPlay
controls, and background sessions.

This is the Kindle Immersion Reading / Audible Read & Listen experience built on
the open EPUB standard rather than a store-side pairing of two purchases. It
needs no account and no matching entitlements: any narrated EPUB you own or
generate plays, on every platform Readest runs on. Storyteller output works
directly, which is what #3924 asked for.

## What it does

- Reads the narration from **EPUB 3 Media Overlays**: a SMIL file per spine
  section whose `<par>` elements pair a text fragment with a `clipBegin`/`clipEnd`
  range of an audio file. Those pairs are the publisher's own sync points, so
  Readest performs no alignment of its own.
- The narrator is preselected and leads the **Voice** list for books that carry
  overlays, named from `media:narrator` or "Book narration" when the book
  declares none. Picking a synthetic voice instead is remembered per book.
- The highlight follows the recording exactly rather than interpolating inside a
  clip, so it cannot drift. Word-level SMIL gives word-by-word highlighting for
  free.
- Sections the publisher left unnarrated are skipped rather than stalling on
  silence.
- Narrated books get a headphones badge in the library.
- The scrubber is exact: a narrated chapter reports the recording's real length
  with no `~`.

## Shape of the change

`TTSClient` already abstracted *where audio comes from* and foliate's `TTS`
abstracted *how text is cut into marks*. Recorded narration is exactly "a
different audio source with a different segmentation", so everything new lives in
`src/services/tts/mediaOverlay/` and the other ~25 files of the TTS subsystem are
untouched:

| File | Role |
| --- | --- |
| `parseSmil.ts` | Pure SMIL parsing: clock values, and walking `<body>`/`<seq>`/`<par>` in document order. |
| `MediaOverlaySection.ts` | Per-section index: resolves each par's text fragment to a DOM `Range`, groups pars into blocks, builds the SSML. |
| `MediaOverlayTTS.ts` | Stands in for foliate's `TTS` with the same navigation surface, driven by the par list. |
| `MediaOverlayClient.ts` | `implements TTSClient`. Plays clips off one `HTMLMediaElement`, emitting a `boundary` as each par becomes audible. |

Mark names are section-global par ordinals, so marks and clips are 1:1 by
construction and there is no text-to-audio matching anywhere in the feature.

## Changes to shared TTS code

Two additive, backward-compatible widenings in `TTSClient.ts`, both because a
recording is one continuous timeline rather than a series of separately rendered
utterances:

- `TTSCapabilities.continuousTimeline?: boolean` - consecutive blocks are one
  recording, so the controller must not insert its own inter-paragraph pause; the
  recording already contains the pauses the narrator made.
- `TTSClient.stop(handover?: boolean)` - distinguishes the stop the controller
  performs between two utterances of one session from a real stop, so a
  continuous engine can stay rolling through it.

`NativeTTSClient`, `WebSpeechClient` and `BufferedTTSClient` are untouched: the
field is optional and the parameter is ignored by engines that synthesize per
utterance, so their behaviour is unchanged.

Also here, and independent of narration: `ensureTimeline` /
`supportsPlaybackInfo` / `getPlaybackInfo` now gate on
`TTSCapabilities.mediaClock` instead of comparing against the Edge client, which
is what `TTSCapabilities` existed for. And a pre-existing hang is fixed:
`WebSpeechClient.init()` waited forever for a `voiceschanged` event on a platform
reporting zero speech voices, wedging `TTSController.init()` and so blocking
*every* engine. It now gives up after a timeout.

## Not in scope

Readest does not do the forced alignment itself. Dropping a bare MP3 next to a
book gives no timings, and inventing them would mean shipping a
speech-recognition model. That is a separate, one-time, offline job and
[Storyteller](https://storyteller-platform.dev/) already does it well;
`docs/read-along-narration.md` points at it and at syncabook/aeneas.

#3924 also asks for cloud-sync and KOReader parity for these books. This PR is
playback only.

## How to verify

Two publicly hosted W3C/IDPF samples, both free and DRM-free. They exercise
different paths, so it is worth trying both.

- **Moby-Dick MO** (10.1 MB) - the canonical Media Overlays sample and a
  deliberately awkward one: `h:mm:ss.mmm` clocks, mixed word- and sentence-level
  pars inside a single `<p>`, a nested `<seq>` carrying `epub:textref`, `.mp4`
  audio, and only 2 of its 144 spine sections narrated, so it covers gap handling.
  https://github.com/IDPF/epub3-samples/releases/download/20230704/moby-dick-mo.epub
- **Kusamakura** (17.9 MB) - Japanese vertical writing plus Media Overlays, with
  CJK-named SMIL files (`OPS/xhtml/*.smil`), which exercises percent-encoded href
  resolution alongside vertical-writing highlight geometry.
  https://github.com/IDPF/epub3-samples/releases/download/20230704/kusamakura-japanese-vertical-writing.epub

Import the book and open Read Aloud:

1. The narrator is preselected and sits at the top of the Voice list.
2. Playback is the recorded voice and the highlight advances with it. On
   Moby-Dick chapter 1 the highlight granularity visibly changes mid-paragraph
   (three per-word pars, then per-sentence). That is the publisher's markup, not
   drift.
3. The scrubber shows an exact chapter length with no `~`, and seeking lands
   where the recording actually is.
4. Pick a synthetic voice: the book falls back to TTS. Close and reopen it, the
   choice persisted. Pick the narrator again to return.
5. Moby-Dick MO: starting Read Aloud in an unnarrated section jumps forward to
   the first narrated one rather than stalling on silence.
6. Kusamakura: the highlight tracks correctly in vertical writing mode.
7. Paragraph transitions are seamless - no stutter repeating the first word, no
   silent gap. This is the most audible thing to check.
8. Select a passage and press Read Aloud: narration starts where that passage is
   narrated, rather than playing for a second and stopping.
9. The library cover shows a headphones badge. Books imported before this branch
   show no badge until re-imported; narration still works on them.
10. Regression check with a synthetic voice on a book with no overlays
    (`moby-dick.epub` from the same release is the natural control): the Voice
    list has no narration group and Read Aloud behaves as before.

Automated:

```bash
pnpm test -- media-overlay          # unit suites, no fixture needed

# opt-in end-to-end run of the real DocumentLoader against a real MO book
curl -sLO https://github.com/IDPF/epub3-samples/releases/download/20230704/moby-dick-mo.epub
READEST_MO_EPUB=$PWD/moby-dick-mo.epub pnpm test -- media-overlay-real-epub
```

`media-overlay-real-epub.test.ts` soft-skips without `READEST_MO_EPUB`, which is
why the ~10 MB fixture is not committed and CI does not run it.

## Tests

About 1840 lines of tests across the `media-overlay-*` suites: the SMIL parser,
the section index, the mark iterator, the client against a fake media element,
and controller-level narration selection, timeline exactness and section
skipping, plus coverage for the no-reseek and handover paths and for reading a
selection aloud.

Verified end to end in a browser against both W3C samples above and a 330 MB
Storyteller-generated novel (bare-seconds clocks, parent-relative audio hrefs, 17
SMIL files, 22 MP3s, no `media:narrator`, unnarrated front matter). Its computed
chapter timeline came out at 2966.8s against the book's declared `media:duration`
of 2966.79s.

That end-to-end pass is where most of the real defects came from - six of them,
each invisible to a jsdom test, and the last two audible only by listening. The
individual commit messages carry the detail.

## Limitations

- No `<seq>` skippability: `epub:type="pagebreak"`/`footnote` escape is not
  implemented. The parser keeps the `<seq>` structure so it can be added without
  a rewrite.
- `media:active-class` is ignored on purpose; the reader's own TTS highlight style
  and colour win.
- Chapter pre-download (Offline Audio) is Edge-only and hidden during narration,
  since the audio already ships inside the book.
- iOS Tauri is unverified. The client uses a plain `HTMLMediaElement`, which is
  better placed there than WebAudio, but Now Playing may not surface correctly. A
  native-player path would be a follow-up.
- Books already in the library carry no headphones badge until re-imported.

One thing worth flagging on the ops side: narrated EPUBs are an order of
magnitude larger than plain ones. The IDPF release above ships the same book both
ways - `moby-dick.epub` is 1.5 MB, `moby-dick-mo.epub` is 10.1 MB - and the
full-length Storyteller novel above came to 330 MB. A shelf of these sits in a
very different bracket from a normal library, which may be worth a thought
against how the sync tiers are currently sized.

Full write-up, including how to generate a narrated EPUB, is in
`apps/readest-app/docs/read-along-narration.md`.


chrox:
This looks great. Thanks.

chrox (approved):

Juansero29:
@cubicruler after doing tests on iOS, I noticed indeed the player stopped a small amount of time after tapping the "Play" button. Now Playing didn't surface correctly.

 I implemented a PR to go trough a native iOS player instead while keeping all the functionality you implemented here #5562. This now works perfectly on iOS.

Thanks a lot for your work.

PR #5487, opened 2026-08-04
feat(bookorbit): add BookOrbit integration with annotations, bookmarks, stats, and status sync

### LINKED ISSUE #5426: FR: Support highlights, bookmarks, annotations, and statistics sync in BookOrbit Integration

Currently, when configuring the KOReader integration in Readest with a **BookOrbit** server (`https://YOUR_BOOK_ORBIT_DOMAIN/api/v1/koreader`), only reading progress is synchronized. 

While BookOrbit provides APIs/endpoints to sync highlights, bookmarks, annotations, and reading statistics (https://bookorbit.app/koreader/) (similar to KOReader's extended sync capabilities), Readest does not yet implement synchronization for these elements over the KOReader sync API.

Extend the existing KOReader sync integration in Readest to support full bi-directional (or outbound/inbound) synchronization of:
- **Highlights**
- **Bookmarks**
- **Annotations/Notes**
- **Reading Statistics**

This will allow users running Readest on non-KOReader devices (e.g., Android app, Web, iOS) connected to a BookOrbit instance to maintain their notes and reading stats seamlessly.

Installing the KOReader sync plugin directly is only possible on devices running KOReader itself, which is not applicable for native Readest mobile/desktop app users.

### Additional Context

- Referenced Discussion: https://github.com/readest/readest/discussions/5393#discussioncomment-17863608
- As confirmed by @chrox in the discussion, BookOrbit supports this sync API, but implementation is required on the Readest side.

chrox:
Closes #5426

## What

Adds a dedicated BookOrbit integration that syncs with a self-hosted BookOrbit server through its KOReader plugin API, beyond the progress-only sync the KOReader integration offers today:

- **Reading progress**, driven through the existing kosync machinery via a provider parameter on `useKOSync` (`bookOrbitProgressProvider` points it at `{server}/api/v1/koreader`), including the conflict resolver.
- **Highlights and notes**, fully bidirectional via `POST /plugin/annotations/exchange` + `exchange-ack`: pushes Readest annotations as crengine xpointers (converted with the existing `xcfi` bridge), pulls annotations created on KOReader devices or the BookOrbit web reader, and propagates edits and deletions both ways with the protocol's identity-key scheme (`md5(datetime|pos0)`), complete-key-list deletion detection, and verified/corrected position acks.
- **Bookmarks** (dogears) via `POST /plugin/bookmarks/exchange`, gated on the server's `bookmarkSync` capability.
- **Reading statistics** via `POST /plugin/page-stats` with a dedicated `bookorbit-push` cursor in the statistics DB; Readest's `PageStatEvent` maps to the wire format one to one.
- **Reading status** (reading/finished/abandoned) via `POST /plugin/book-states`.

## How

- New `src/services/bookorbit/` module: wire types, pure note mapping (drawer/style and color tables matching `readest.koplugin`), pure exchange builders/apply-plans, a `BookOrbitClient` (KOSync-style `x-auth-user`/`x-auth-key` auth, tauri/LAN direct fetch, web proxy at `/api/bookorbit` with an anchored endpoint whitelist and the same SSRF hardening as the kosync proxy), a persisted identity map store (SQLite `bookorbit-sync` schema), and the `runBookOrbitNotesPass` orchestrator.
- Identity datetimes are derived from `createdAt` in the UTC frame and `deviceTime` is reported as UTC, so all Readest devices derive identical exchange keys; identities of remote-born notes are persisted since they cannot be re-derived. Pulled annotations get deterministic note ids matching `readest.koplugin`'s `generateNoteId`, so notes dedupe across sync paths.
- Settings: `BookOrbitSettings` slice with per-category toggles, replica whitelist + encrypted credential fields mirroring kosync, backup blacklists, lazy deviceId. New BookOrbit form and Integrations row (verified in light and e-ink rendering).
- The exchange respects the server limits (20 books / 50 changes per request, push-down paging via `more`), never calls `users/create` (BookOrbit credentials are minted in its web UI), and only advances per-book watermarks after a fully successful pass.

Fixed-layout (PDF posFormat) annotations are out of scope for v1, as is the catalog API (OPDS covers browsing).

## Testing

- 46 new unit tests across settings, mapping, client transport, proxy whitelist, sync store, both exchange cores, the orchestrator, and stats chunking/cursor semantics; full suite green (8567 passed), lint and format checks clean.
- Protocol shapes were implemented against the BookOrbit server source (`server/src/modules/koreader/` DTOs and services) and its bundled KOReader plugin.
- Pending: manual smoke against a live BookOrbit instance (highlight round-trip with the web reader and a KOReader device, deletion propagation, dogears, stats ingestion, status update, progress conflict prompt).


DraxonDraxoff:
Thanks so much for this. I've just shared in the BookOrbit discussion board. This functionality works really well. Much appreciated!

Issue #4901, opened 2026-07-02
Header and footer is unreadable when reading light mode PDF in dark mode


dastarruer:
## Bug Description

> Note how the header ('Process of Elimination (POE)')  in the top left corner and page number in the bottom right corner overlaps with the PDF, making it unreadable.
> 
> <img width="1872" height="998" alt="Image" src="https://github.com/user-attachments/assets/7924c0be-4295-434a-b7c7-f0a2bc50a39a" />

## Expected behavior

> Place a background behind the header and footer text? Not sure what the fix would be honestly.

## Technical inputs

Operating System: NixOS (unstable)

Readest Version: v0.11.12

PR #5503, opened 2026-08-05
fix(reader): render fixed layout documents edge to edge in scrolled mode

chrox:
## Problem

For fixed layout documents in scrolled mode, the document should render from edge to edge, but an opaque band the height of the top safe area inset covered the page at the display cutout (camera hole). The page scrolled under the band and was visibly clipped.

## Root cause

In scrolled mode `SectionInfo` paints a `notch-masked bg-base-100` strip clipped to the top inset so reflowable text does not show clipped under the status bar. The fixed layout DOM underneath is already edge to edge (the `foliate-fxl` host spans the full viewport and ignores renderer margins), and FXL chrome is designed to overlay the page rather than reserve bands (the section title already uses `mix-blend-difference`, #4901). The mask therefore wrongly occluded the document at the camera hole.

## Fix

Skip the mask classes when `bookData.isFixedLayout`, keeping the transparent notch area as the scroll-to-top tap target.

## Verification

- New unit test (fails before, passes after) plus the full suite, lint, and format checks
- Reproduced and verified on a Xiaomi 13: with the mask classes removed the page renders under the camera hole edge to edge

### DIFF
--- a/apps/readest-app/src/__tests__/app/reader/components/SectionInfo.test.tsx
+++ b/apps/readest-app/src/__tests__/app/reader/components/SectionInfo.test.tsx
@@ -78,6 +78,20 @@ describe('SectionInfo notch mask', () => {
     expect(notch.classList.contains('notch-masked')).toBe(false);
     expect(notch.classList.contains('bg-base-100')).toBe(false);
   });
+
+  it('keeps the notch transparent for fixed-layout documents so they render edge to edge', () => {
+    // FXL pages in scrolled mode fill the screen and their chrome overlays the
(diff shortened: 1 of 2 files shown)


chrox:
Device-verified on a Xiaomi 13 with the release build installed: in scrolled mode both a fixed layout PDF and an FXL EPUB now render from the very top edge under the camera hole, and CDP confirms the notch-area element no longer carries the notch-masked/bg-base-100 classes while foliate-fxl is in scrolled flow. Reflowable scrolled mode keeps the mask as before.

PR #5512, opened 2026-08-05
feat(send): clip locally opened html and xhtml pages with the browser extension

chrox:
## What

Makes Send to Readest work on a page the user opened from disk: a saved web page, an exported HTML report, a loose `.xhtml` chapter. Previously the extension refused any tab that was not `http(s)`.

## Why it did not work

Four independent gates, each fatal on its own:

| Gate | Symptom |
|---|---|
| `popup.ts` / `service-worker.ts` both tested `/^https?:/i` | "This page cannot be clipped", Send disabled |
| Chrome's per-extension **Allow access to file URLs** toggle | `chrome.scripting.executeScript` fails with an opaque host-permission error; `host_permissions` alone is not enough |
| `/api/send/inbox/file` 400s on a non-`http(s)` `X-Readest-Url` | upload rejected even when everything upstream worked |
| The asset bundler only ever fetched `http(s)`/`data:` | local images had nowhere to come from |

## Changes

**Extension** — new `src/lib/localPage.ts` (scheme test, file-access check, settings deep link). The popup accepts `file:` and, when access is off, disables Send and offers a button straight to `chrome://extensions/?id=<id>` rather than letting the clip fail opaquely. The service worker mirrors the gate. `uploadEpub` sends `X-Readest-Url` only for `http(s)`, which doubles as privacy: an absolute path out of someone's home directory should not reach the server.

The access check **fails open** — `chrome.extension` is trimmed under MV3, so a context that cannot ask surfaces the real error instead of a false "missing permission".

**Shared converter** — a local page keeps its images. An extension page holding file access *can* read `file://` URLs, so a "Save Page As, Complete" keeps its `<name>_files/` assets. (A content script in the page cannot — `fetch` and `XMLHttpRequest` are both blocked and the rendered `` taints a canvas — so the bundler is the only route.) `fetchBestFavicon` drops unfetchable candidates and no longer builds `"null/favicon.ico"` out of a `file://` page's `"null"` origin. A local page with no `<title>` is named after its file rather than its absolute path, and the cover label falls back to the file name instead of a placeholder avatar.

`.xhtml` needed no converter work: `DOMParser` handles Chrome's XML serialization. There are tests pinning that.

### Reading local files is fenced twice

The EPUB we build gets uploaded, so this capability is deliberately narrow:

1. **The clipped page must itself be `file://`.** A remote page pointing an `` at the disk reads nothing.
2. **The asset must sit under that page's own directory.** `new URL` has already normalized away any `..`, so a prefix test keeps a crafted local page from harvesting the rest of the disk. A page at the filesystem root gets nothing, since its "directory" would be `/`.

Both rules have unit tests, and rule 1 is also verified end to end in a real browser (below).

## Determinism

A clip of one URL, or of one offline page, has to produce the same EPUB with the same hash every time. Two bugs broke that; both are fixed here.

**1. Base URI.** `DOMParser` gives the parsed document the **calling realm's** base URI, and Readability rewrites every relative `href`/`src` against it. So relative image paths resolved to `chrome-extension://<id>/...` in the extension and `tauri://localhost/...` on desktop. Those images were silently dropped, and the same article clipped from two channels did **not** produce the byte-identical EPUB that `convertPageToEpub` documents. `parsePageDocument()` now injects `<base href>` from the page URL before Readability runs.

**2. Asset budget applied in completion order.** Which images survived a 30 MB overrun depended on which download won the race — identical inputs, different bytes, different hash. Results are now parked by index and the budget applied strictly in document order, so an overrun always truncates the tail. Dispatch still stops early once the settled document prefix has spent the budget, so an image-heavy page does not download bytes it will discard. The regression test forces the first image to resolve last; it fails on the old code with `[1,2,3,4,5,6]` instead of `[0,1,2,3,4,5]`.

What remains non-deterministic is input, not code: a transient fetch failure, or the page's own HTML differing between loads. That is exactly why library identity rests on the URL-derived identifier rather than on byte equality.

> [!NOTE]
> This changes remote clips too: a page with relative image paths now produces different (correct) EPUB bytes. It does **not** duplicate library entries. Dedup falls back to `metaHash` = `md5(title | authors | identifiers)`, and the identifier is `stableIdentifier(url)`, which nothing here touches — so `importBook` misses on `byHash`, hits on `byMetaKey`, and updates the existing entry in place (`existingBook.hash = hash`, config migrated to the new directory). `books.push` runs only when there is no existing book. What the user sees is the normal `byMetaKey` behaviour for any re-clip of changed content: the entry re-uploads (`uploadedAt` cleared) and its title/author are refreshed from the file, so a manual rename would be overwritten. Pinned by a test that asserts differing bytes but an identical `getMetadataHash`.

## Verification

Unit tests written before each fix, each confirmed to fail without it (including reverting `titleFallback` to check it was not passing for the wrong reason).

One caveat for anyone editing the budget-determinism test: the first version of it **passed on the buggy code**, because staggered fetch delays let the scheduler accidentally produce document order. It only becomes a real test once image 0 is forced to resolve well after all the others. A race test that does not reliably lose is not a test.

Then driven in a **real Chromium** with the extension loaded via Playwright, using the actual MV3 service worker, a real `file://` tab and a mock inbox endpoint:

- **A saved page clips end to end.** Unzipped: article body present; nav/banner/footer gone; the `data:` image embedded; the `<name>_files/` sibling embedded with bytes **identical** to the source file (`cmp` clean); cover labelled from the file name; no `X-Readest-Url`.
- **Security:** a remote page referencing `file:///…` uploads with no image entry and no leaked bytes.
- `chrome.extension.isAllowedFileSchemeAccess` **is** callable from an MV3 service worker and returns a real boolean.
- A local `.xhtml` reports `contentType: application/xhtml+xml` and serializes via `outerHTML` as XML with `xmlns` and self-closed tags.
- Denied path: the real service worker reports `{phase:'error', code:'restricted-page', message:'Readest needs permission to read local files.'}`.
- Remote clip still uploads and still sends `X-Readest-Url`.
- **Before/after on the base-URI fix:** `../img/hero.png` on a page served at `/posts/article` embeds with the fix; reverting the `<base>` injection and rebuilding drops it silently.

`pnpm test` 8665 passed / 16 skipped - `pnpm lint` clean - `pnpm format:check` clean - `pnpm build-browser-ext` compiles - extension `i18n:check` exits 0 (2 new strings translated across all 33 locales). No Rust touched.

## Known limitation

The favicon path stays `http(s)`-only on purpose, so a saved page does not carry its real site icon onto the generated cover; the file-name label and initial stand in. Cover art is cosmetic, and article images were the thing worth widening the blast radius for. Extending `fetchBestFavicon` under the same two fences would be a small follow-up if it's wanted.

## Not verified

Chrome's own enforcement (that it blocks the inject when the toggle is off) is not reproducible under `--load-extension`, which re-installs with the `ALLOW_FILE_ACCESS` creation flag on every launch and overwrites the pref. The denied path was exercised by stubbing the verdict in the live service worker, which covers our guard but not Chrome's. A 30-second manual check would close it: load `extensions/send-to-readest/dist` unpacked, leave **Allow access to file URLs** off, open the popup on a local page, and confirm the prompt appears instead of a failed clip.

### DIFF
--- a/apps/readest-app/extensions/send-to-readest/README.md
+++ b/apps/readest-app/extensions/send-to-readest/README.md
@@ -30,6 +30,36 @@ server writes the bytes to R2 and inserts a `send_inbox` row; the next
 Readest client to open drains the inbox and imports the EPUB as-is — no
 further server-side conversion.
 
+### Locally opened pages (`file://`)
+
+A saved web page, an exported HTML report, a loose `.xhtml` chapter — if
+the user can open it in a tab, the extension clips it through the exact
+same converter. Three things differ:
+
+- **Chrome requires "Allow access to file URLs"** on the extension's own
+  row in `chrome://extensions`; `host_permissions` alone does not grant it.
+  The popup checks `chrome.extension.isAllowedFileSchemeAccess()` first and
+  offers a button straight to that page, because without the toggle the
+  capture-script inject dies with an opaque host-permission error.
+- **Images survive, including local siblings.** An extension page holding
+  file access *can* read `file://` URLs (a content script in the page
+  cannot, and the rendered `<img>` taints a canvas, so the bundler is the
+  only route). A "Save Page As, Complete" therefore keeps its
+  `<name>_files/` images, a single-file save keeps its inlined `data:`
+  images, and an HTML-only save keeps whatever it still points at on the
+  original CDN.
+
+  Reading local files is fenced twice, because the EPUB we build is
+  uploaded: the clipped page must itself be `file://` (so a remote page
+  pointing an `<img>` at the disk reads nothing), and the asset must sit
+  under that page's own directory (so a crafted local page can't harvest
+  the rest of the disk). A page at the filesystem root gets nothing.
+- **The source URL never leaves the machine.** `X-Readest-Url` is sent only
+  for `http(s)` pages — the inbox endpoint rejects anything else anyway, and
+  an absolute path out of someone's home directory isn't ours to upload. The
+  file name stands in for the site name on the generated cover and, when the
+  page declares no `<title>`, as the book title.
+
 ## Why client-side conversion
 
 The previous version (`0.1.0`) sent only the page URL — the server then
@@ -133,7 +163,7 @@ toolbar badge updates, the lazy-load scroll dance (incl.
 phase. From the `apps/readest-app` workspace root:
 
 ```bash
-pnpm test:extension      # 47 shell tests, ~1 s
+pnpm test:extension      # 54 shell tests, ~1 s
 pnpm build-browser-ext   # production webpack build (catches alias / stub regressions)
 pnpm test                # full suite — also runs the extension tests via vitest's default glob
 ```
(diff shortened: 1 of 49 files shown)


ybbluesky:
一次可多个页面剪切吗？

chrox:
> 一次可多个页面剪切吗？

This is currently not planned.

chrox:
> 一次可多个页面剪切吗？

Please checkout the Sigil or Calibre for converting multiple webpages to an EPUB file.

ybbluesky:
> > 一次可多个页面剪切吗？
> 
> Please checkout the Sigil or Calibre for converting multiple webpages to an EPUB file.

Can the EPUB generated by clipping be downloaded locally first, rather than being uploaded directly?
剪切生成的epub不是直接上传，是可先自己下到了本地的吗？

chrox:
> Can the EPUB generated by clipping be downloaded locally first, rather than being uploaded directly?
> 剪切生成的epub不是直接上传，是可先自己下到了本地的吗？

This is not planned.

PR #5529, opened 2026-08-06
fix(annotator): fall back to the selection toolbar when the dictionary quick action gets a multi-word selection

### LINKED ISSUE #5213: Quick action improvement

When you use the Dictionary quick action, there’s two issues:

1. It activates even when selecting very long strings, which I’d imagine is never the intended behavior with dictionary
2. When closing the dictionary popup, there’s no way to trigger another action like highlighting or copying the text. It’s simply gone

It would be amazing if:

1. You could set the quick action only to trigger on single word long tap, not on highlighting a larger string
2. When you closed the dictionary popup, it still showed the toolbar

You currently have to disable the quick action altogether if you want to do anything else with the text

### Additional Context

_No response_

chrox:
## Summary

A dictionary lookup only makes sense for a single word or a short CJK term. When the selection was longer, the instant dictionary quick action still fired and left no way to highlight or copy the text. Now such selections show the normal annotation toolbar instead, keeping all actions reachable.

- New `isSingleLookupTerm` helper in `src/utils/word.ts`: a selection qualifies for the instant dictionary when it is a single whitespace-free token for space-delimited scripts, or a CJK run of at most 8 characters with no punctuation. CJK has no word spaces and `Intl.Segmenter` splits exactly the units worth looking up (4-char idioms, conjugated Japanese verbs, katakana compounds) into multiple word-like segments, so a length cap stands in for the word boundary there.
- Only the `dictionary` case of the quick action dispatch changes; translate/copy/search/TTS/share still make sense on passages and are untouched. The Android touchend deferral and the desktop long-press gate stay in front of the fallback.
- The other half of the issue (returning to the selection toolbar after closing the lookup popup) was already handled in #5526.

Closes #5213

## Test plan

- New unit tests for `isSingleLookupTerm` covering single words, hyphenated and apostrophe forms, multi-word rejection, CJK compounds/idioms up to the 8-char cap, code-point counting for surrogate-pair Han, and whitespace/punctuation rejection.
- Full suite: 8725 passed; `pnpm lint` and `pnpm format:check` clean.
- Verified live in Chrome (web build): with Quick Action = Dictionary, a multi-word drag now shows the annotation toolbar while a single-word drag still opens the dictionary popup; closing the popup returns to the toolbar per #5526.

### DIFF
--- a/apps/readest-app/src/__tests__/utils/word-lookup-term.test.ts
+++ b/apps/readest-app/src/__tests__/utils/word-lookup-term.test.ts
@@ -1,0 +1,60 @@
+import { describe, it, expect } from 'vitest';
+import { isSingleLookupTerm } from '../../utils/word';
+
+describe('isSingleLookupTerm', () => {
+  it('rejects empty and whitespace-only text', () => {
+    expect(isSingleLookupTerm('')).toBe(false);
+    expect(isSingleLookupTerm('   ')).toBe(false);
+    expect(isSingleLookupTerm('\n\t')).toBe(false);
+  });
+
+  it('accepts a single word in space-delimited scripts', () => {
+    expect(isSingleLookupTerm('serendipity')).toBe(true);
+    expect(isSingleLookupTerm('café')).toBe(true);
+    expect(isSingleLookupTerm("don't")).toBe(true);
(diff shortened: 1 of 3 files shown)


ajdubovoy:
This is awesome, thank you! 🙏

PR #5555, opened 2026-08-07
fix(translate): restore Azure Translator, keep paragraph layout, preserve inline formatting

### LINKED ISSUE #1582: Translated text formatting
## Bug Description

After translation, text formatting doesn't work. There's no italics or bold. Also, the text size in the original can change specifically for formatting, which doesn't work when translating.

Image

## Expected behavior

The translated text should be formatted according to the original. Also, in the original, sometimes the text is formatted using size, for example, whisper is written in a smaller font.

## Technical inputs

Operating System:
Android 16 (WebView 137.0.7151.118)
Readest Version: 0.9.63

chrox:
## Why

Azure Translator stopped working entirely: every paragraph failed with
`Failed to get auth token: 404`. Microsoft retired
`https://edge.microsoft.com/translate/auth` as part of the Edge translation
shutdown dated **2026-07-30** — the whole `/translate/*` namespace on that host
now 404s for any UA, header or path variant. The translate API itself is still
alive (it answers a clean 401), but no public path can mint a token for it
anymore. Nothing regressed on our side; `azure.ts` had not changed since it was
written.

Fixing that surfaced three further problems in the same area, so they are fixed
together here.

### 1. Azure Translator restored (first commit)

Auth material now comes from the `www.bing.com/translator` page and requests go
to `www.bing.com/ttranslatev3`, which returns the same response shape the
provider already parsed. Verified against the live service:

- **bing.com sends no CORS headers**, unlike the old Microsoft endpoints which
  allowed any origin. A browser cannot call it at all, so web builds go through
  a new same-origin proxy at `/api/azure-translate` (mirroring the existing
  yandex proxy: auth, origin check, rate limit, fixed endpoints only, no SSRF
  surface). The provider now reports `authRequired` in web builds. **Tauri is
  unaffected and still calls out directly with no sign-in.**
- **`IG` and `IID` are mandatory**, not telemetry decoration — omitting either
  is answered with `statusCode: 400`.
- **Failures hide inside HTTP 200**: an expired token is `statusCode: 205`, an
  over-long text is `statusCode: 400`. `response.ok` alone is not enough. A 205
  invalidates the cached params and retries once.
- **The cap is exactly 1000 UTF-16 code units**, counting code units rather than
  bytes — CJK and emoji measure the same as ASCII (1000 passes, 1001 fails, for
  all three). Longer texts are chunked, reusing the splitter that already lived
  in the yandex provider; it moves to `translators/utils` so both share it.

Outbound requests are capped at the concurrency the proxy allows. Bing
translates one text per request, so an unbounded fan-out over a page of
paragraphs had the surplus come straight back as 429.

### 2. Source paragraphs no longer lose their layout

`getParagraphLayoutStyles` treats a `<div>` as paragraph-like only when every
descendant is inline formatting. The translation target is appended inside the
source paragraph as a `<font>`, which was not on that list — so the div stopped
matching and silently dropped the **whole** rule: line-height, word and letter
spacing, text-indent and hyphenation all reverted to the book default. A book
that uses `<font>` markup of its own hit the same latent bug with no translation
involved.

The tag list becomes `INLINE_FORMATTING_TAGS`, shared by the layout rule and the
translation sanitizer so preserved formatting can never emit a tag that breaks
layout. This is the third time this rule has been broken this way (after the
a11y skip link), so the constant now carries the reasoning.

### 3. Translated text keeps its inline formatting — closes #1582

Formatting died because the pipeline sent `el.textContent` and wrote back plain
text. It now round-trips the paragraph's inline markup, because the translator
is what maps formatting onto the translated wording:

```
IN : The <b>quick</b> brown fox jumps over the <i>lazy</i> dog.
OUT: 那只<b>敏捷</b>的棕色狐狸跳过了那只<i>懒惰</i>的狗。
```

The tags land on the semantically matching words despite the reordering.
Providers declare this with a `preservesMarkup` capability flag, set only for
the provider actually verified against a live API; everything else falls back to
today's plain-text path.

The response is untrusted markup destined for the book DOM, so it is parsed
detached and rebuilt through an allowlist — never assigned as `innerHTML`.
Attributes use **echo validation**: an attribute survives only when that exact
tag/name/value appeared in the source we sent. So `<span class="whisper">` (the
small-font case in the issue) survives, while the response cannot introduce a
class, `href`, `style` or event handler of its own.

Markup is split at run boundaries so a formatted paragraph over the provider's
cap keeps its formatting rather than falling back. Formatting straddling a
boundary is re-opened in the next chunk, so every chunk is well-formed and
independently translatable. Content Readest injects (skip links, WordLens
glosses) is stripped before sending, so it is never translated as if it were
prose.

### 4. Highlights on the original survive hiding it

Hiding the original erased its text nodes, which destroyed the anchor for any
highlight on it. It is now wrapped in a `cfi-skip` element and hidden with CSS:
`cfi-skip` hoists the wrapper's children for indexing, so text nodes keep the
exact index and offset they had unwrapped. The stash-and-restore of original
text is no longer needed and is removed.

The target node also collapses from three nested wrappers to one, shortening the
CFI path into translated text and putting the block display on the element
itself rather than a block nested inside an inline parent. Since translations
are inserted lazily, each insert now asks the annotator to re-attach annotations
falling inside it.

**Translation nodes stay visible to CFI**, so highlights on translated text keep
working.

## Verification

- `pnpm test` — 8899 passed
- `pnpm test:browser` — 353 passed
- `pnpm lint`, `pnpm format:check` — clean
- No Rust, Lua or i18n changes.

Verified in Chrome against a real book:

- 21/21 proxy calls returned 200, **zero 429s**, console clean.
- All 4 markup-bearing paragraphs kept their formatting, including one at 3389
  characters of HTML that needed chunking. Before this change, 0 of 4 did.
- 8 `<div>` paragraphs, **0 with lost line-height**, no disallowed tags emitted
  into translations.

New regression tests cover the CFI invariants (including the original hidden),
the sanitizer's adversarial cases, the chunker's guarantees, and the
paragraph-layout rule under real `:has()`.

## Notes for review

- **Web builds now require sign-in for this provider**, because bing.com cannot
  be called from a browser and the proxy is authenticated, following the yandex
  precedent. Desktop and mobile are unaffected.
- `preservesMarkup` is enabled only for the provider verified against the live
  API. Google, DeepL and Yandex document equivalent modes but are left off until
  someone runs them.
- **Known, deliberate:** reading progress can still anchor inside translated
  text, since `paginator.js` excludes only `cfi-inert` nodes from its
  visible-range walk. On a book with no native `<font>` this makes *remote*
  progress sync fail (`Element index 0 out of bounds for tag font`). This is
  pre-existing and harmless locally. A fix keyed on `translation-element-mark`
  was considered and deliberately not taken, because translations are meant to
  stay visible to CFI.
- Collapsing the wrappers changes the CFI path into translated text, so a
  highlight saved inside translated text before this change will not resolve. It
  degrades silently (skipped, record kept) rather than mis-anchoring, and
  translated-text highlighting was not a shipped feature.
- The visual separator between original and translation, requested in a
  follow-up comment on #1582, is not addressed here.

### DIFF
--- a/apps/readest-app/src/__tests__/app/api/azure-translate.test.ts
+++ b/apps/readest-app/src/__tests__/app/api/azure-translate.test.ts
@@ -1,0 +1,200 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { NextRequest } from 'next/server';
+
+const validateUserAndTokenMock = vi.hoisted(() => vi.fn());
+vi.mock('@/utils/access', () => ({
+  validateUserAndToken: (...args: unknown[]) => validateUserAndTokenMock(...args),
+}));
+
+import { POST } from '@/app/api/azure-translate/route';
+
+const BING_PAGE = `
+  <html><script>var IG:"01CE353230DE4BFD8A44466FDD91401A";</script>
+  <div data-iid="translator.5025"></div>
+  <script>var params_AbusePreventionHelper = [1786092445798,"page-token",3600000];</script>
+  </html>
+`;
+
+const makeReq = (
+  query = 'endpoint=translate',
+  init: {
+    body?: string;
+    origin?: string | null;
+    contentLength?: string | null;
+    authorization?: string;
+  } = {},
+) => {
+  const body = init.body ?? 'text=Hello&to=fr';
+  const headers: Record<string, string> = {
+    authorization: init.authorization ?? 'Bearer test-token',
+  };
+  if (init.origin !== null) headers['origin'] = init.origin ?? 'https://web.readest.com';
+  if (init.contentLength !== null) {
+    headers['content-length'] = init.contentLength ?? String(new TextEncoder().encode(body).length);
+  }
+  return new NextRequest(`https://web.readest.com/api/azure-translate?${query}`, {
+    method: 'POST',
+    body,
+    headers,
+  });
+};
+
+let fetchSpy: ReturnType<typeof vi.fn>;
+
+beforeEach(() => {
+  validateUserAndTokenMock.mockReset().mockImplementation(async (authorization: string | null) => ({
+    user: authorization ? { id: authorization } : null,
+    token: authorization,
+  }));
+  fetchSpy = vi.fn().mockResolvedValue(
+    new Response('[{"translations":[{"text":"Bonjour"}]}]', {
+      status: 200,
+      headers: { 'Content-Type': 'application/json' },
+    }),
+  );
+  vi.stubGlobal('fetch', fetchSpy);
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+  vi.restoreAllMocks();
+});
+
+describe('azure-translate proxy route', () => {
+  it('returns 403 before reading the body or fetching when unauthenticated', async () => {
+    validateUserAndTokenMock.mockResolvedValue({ user: null, token: null });
+    const request = makeReq();
+    const textSpy = vi.spyOn(request, 'text');
+
+    const res = await POST(request);
+
+    expect(res.status).toBe(403);
+    expect(await res.json()).toEqual({ error: 'Not authenticated' });
+    expect(textSpy).not.toHaveBeenCalled();
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
+
+  it('returns 400 for an unknown endpoint without fetching', async () => {
+    const res = await POST(makeReq('endpoint=evil'));
+    expect(res.status).toBe(400);
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
+
+  it('never lets a caller redirect the upstream fetch elsewhere', async () => {
+    // `endpoint` is matched against a fixed allowlist rather than used as a
+    // URL, so a caller cannot point the proxy at an arbitrary host.
+    const res = await POST(makeReq('endpoint=https://evil.example.com'));
+    expect(res.status).toBe(400);
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
+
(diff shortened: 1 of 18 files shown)


chrox:
### Provider coverage for #1582 — verified against the live APIs

Pushed `8bdcd4fa1`, which enables markup preservation for **Google** as well. Both providers were measured against their real endpoints rather than their documentation:

| Case | Azure/Bing | Google | DeepL |
| --- | --- | --- | --- |
| `<b>` / `<i>` / `<strong>` alone | ok | ok | ok |
| `<em>` alone | ok | ok | **tag dropped entirely** |
| bold + italic in one sentence | ok | ok | **`<b></b>` emptied — nothing renders bold** |
| nested tags | ok | ok | — |
| `class` / `href` attributes | ok | ok | ok |
| non-Latin target | ok | ok | — |

Google needs no extra parameter; adding `format=html` actually makes it **strip** the tags, so that would have been the wrong move.

**DeepL stays on the plain-text path.** It fails in two different silent ways, so markup would sometimes claim formatting that is not rendered — worse than losing the formatting but keeping correct text. DeepL proper does support `tag_handling=html`, but that has to be set by the `/deepl/translate` service, which lives outside this repo; passing the field from the client is ignored.

### This does not fully close #1582 yet

`translationProvider` defaults to `deepl`, so a user on default settings still sees unformatted translations. Closing the issue would be premature until either the `/deepl/translate` service sets `tag_handling`, or the default changes. Worth deciding separately.

The follow-up comment on that issue — a visual cue separating original from translation — is also still unaddressed and is a design question rather than a defect, so it probably belongs in its own issue.

PR #5611, opened 2026-08-10
feat(localsend): send and receive books over the local network

chrox:
## Summary

Adds LocalSend support so Readest can send and receive book files over the local network, fully wire-compatible with LocalSend protocol v2.2 — it interoperates with the official LocalSend apps (desktop/mobile) and with other Readest devices.

The protocol layer embeds LocalSend's own Rust core (`localsend` crate, pinned) behind a Tauri service module; the frontend adds the transfer UI and hands received books to the existing `ingestFile` pipeline so a sent book behaves exactly like a locally-imported one.

## What's included

- **Receive:** an opt-in service (Settings → Integrations → LocalSend). While enabled, Readest announces itself and shows an accept/decline dialog per incoming request. Only supported book formats are accepted (partial-accept is protocol-native — other files are declined and the sender sees them skipped); accepted books are imported into the library.
- **Send:** "Send to Nearby Device" in the book context menu and the multi-select bar opens a device picker of discovered peers.
- **Discovery:** multicast (UDP 224.0.0.167:53317, shared via `SO_REUSEPORT`) with an HTTP subnet-scan fallback.
- **Coexistence:** Readest's HTTP server binds `53318+`, leaving the LocalSend default `53317` free so both apps run on one device without a port conflict; the multicast announce carries Readest's real port so it stays discoverable.
- **Device tag:** peers see the OS name (macOS / iOS / iPadOS / Android / Windows / Linux).
- **Platforms:** all Tauri targets. Android gets multicast permissions + a `MulticastLock`; iOS/macOS get `NSLocalNetworkUsageDescription`. iOS discovery degrades to the HTTP scan until Apple's multicast entitlement is granted. Web build hides the feature.

## Testing

- Unit: vitest (format filter, transfer store, device-model + context-menu gating) and Rust (`cargo test`: identity persistence, port fallback, event serialization).
- `pnpm test`, `pnpm lint`, `pnpm fmt:check`, `pnpm clippy:check`, `pnpm test:rust` all green.
- **Verified on hardware (macOS Readest ↔ Xiaomi Android LocalSend), both directions:** discovery, partial-accept (PDF imported, SVG skipped — sender saw Done/Skipped), library import, and sending books to the phone. Same-host port fallback confirmed against the desktop LocalSend app.
- iOS: Rust cross-compiles cleanly for `aarch64-apple-ios`; not yet run on device/simulator (a real transfer needs hardware + the multicast entitlement).


archisman-panigrahi:
How about adding support for KDE Connect? It is available on Linux, Windows, Android, iOS and also for Mac via this (https://github.com/TURX/kdeconnect-mac).

Perhaps readest will show up us a different computer and we can send books to it?

chrox:
> How about adding support for KDE Connect? It is available on Linux, Windows, Android, iOS and also for Mac via this (https://github.com/TURX/kdeconnect-mac).

It might be a good idea. But LocalSend already works great on all my devices (Linux, Windows, macOS, Android, iOS and Kindle with the LocalSend plugin). I see no good reason to switch to kdeconnect.

archisman-panigrahi:
> I see no good reason to switch to kdeconnect.

It can work as just another option. KDE Connect can be made to work from anywhere in the world, beyond just local networks https://askubuntu.com/q/1565782/124466. I don't know if localsend supports it. But with KDE Connect, it may be possible to "Send to Readest" from anywhere.

chrox:
> I don't know if localsend supports it. But with KDE Connect, it may be possible to "Send to Readest" from anywhere.

I don't think it's possible without a relay server to send files across networks. This is an approach we are considering to support but not necessarily with KDE Connect.

PR #5618, opened 2026-08-10
fix(reader): isolate the book cell to drop the Linux black corner

### LINKED ISSUE #5609: BUG DESCRIPTION: while reading, a black square on the upper left hand corner
### Bug Description

There is a black square at the upper left hand corner.

Attached the video:

https://github.com/user-attachments/assets/f61cfbdc-9710-4043-9d02-2bdcea8f7c65

### Steps to Reproduce

Just open a book and the black square will be on the reading mode

### Expected Behavior

No black square

### Reproduction File

_No response_

### Operating System

Linux

### OS Version

CachyOS

### Readest Version

0.12.1

### Custom CSS

_No response_

PolybiusPro:
Fixes #5609.

The black corner square on Linux is a WebKitGTK damage-tracking defect, not a blend artifact: an ~80×124 device-px rect at the window origin stops repainting while the reader mounts and freezes the pre-sidebar `base-100` paint (black on dark themes). Live instrumentation of the Flatpak build showed it reproduces on reflowable books with no active blend and with view transitions stubbed, on both the GNOME 49 (WebKit 2.50) and 50 (WebKit 2.52) runtimes.

`isolate` on the book cell mitigates it by giving the cell its own stacking context, which re-layerizes the corner so the frozen region only ever captures already-correct pixels — the same masking the resting Ribbon's top-left layer provided before #5493 moved the ribbon to the top-right. The underlying defect is upstream in WebKitGTK.

Verified on the Flathub packaging (GNOME 50 runtime): black square on every book open without the change, gone with it.


winstonma:
Thanks. On my compiled version it seems the PR fixed the black square when the sidebar is opened. But the black square exist when the sidebar is closed. Please check the video:

https://github.com/user-attachments/assets/4b7e4f33-35a1-478c-81a9-93c68901b9da

BTW I am not using flatpak. I am using [Arch PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/?h=readest-git) script to build.

Thank you very much

chrox:
@winstonma Thanks for the report and the video — you were right, and this PR's fix turned out to be a mitigation rather than a cure: the \`isolate\` only changed *which* pixels get stuck in the corner (with it, the frozen rect captures the open sidebar, so it shows exactly when you dismiss the sidebar).

We tracked down the real root cause: since #5351 the popup pointer triangle carries a CSS \`drop-shadow\` filter, and the footnote popup keeps a closed popup mounted in every book cell — so an invisible filtered element sits at the window origin whenever a book is open. On WebKitGTK that makes an ~80×124 device-px rect at the origin stop repainting entirely (bisected across release builds: 0.11.18/0.11.20 are clean, 0.12.x reproduces).

The proper fix is #5628 — it stops mounting the filtered triangle at rest and reverts this PR's \`isolate\`. Verified on the GNOME 50 Flatpak runtime: the corner stays clean through repeated sidebar open/close cycles and window resizes. It would be great if you could confirm on your build once it lands.

winstonma:
Yes it works. Thanks

PR #5651, opened 2026-08-12
fix(sanitizer): render Persian/Arabic half-space by converting misused RLM to ZWNJ

### LINKED ISSUE #5216: Readest fails to correctly render Persian text containing U+200F (RIGHT-TO-LEFT MARK / RLM)
### Bug Description

Many older Persian ebooks (produced with tools like InDesign or legacy converters) use RLM instead of a proper ZWNJ to create the "half-space" between word components (e.g. می‌روم, کتاب‌ها, رفته‌ام). This is a known authoring mistake, but it's still very common in existing epub files.
Other reader apps, such as Moon+ Reader, Al-reader, render this correctly (i.e., they treat it as a zero-width/invisible formatting character and don't disrupt word rendering), but Readest appears to mishandle it, producing broken/garbled word rendering wherever U+200F appears.

### Steps to Reproduce

Attached files:
Zip file containing test epub
Some screenshots

[readest_rlm_test_u+200f.zip](https://github.com/user-attachments/files/30168749/readest_rlm_test_u%2B200f.zip)
Image
Image
Image
Image
[readest_zwnj_test_u+200c.zip](https://github.com/user-attachments/files/30168854/readest_zwnj_test_u%2B200c.zip)

Steps to reproduce 

Open readest_rlm_test_u+200f.epub in Readest.
Look at the first block of text (words joined with U+200F).
Compare against readest_zwnj_test_u+200c.epub, which renders the same words correctly using U+200C (ZWNJ).
The rendering difference between the two characters should be visible. 

### Expected Behavior

U+200F should be treated as a zero-width, non-rendering bidi control character (as it is defined in the Unicode standard) and should not break or distort word shaping/rendering, consistent with how ZWNJ (U+200C) is already handled and how other reader apps (e.g. Moon+ Reader, ALReader) already handle it.

### Reproduction File

_No response_

### Operating System

Android

### OS Version

17 on Pixel 9 Pro 

### Readest Version

Version 0.11.18 (WebView 150.0.7871.46)

### Custom CSS

_No response_

shahram7:
## Problem

Many older Persian/Arabic EPUBs misuse the Right-to-Left Mark (RLM, U+200F)
as a half-space between the parts of a compound word. RLM is an invisible bidi control that does **not** break
the cursive joining form, so these words render joined and broken.

#5216/#5361 fixed a related-but-different issue: it *preserves* U+200F through
sanitization so XMLSerializer no longer turns it into `&#x200f;`. But
preserving the RLM doesn't fix rendering - the character that actually
produces a half-space is the Zero Width Non-Joiner (ZWNJ, U+200C). So the
bug persists in the current release.

## Fix

In `sanitizer.ts`, after the existing chain restores U+200F to its literal
form, swap RLM -> ZWNJ **only when it sits between two Arabic-script letters**.
This is the safe-swap rule: an RLM next to a digit, a Latin letter, or at a
text boundary can be a genuine direction hint, so it is left untouched.

Both marks are single UTF-16 code units, so the transformation is
length-preserving and existing CFIs / annotations stay valid.

## Testing

Verified the swap logic and its placement in the sanitizer pipeline against
the cases above (including consecutive RLMs and length preservation), and
confirmed correct rendering in the reader with a sample Persian EPUB.
[rlm_test_u+200f.zip](https://github.com/user-attachments/files/30986949/rlm_test_u%2B200f.zip)

Closes #5216

### DIFF
--- a/apps/readest-app/src/__tests__/services/transformers/transformers.test.ts
+++ b/apps/readest-app/src/__tests__/services/transformers/transformers.test.ts
@@ -723,6 +723,46 @@ describe('sanitizerTransformer', () => {
     expect(result).toMatch(/^<\?xml version="1\.0" encoding="utf-8"\?>/);
     expect(result).toContain('<!DOCTYPE html');
   });
+
+  const sanitize = async (text: string) => {
+    const html = `<html><head></head><body><p>${text}</p></body></html>`;
+    return sanitizerTransformer.transform(makeCtx({ content: html }));
+  };
+
+  test('converts RLM misused as half-space between Arabic letters to ZWNJ', async () => {
+    // "mi-ravam" written with RLM (U+200F) instead of ZWNJ (U+200C)
+    const result = await sanitize('می\u200Fروم');
+    expect(result).toContain('می\u200Cروم');
+    expect(result).not.toContain('\u200F');
+  });
+
+  test('converts a run of consecutive RLMs between Arabic letters, preserving length', async () => {
+    const result = await sanitize('ا\u200F\u200Fب');
+    expect(result).toContain('ا\u200C\u200Cب');
+    expect(result).not.toContain('\u200F');
+  });
+
+  test('converts RLM after an Arabic letter with a diacritic', async () => {
+    // letter + kasra (U+0650) + RLM + letter
+    const result = await sanitize('بِ\u200Fب');
+    expect(result).toContain('بِ\u200Cب');
+  });
+
+  test('leaves RLM between Arabic-script digits untouched', async () => {
+    // RLM between digits is a live bidi hint that keeps the digit runs
+    // separate; swapping it for ZWNJ would flip their visual order.
+    const persianDigits = await sanitize('۱\u200F۲');
+    expect(persianDigits).toContain('۱\u200F۲');
+    const arabicDigits = await sanitize('١\u200F٢');
+    expect(arabicDigits).toContain('١\u200F٢');
+  });
+
+  test('leaves RLM next to Latin text or at a text boundary untouched', async () => {
+    const latin = await sanitize('abc\u200Fا');
+    expect(latin).toContain('abc\u200Fا');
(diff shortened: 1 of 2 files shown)


chrox:
Thanks @shahram7, the core fix is correct and well reasoned: RLM is join-transparent, so it never breaks the cursive join, and ZWNJ is the proper half-space. The placement after the `&#x200f;` restoration chain is exactly right, and the length-preservation reasoning for CFIs/annotations holds.

I pushed a follow-up commit (1529d2e21) to this branch with three adjustments found during review:

1. **Digits were inside the character classes.** `U+0600-06FF` includes the Arabic-Indic (`U+0660-0669`, bidi class AN) and extended/Persian (`U+06F0-06F9`, bidi class EN) digits, so `۱‏۲` was converted even though the comment promised digits are left intact. An RLM between two digits is a live direction hint: RLM (strong R) keeps adjacent digit runs visually separate, while ZWNJ (bidi class BN, removed by rule X9) merges them into one number, flipping the visible digit order. The classes now exclude both digit ranges, and the presentation-forms range ends at `U+FEFC` so the BOM (`U+FEFF`) is not matched as a letter.

2. **The do/while loop could never make progress after the first pass.** The lookahead means a single global replace already catches every eligible RLM, and the replacement ZWNJ cannot create a new match. Consecutive RLMs (`ا‏‏ب`) were in fact never converted, because the inner RLM never has a letter neighbor on both sides. The loop is now a single replace that matches whole RLM runs and swaps them one-for-one, which handles the consecutive case while keeping the text length unchanged.

3. **Unit tests** for the swap and its guard cases (digit neighbors, Latin/boundary neighbors, diacritics, RLM runs) in the existing sanitizer test block, plus the regex is hoisted to a module-scope literal (fixes a Biome `useRegexLiterals` warning).

Verified locally: full unit suite, `pnpm lint`, and `pnpm format:check` all pass.

PR #5662, opened 2026-08-12
fix(ui): size the alert surface off its container, not its content

WantenMN:
## Summary

Fixed the delete confirmation popup's width jumping when toggling the purge option.

https://github.com/user-attachments/assets/edc5db83-121c-4c67-9f4d-d67e2060d8a3

## Fix

Every call site mounts `Alert` as the lone child of a `flex justify-center` bar. A flex item without a definite width sizes to its content, so the alert box measured itself off its own longest line: the delete confirmation resized mid-dialog whenever the purge toggle swapped in longer copy, and no two alerts in the app rendered at the same width.

Added `w-full` to the alert wrapper so the existing `max-w-md sm:max-w-lg md:max-w-xl` caps decide the width instead. This supersedes the original `max-w-md` wrapper around `DeleteConfirmAlert`, which pinned that one component while leaving the other alerts jittery, and pinned it 160px below the shared desktop cap.

Covered by `alert-width-stability.browser.test.tsx`, which measures real laid-out widths in Chromium (jsdom has no layout engine). It fails on `main` with the reported jump.

### DIFF
--- a/apps/readest-app/src/__tests__/components/alert-width-stability.browser.test.tsx
+++ b/apps/readest-app/src/__tests__/components/alert-width-stability.browser.test.tsx
@@ -1,0 +1,114 @@
+/**
+ * The alert surface must not resize with its own content.
+ *
+ * Every call site mounts the alert as the lone child of a
+ * `fixed inset-x-0 flex justify-center` bar. A flex item sizes to its content
+ * unless it is given a definite width, so the alert box measured itself off its
+ * longest line and jumped whenever that text changed — most visibly in
+ * DeleteConfirmAlert, where flipping the purge toggle swaps in a longer
+ * description and relabels the confirm button (issue: the popup grows/shrinks
+ * mid-interaction).
+ *
+ * Needs real layout and real Tailwind, so it runs as a browser test.
+ */
+
+import { describe, it, expect, afterEach, beforeAll, vi } from 'vitest';
+import { render, cleanup, fireEvent, screen } from '@testing-library/react';
+import { page } from 'vitest/browser';
+
+vi.mock('@/hooks/useTranslation', () => ({
+  useTranslation: () => (s: string) => s,
+}));
+vi.mock('@/context/EnvContext', () => ({
+  useEnv: () => ({ appService: null }),
+}));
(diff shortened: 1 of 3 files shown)


chrox:
Thanks for catching this — the bug and the repro video are spot on. I pushed a follow-up commit that moves the fix down a layer; details below so the change is not a surprise.

**Root cause is in `Alert`, not `DeleteConfirmAlert`.** Every call site (`Bookshelf`, `BookDetailModal`, `StorageManager`, `ClipSignInAlert`, `ReceiveRequestDialog`, `ModalPortal`) mounts the alert as the lone child of a `flex justify-center` bar. `Alert`'s outer wrapper carried no width, and a flex item without a definite width sizes to its *content* — so the `w-full max-w-md sm:max-w-lg md:max-w-xl` on the box inside was resolving against a shrink-to-fit parent. The surface measured itself off its own longest line. The purge toggle just made it obvious by swapping in longer copy plus a longer button label.

Measured widths at a 1024px viewport:

| | before | this PR as submitted | after |
|---|---|---|---|
| Delete alert, purge OFF | 395.7px | 416px | 576px |
| Delete alert, purge ON | 576px | 416px | 576px |
| Plain alert, short message | 192.7px | 192.7px | 576px |
| Plain alert, long message | 571.8px | 571.8px | 576px |

The `max-w-md` wrapper fixed the jump for this one component, but every other alert kept resizing with its content, and `max-w-md` also overrode the shared `sm:max-w-lg md:max-w-xl` caps — leaving the delete dialog ~160px narrower than every other alert on desktop.

**The fix is now `w-full` on `Alert`'s wrapper**, so the existing `max-w-*` caps decide the width. That reverts the wrapper here, taking the net diff from +41/-39 down to a single class. All alerts in the app now share one surface and none of them resize with their text.

Also added `src/__tests__/components/alert-width-stability.browser.test.tsx`. jsdom has no layout engine, so it runs as a browser test and measures real `getBoundingClientRect()` widths in Chromium with real Tailwind. Verified it fails on `main` with exactly the reported jump (395.7 -> 576) before passing with the fix.

PR #5684, opened 2026-08-13
feat(tts): scroll to the current chapter when opening Offline Audio

### LINKED ISSUE #5683: FR: jump to the current chapter when opening Offline Audio

For big books, the Offline Audio chapter list opens at the top, so reaching the current and next chapters means scrolling through hundreds of entries manually.

When the list opens, instantly scroll (no animation) to the currently playing chapter — the same one already marked with the now-playing icon.

_No response_

### Additional Context

_No response_

gfreitash:
closes #5683

### Cause
For big books the chapter list in Offline Audio can be hundreds of entries long. The app already marks the currently playing chapter with an icon, but the list opens scrolled to the top, so reaching the current and next chapters means manually scrolling past everything above them.

### Fix
When the Offline Audio view opens, the chapter list instantly scrolls the playing chapter to the top of the visible area (`scrollIntoView` with `behavior: 'instant'`, `block: 'start'` — no animation). It reuses the same `isPlaying` logic that drives the now-playing icon, so there is no new bookkeeping. The effect runs once on open, so the list does not re-jump as playback advances while it stays open, and it is a no-op when no chapter is active.

### Verification
- New unit tests: the list scrolls exactly once with `{ behavior: 'instant', block: 'start' }` when a chapter is playing, and not at all when none is.
- Manual: debug APK built for Android, installed on the emulator; the list opens on the current chapter with the next chapters right below.
- The 11 full-suite test failures (zip.js in novel-import/document/series-metadata/cover tests) were confirmed pre-existing: identical failures on the previous commit `b07baf52` without this change.

### DIFF
--- a/apps/readest-app/src/__tests__/components/tts/TTSChaptersView.test.tsx
+++ b/apps/readest-app/src/__tests__/components/tts/TTSChaptersView.test.tsx
@@ -1,6 +1,8 @@
 import { afterEach, describe, expect, test, vi } from 'vitest';
 import { cleanup, fireEvent, render, screen } from '@testing-library/react';
 
+const originalScrollIntoView = Element.prototype.scrollIntoView;
+
 vi.mock('@/hooks/useTranslation', () => ({
   useTranslation: () => (key: string, opts?: Record<string, unknown>) =>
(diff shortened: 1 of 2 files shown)


gfreitash:
Apparently the failed tests on pre-push hook were issues in my environment, since all passed here

chrox:
LGTM. Thanks.

PR #5690, opened 2026-08-13
feat(tts): queue chapter downloads with per-book persistence

### LINKED ISSUE #5688: TTS: starting a second chapter download stalls all downloads until the app restarts
### Bug Description

Downloading is a cancel-and-replace flow: tapping the download badge on chapter B while chapter A is downloading aborts A and starts B. The abort is only cosmetic — the sentence currently being synthesized keeps running for seconds, and when its run finally tears down it unconditionally resets the shared progress state, wiping B's active download from the UI even though B keeps downloading invisibly in the background.

`useTTSDownloads.runDownload` (apps/readest-app/src/app/reader/hooks/useTTSDownloads.ts) aborts the previous run via a shared AbortController, but the abort never reaches the in-flight synthesis (`warmSentence` passes its own fresh AbortController and the Edge provider ignores the signal), so the previous run's `finally` block fires asynchronously and resets the shared `download` state that the new run owns — a race that re-triggers on every subsequent download until the ghost runs are gone (restart).

### Steps to Reproduce

- tap to download A
- while A is downloading, tap to download B
- tap to download any other chapter or the ones above, you will be able to see the bugged behavior

### Expected Behavior

- Downloads keep working normally after any download is started, finished, or cancelled — no app restart should ever be required to unstick them.
- Tapping another chapter's badge mid-download never leaves that chapter stuck at a tiny fraction of progress while the rest of its download silently stops.

### Reproduction File

_No response_

### Operating System

Android

### OS Version

Android 16, Android 14

### Readest Version

0.12.1

### Custom CSS

_No response_

### LINKED ISSUE #5689: FR: TTS - queue chapter downloads and persist them across restarts

There is no download queue: tapping a chapter while another is downloading cancels the current one instead of waiting, and nothing about a download (queue position, progress) survives an app restart. For big books, downloading several chapters means babysitting the list one tap at a time.

### Proposed
- Spotify-style enqueue: tapping a chapter while another downloads adds it to a queue, processed one at a time; tapping a queued chapter removes it from the queue.
- Per-chapter states: pending / downloading / failed, with sentence-level progress (done/total) aggregated across the chapter's sections.
- Queue persistence across restarts: interrupted chapters resume as pending when the book is opened again.
- A "Download all" / "Cancel all" toggle for the whole book.
- Retry affordance for failed chapters (tap to re-queue).

### Done when
- Tapping chapter B while A downloads queues B; B starts automatically when A finishes.
- The queue survives an app restart and resumes pending chapters.
- "Download all" fills the queue; "Cancel all" empties it.

_No response_

### Additional Context

Relates to #5688

### LINKED ISSUE #5692: TTS: storage-limit eviction silently deletes downloaded chapters on big books
This is a draft issue for maintainer discussion rather than a ready-to-implement proposal: the fix depends on a product decision, how downloaded audio should relate to the cache storage limit, so I'd like input before opening a PR.

--- 

Downloaded chapters and the warm playback cache live in the same per-book audio store, and eviction treats them identically: when the storage limit is reached, the least-recently-used audio is deleted to make room for new audio (`evictUntilFits`).

For a book whose full audio exceeds the storage limit, "Download all" therefore thrashes — each newly downloaded chapter evicts the earliest-downloaded ones, so the queue reports success while the front of the book is silently deleted from the device. Hours of synthesis and bandwidth are wasted, and the user only discovers it when the chapter badges flip back to "not downloaded".

On top of that, there is no user-facing way to free TTS storage: downloads cannot be removed individually, the warm cache cannot be cleared, and the only lever is the storage-limit setting (50MB–1GB).

---

We could  treat downloads as pinned storage within the current shared limit: downloaded chapters are protected from eviction forever (removal only by explicit user action), eviction applies only to warm cache audio, and the download queue stops with a clear "storage limit reached" state when pinned audio fills the limit, resuming once the limit is raised. Per-chapter "remove download" and a "clear cache" action (warm audio only) give users control of their space.
We could also let downloads have their own storage limit, separate from the cache limit, each with its own eviction rules.

Should a per-book storage-limit sweep across multiple books ever delete a book whose chapters were explicitly downloaded? Should downloads removed on one device stop syncing to others?

gfreitash:
## Summary

Reworks Offline Audio into a durable, restart-safe download queue and fixes the cancellation race that could leave later chapter downloads running invisibly or stalled.

- Persists queued, active, and failed chapters by stable book hash, with sequential processing, retry, deduplication, aggregate progress, **Download all**, and **Cancel all**.
- Pins completed explicit downloads in SQLite so normal cache-budget eviction cannot remove them. Pinned audio may exceed the warm-cache budget by design.
- Adds one **Clear all** action for explicit downloads while preserving ordinary warm-cache audio.
- Makes cancel/requeue attempt-safe and verifies cache state before marking a download complete.
- Coordinates queue cancellation, reader-session teardown, cache closure, Clear all, and bookshelf deletion so late synthesis cannot recreate deleted audio or clobber a replacement reader session.
- Repairs stale/missing pack metadata and protects downloaded books from cross-book cache sweeps.

## Issue coverage

Closes #5688

- Replaces shared cancel-and-replace state with per-attempt ownership and settlement barriers, preventing an old run from clearing or aborting its successor.

Closes #5689

- Adds the persistent Spotify-style chapter queue, progress/status badges, retry, **Download all**, and **Cancel all**.
- Restored queue rows resume when the book is reopened.

Closes #5692

- Explicitly downloaded sections are pinned and are never evicted to satisfy the warm-cache storage limit.
- Provides one **Clear all** action and removes explicit downloads when the local bookshelf copy is deleted.
- Cloud-only deletion keeps local queue/audio state intact.

## Test coverage

```text
Offline Audio logical paths: 24/29 directly covered (83%)
├─ UI + hook queue ........ 5/6
├─ queue manager/store .... 6/6
├─ synthesis/cache chain .. 4/5
├─ SQLite pins/clear ...... 6/8
└─ session/delete order ... 3/4
```

The remaining five items are test-depth gaps around deferred persistence, legacy-schema upgrade, partial-download clearing, the hook-level Download-all boundary, and one local-delete ordering variant. No production defect was observed in those paths.

## Verification

- [x] `pnpm test`: 735 passed, 4 skipped test files; 9,250 passed, 16 skipped tests
- [x] `pnpm lint`
- [x] `pnpm format:check`
- [x] `pnpm build`
- [x] `git diff --check`
- [x] Pre-landing code and lite design review: no remaining findings
- [x] Native Android smoke test (Xiaomi 13, PR build): cancel/requeue mid-synthesis, restart recovery, Cancel all, Clear all, and bookshelf deletion — all passed; cache-limit survival covered by unit tests only

## Scope

The final diff is limited to Offline Audio queue/cache behavior, its reader controls, deletion/session integration, and regression tests. No unrelated product changes are included.

## Documentation

- `docs/architecture.md`: documents the durable Offline Audio queue, pinned packs, clearing behavior, and deletion cleanup.
- `docs/code-layout.md`: adds the queue/store and SQLite cache modules to the TTS reference.

## Summary by CodeRabbit

* **New Features**
  * Added persistent text-to-speech downloads with per-chapter queues, progress tracking, cancellation, retries, and resume support.
  * Added “Download all,” “Clear all,” and chapter-specific download controls.
  * Preserved downloaded audio during cache cleanup and removed it safely when local books are deleted.
  * Added Brazilian Portuguese translations for download queue actions.

* **Bug Fixes**
  * Improved TTS session shutdown and download cleanup reliability.
  * Prevented stale or interrupted downloads from replacing current playback state.


chrox:
Ran the native Android smoke test on a Xiaomi 13 with a PR build (aarch64 release + devtools), driving the app over CDP. Results:

**Passed**
- **Queueing**: chapters process strictly one at a time; waiting rows show the clock badge with "Waiting…", the active row shows the progress ring, and the header shows "Cancel all (N)". Queue rows persist with their section-range snapshots.
- **Cancel/requeue mid-synthesis (#5688)**: cancelling the active chapter lets the next queued chapter take over cleanly; requeuing parks the chapter behind it. Cancel all drained 11 rows while keeping downloaded chapters intact.
- **Restart recovery (#5689)**: force-killed the app mid-synthesis; completed chapters stayed pinned, the interrupted row demoted to pending, and the queue resumed on the next session attach.
- **Clear all / pinning (#5692)**: Clear all removed pinned packs (3 -> 0 chapters, 4.39 -> 2.67 MB) while ordinary warm-cache audio survived as "Partly downloaded". Fresh downloads pin and complete correctly.
- **Bookshelf deletion (#5692)**: local delete removed the book, its queue rows, and its downloaded audio with no cleanup errors in logcat.

**Observations (non-blocking)**
1. When playback reaches the end of the book, the session stops, which closes the player sheet and parks the remaining queue until the next session; a chapter that is downloading at that moment is marked "Download failed" and needs a manual re-tap. Defensible, but might be worth a follow-up so a book-end stop requeues the active item as pending instead of failed.
2. Saw once (could not reproduce): tapping download seconds after Clear all, while playback was ending at book end, dropped the queue row without downloading and without a failed state. Suspect the "everything already packed" early-exit reading statuses from a closing store. Re-tapping under a live session downloads fine.

**Housekeeping pushed as d806a1f** — added the manual `_one`/`_other` en plural entries for `Cancel all ({{count}})` per the project convention. Remaining before merge: `pnpm i18n:extract` + translations for the other locales (the new strings currently exist only in pt-BR).

PR #5704, opened 2026-08-14
feat(sync): optional document metadata on KOSync progress uploads

chrisbutler:
KOReader 2026.05 added an opt-in **"Send document metadata"** setting ([koreader/koreader#15306](https://github.com/koreader/koreader/pull/15306)): when enabled, progress uploads carry an optional `metadata` field with the document's filename, title and authors. The official sync server ignores it; custom KOSync-compatible servers use it to tell which book a document hash is, so books identify themselves instead of needing a manual pairing step per book. CrossPoint shipped the same field in 1.5.0. Readest is currently the main KOSync client without it.

This adds the same opt-in to Readest: a **Send Document Metadata** toggle in the KOReader Sync settings (off by default, matching KOReader's privacy stance), and the `metadata` field on progress uploads when enabled. Field shapes follow KOReader exactly: `authors` is a single string, newline-joined when there are several. Readest derives that value from structured metadata instead of the locale-formatted `Book.author` display string.

Real-world case: I run a KOSync-compatible tracker ([shelvd.io](https://shelvd.io)), with a user who syncs from a CrossPoint device and Readest on iOS against the same account. Their CrossPoint books now identify and match themselves automatically, while their Readest books still land as anonymous hashes to pair by hand. This would also help the hash-divergence problem discussed in #1838: when a Calibre re-export changes a file's checksum, a server that receives title and authors can still recognize both hashes as the same book, which a hash-only upload never allows.

Implementation notes:

- `filename` derives from `sourceTitle`/`title` plus the lowercased format rather than importing `EXTS`, keeping `KOSyncClient` off the document lib's foliate-js dependency chain (values are identical today).
- Stored configs without the new key behave as before (`sendMetadata` is optional and treated as off).
- Tests: metadata absent by default, exact field shape when enabled, multi-author newline serialization and fallbacks, setting persistence, and `sourceTitle` naming the file while `title` stays the user-visible one.

### DIFF
--- a/apps/readest-app/src/__tests__/components/settings/KOSyncFormSendMetadata.test.tsx
+++ b/apps/readest-app/src/__tests__/components/settings/KOSyncFormSendMetadata.test.tsx
@@ -1,0 +1,74 @@
+import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+
+import type { SystemSettings } from '@/types/settings';
+import { useSettingsStore } from '@/store/settingsStore';
+
+const saveSettings = vi.fn(async () => {});
+
+vi.mock('@tauri-apps/plugin-os', () => ({ type: vi.fn(async () => 'macos') }));
+
+vi.mock('@/context/EnvContext', () => ({
+  useEnv: () => ({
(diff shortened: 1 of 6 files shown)


chrisbutler:
Please let me know if anything here needs to be changed!

p.s. thanks for all the work y'all do 😁

Issue #5438, opened 2026-08-02
RSS feeds do not show up in android only on web. AND RSS feeds metadata is not presistent


mahumt:
### Bug Description

This is related to issue #5307 

RSS feeds are not showing up in android, when added from web.readest.com
And if metadata is edited for this RSS, then that change is not showing up anywhere else. Not in other instances of web.readest.com or in android [ I tried changing language to Swedish SV, so that TTS works properly]

Couple of steps that I took as troubleshooting.

I ended up logging out and clearing my data/cache in android.
Couple of things happened:

1. When I log out but have not yet cleared by data/cache. If I close out the sign in pop up, i can still see me books show up main page. I was even able to open up one of my books while still signed out. This also happens in web browser periodically.
2. When I sign back in, I had around 15 books in the app originally. Only one downloaded right there and then. The rest would not populate even when i did the pull-down-and-refresh. I ended up leaving the app alone for a day
3. Next day when i did the pull-down-and-refresh. The popup said 15 books to be synced and then showed all the books/groups properly
4. BUT the RSS feed i had added in from my android was gone from android. I can still see that RSS in my web.readest.com instances on two different laptops.
5. Also the RSS feeds added from wewb.readest.com are syncing and showing up in both laptops. But not on my phone

### Steps to Reproduce

Add RSS feed https://[8sidor.se/feed/](https://8sidor.se/feed/) in web
Sync android and check feed is not showing up 

Change metadata in web.readest.com for languages. change to SV and save. the language metadata changes on those instances only, not anywhere else.

### Expected Behavior

RSS feeds should be shown in mobile apps
RSS feeds should have persistent metadata

### Operating System

Web, Android

### OS Version

windows 11 (Edge 150.0.0.0); android 16

### Readest Version

Web: Version 0.11.20; Android: 0.11.20

chrox:
Fixed in #5442.

Issue #4634, opened 2026-06-17
FR: Sync reading status


zzzh3ng:
**Does your feature request involve difficulty for you to complete a task? Please describe.**

Sometimes I change the reading status of a book when it's at 95%~ progress. But this status doesn't sync across different devices, so I have to change the status on every device manually.

**Describe the solution you'd like**

If I change the reading status to unread or read on one device, the status would be synced to other devices, if the reading progress sync option is checked. Ideally, it would also sync with koreader, if it's feasible.

Issue #4544, opened 2026-06-12
Customized Book Cover Not Synced


gitkeniwo:
## Bug Description

For some reason, a pdf file kept showing an already-deleted page as the cover, even after I deleted that whole page from the pdf and replaced it with my desired pic in the PDF file. Tried labeling the desired cover pic as "cover" or "1" in pdf expert and nothing helped.

As this was driving me nuts, I tried changing the cover pic in readest's metadata editor, it worked for my mac where i was operating from, but the customized cover was not synced to my ipados and other devices. 

On my mac:
<img width="603" height="535" alt="Image" src="https://github.com/user-attachments/assets/e5e640d1-5d6d-449f-83b8-539df5846fb6" />

iPad:

<img width="211" height="379" alt="Image" src="https://github.com/user-attachments/assets/0d70cf4b-7e70-4a63-94c2-a0b5237928ac" />

## Expected behavior

Customized cover should be synced across devices

## Technical inputs

Operating System: macos

Readest Version:

Version 0.11.4

PR #5717, opened 2026-08-15
fix(sync): keep a book's reference page count on all devices, closes #5716

### LINKED ISSUE #5716: A reference page count set on one device never appears on other devices, and is then erased
### Bug Description

A reference page count set on one device never appears on the other devices. The
device without the count then erases it from the cloud, so the value survives
only where you typed it.

A second fault appears on import. Readest writes an empty config over an existing
Books/<hash>/config.json, so a backup restore loses that book's reading position,
bookmarks and annotations.

### Steps to Reproduce

1. Sign in to the same account on two devices, A and B.
2. On device A, open a book that has no publisher page list.
3. Set the progress style to Reference.
4. Set the reference page count, for example 350.
5. Wait for device A to sync.
6. On device B, open the same book.
7. Look at the footer. It shows a percentage, not page numbers.
8. Read one page on device B, then close the book.
9. Open the book on device A again. The page count is gone.

For the second fault:
1. Restore a backup that holds a book directory which the archive's library.json
   does not list.
2. Open that book.
3. The reading position, the bookmarks and the annotations are gone.

### Expected Behavior

The count stays in the cloud record instead of being erased by a device that does not have it.

### Reproduction File

_No response_

### Operating System

Linux

### OS Version

Bazzite (Current)

### Readest Version

Version 0.12.1 (WebView 605.1.15)

### Custom CSS

_No response_

bmfrosty:
Closes #5716

## The problem

The reference page count belongs to the book, not to the device. The reader sets it when the book has no publisher page list. The `reference` progress style maps the reading position onto that number.

The app sends the count to the server. It does not read the count back. Thus the count stays on one device.

A second device then makes the fault worse. That device holds no count. Its next push replaces `view_settings` in the record. The server keeps `view_settings` as one value. Thus the count is lost on all devices.

Two standard installations show this behaviour. An external tool is not necessary.

## What this change does

The reader now takes the count from a pulled record.

The last writer wins on the config `updatedAt` clock, and a tie goes to the remote value. A device that holds no count takes the remote count, whatever the clock shows.

The app keeps the value in memory in all cases. It writes the value to the disk only from the primary view, and never during a deep-link preview. In a preview the config holds the previewed position, which `useProgressAutoSave` does not persist.

The app examines the value first. The value comes through `JSON.parse` from a server column. A bad value would go to the disk and back to the server. Thus the app accepts a number from 0.1 to 10000 only.

## Limits

- Each page turn stamps the config clock. Thus a device that only reads can win against a device that edited. This fault is not new, and it applies to each value on that clock. The project corrected the same fault three times with one clock for each field: `reading_status` (#4634), `cover_hash` (#4544) and the metadata group (#5438). A correction for this one field does not help, because peers still push the full `view_settings` value on the row clock.
- On the wire, 0 and "not set" are the same. Thus you cannot clear the count, and the app undoes the clear. The device reads the old count again on the next open, because the clock test does not apply when the local count is 0. A device that still holds the old count also sends it back.
- The count moves on Readest Cloud sync only. The file-sync backends remove `viewSettings` from the wire.
- If the first pull fails each time, the app opens the push gate without a pull. That device holds no count, and its push removes the count from the record.
- A KOReader push writes NULL to `view_settings`. Thus that push removes the count from the record.
- A file-sync pull can remove the count for one session. The reader continues to show it, and the next open takes the count again.

## The second fault

`importBook` wrote an empty config for each new book. That write removed an existing `Books/<hash>/config.json`, which holds the reading position, the bookmarks and the notes.

The backup restore shows the cost. For a book directory that the archive's `library.json` does not list, the restore extracts each file, and `config.json` is one of them. Then it imports the book file. The import removed what the restore had recovered.

The comment above that code already gave the correct rule. The code did not obey it. The app now writes the empty config only when no config is present. It does this at both places that always wrote one.

The guard changes behaviour only when the library holds no record for that hash, because `buildBookLookupIndex` keeps a deleted book in `byHash`. A delete that is not a purge keeps the record, so a new import of that book already took the other path. The orphan directory in a backup, or a lost `library.json`, are the cases that change.

## The second commit

A review of the first commit found that the validation branch had no test. The second commit adds the two tests, and corrects three comments. It also corrects two statements in the first commit's message. This description holds the corrected text.

The proofread write below the new code lost its explicit `updatedAt: Date.now()`. `saveConfig` stamps that value itself, so the behaviour does not change.

## Tests

- 19 new tests: 14 in `useProgressSync.test.tsx` and 5 in `import-metahash.test.ts`. If you remove the two changes, 14 of them fail. The other five state that unchanged behaviour holds.
- Four tests examine paths that are easy to leave untested. A `saveConfig` that fails must not stop the pull. The position comparison must read the config again after the write. A value that is not a number, and a value outside the range, must not be taken.
- The full test suite shows the same failures as `upstream/main`, and 19 more tests pass.
- `tsgo --noEmit`, `biome lint` and `biome format` are clean.
- A person tested the change on a Linux build with a live account. A book whose record held a count showed a percentage before the change, and the correct pages after it.

### DIFF
--- a/apps/readest-app/src/__tests__/hooks/useProgressSync.test.tsx
+++ b/apps/readest-app/src/__tests__/hooks/useProgressSync.test.tsx
@@ -37,7 +37,7 @@ const h = vi.hoisted(() => {
     user: { id: 'u1' },
     syncConfigsMock: vi.fn(async () => {}),
     syncBooksMock: vi.fn(async () => {}),
-    saveConfigMock: vi.fn(async () => {}),
+    saveConfigMock: vi.fn(async (..._args: unknown[]) => {}),
     setViewSettingsMock: vi.fn(),
     recreateViewerMock: vi.fn(),
     cfiCompareMock: vi.fn((_a: string, _b: string) => 0),
@@ -50,7 +50,13 @@ const h = vi.hoisted(() => {
     state: {
       syncedConfigs: [] as unknown[] | null,
       progress: { location: 'cfi-loc' } as { location: string; fraction?: number } | null,
-      viewSettings: { proofreadRules: [] } as { proofreadRules: unknown[] } | null,
+      viewSettings: { proofreadRules: [] } as {
+        proofreadRules: unknown[];
+        referencePageCount?: number;
+      } | null,
+      isPrimary: true,
+      hasView: true,
+      previewMode: false,
       bookDoc: {} as unknown,
     },
     eventListeners: new Map<string, Set<(e: CustomEvent) => void>>(),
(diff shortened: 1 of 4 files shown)


bmfrosty:
Closing the PR while I update it.

PR #5736, opened 2026-08-16
feat(library): show download progress overlay on book covers

WantenMN:
Clicking a book's cloud download icon gave no visual feedback: third-party file-backend downloads (WebDAV/Drive/S3/OneDrive) ran without reporting any progress, and the queue path's initial 0% was swallowed by a || null in Bookshelf.

## Changes
- Thread byte progress from the native transfer plugin through FileSyncProvider.downloadStream → engine.downloadBookFile → runFileBookDownload; fixed the gdrive/s3/onedrive provider wrappers that dropped the onProgress argument.
- Mirror transfer-queue download progress into booksTransferProgress so Readest Cloud queue downloads show the overlay too.
- Render progress on the cover as a percentage overlay (indeterminate spinner when the backend reports no byte total), replacing the old row radial.
demo

## Summary by CodeRabbit

* **New Features**
  * Added full-cover book transfer indicators with spinners for indeterminate progress and percentages for measurable progress.
  * Added progress reporting for downloads from supported cloud and file-sync providers.
  * Transfer actions remain hidden throughout active transfers.

* **Bug Fixes**
  * Corrected display of transfers at 0% progress.
  * Improved progress cleanup after transfers complete or fail.
  * Prevented stale progress events from updating finished transfers.

### DIFF
--- a/apps/readest-app/src/__tests__/app/library/recent-shelf-select-mode.test.tsx
+++ b/apps/readest-app/src/__tests__/app/library/recent-shelf-select-mode.test.tsx
@@ -72,6 +72,7 @@ const baseProps = {
   selectedBooks: new Set<string>(),
   toggleSelection: noop,
   handleSetSelectMode: noop,
+  transferProgress: {},
 };
 
 const tapSlide = (title: string) => {
--- a/apps/readest-app/src/__tests__/app/library/useBookTransferActions.test.tsx
+++ b/apps/readest-app/src/__tests__/app/library/useBookTransferActions.test.tsx
@@ -1,9 +1,11 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { renderHook } from '@testing-library/react';
+import type { SetStateAction } from 'react';
 
 import type { Book } from '@/types/book';
 import type { EnvConfigType } from '@/services/environment';
 import type { AppService } from '@/types/system';
+import type { ProgressHandler } from '@/utils/transfer';
 
 /**
  * Issue #5062 — cloud sync providers are independently selectable, so a
@@ -69,13 +71,20 @@ const makeBook = (over: Partial<Book> = {}): Book => ({
   ...over,
 });
 
+type ProgressState = { [key: string]: number | null };
+
 const setup = (appService: AppService | null = null) => {
   const updateBook = vi.fn(async (_envConfig: EnvConfigType, _book: Book) => {});
-  const setBooksTransferProgress = vi.fn();
+  // Mimic React's functional state updates so tests can assert the actual
+  // progress transitions (indeterminate start, percentage updates, cleanup).
+  let progressState: ProgressState = {};
+  const setBooksTransferProgress = vi.fn((updater: SetStateAction<ProgressState>) => {
+    progressState = typeof updater === 'function' ? updater(progressState) : updater;
+  });
   const { result } = renderHook(() =>
     useBookTransferActions(envConfig, appService, updateBook, setBooksTransferProgress),
   );
-  return { result, updateBook };
+  return { result, updateBook, getProgress: () => progressState };
 };
 
 beforeEach(() => {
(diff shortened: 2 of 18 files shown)


chrox:
Reviewed and pushed fixes as `43bffed`. The plumbing half of this (threading `onProgress` from `tauriDownload` through `downloadStream` -> `downloadBookFile` -> `runFileBookDownload`, and the three provider wrappers that were dropping the argument) was correct as written. The issues were all in how progress reached the covers.

### The structural one

`booksTransferProgress` had two writers with no ownership marker: this hook, and the new effect mirroring transfer-queue rows into it. Each had to guess whether an entry was its own, and got it wrong three ways:

- The queue keeps completed and failed rows as history. They are persisted to localStorage and restored on launch, so any store mutation re-ran the mirror, which took its `!active` branch on a stale row and deleted progress just written for a live download of that book. Toggling the queue panel was enough to trigger it.
- `clearPending` / `clearCompleted` / `clearAll` rebuild `transfers` without the removed rows instead of transitioning their status, so the mirror never reached its delete branch for them. The entry was orphaned and the cover kept a frozen overlay with no action button until the page remounted.
- `restoreTransfers` resets an interrupted transfer to `pending` with `progress: 0`. Combined with the `|| null` -> `?? null` fix (correct on its own), a restored row painted a stuck "0%" on every launch, permanently while the queue was paused.

The mirror is gone. Queue progress is now derived where it is consumed, via `selectActiveBookDownloadProgress`, which reports only `pending`/`in_progress` rows. That makes the result independent of iteration order and of whatever history the queue holds, and a pending row reports the indeterminate sentinel rather than a frozen 0%. `Bookshelf` merges it with the hook's direct-download progress and feeds both the grid and the recently-read strip, which fixes `RecentShelf` too: it hard-coded `transferProgress={null}`, so the strip offered a Download button for a book the grid already showed downloading.

### Also fixed

- **Late progress event.** Tauri delivers `Channel` messages over IPC independently of the invoke response, so a final payload could land after `flush(); clearProgress()`, re-arm the throttle and re-create the entry that nothing would clear again. This is not theoretical: the added test reproduces it as `{ 'book-1': 90 }` surviving cleanup, which is a permanently stuck overlay with no action button. The teardown now latches.
- **Overlay and buttons disagreed.** The overlay hid at `100` but the buttons stayed hidden for any non-null value, so a book could show neither. Both now read one condition.
- **`flush()` then `clearProgress()`** emitted a state update the next updater immediately deleted. Now `cancel()`, matching what the catch path already did.
- **Direct path had no initial kick,** so `queued:false` callers (`useOpenBook`, the details modal's Redownload) stayed blank until the first byte.
- **E-ink.** `globals.css` states outright that e-ink cannot render a translucent wash, and `drop-shadow-sm` is a filter so the `[class*='shadow-']` rule does not neutralize it. The scrim now becomes a solid `base-100` panel with a 1px `base-content` border and ink-colored content under `[data-eink]`.
- **Sentinel and formula** named in `utils/transfer` instead of a bare `-1` repeated across files.
- **Throttle back to 500ms.** It now runs once per in-flight transfer rather than once for the page, so 100ms multiplied full-shelf re-renders by the batch size on a bulk download.
- **Progressbar accessible name** was the bare, untranslated book title.

### Tests

Added coverage for `selectActiveBookDownloadProgress` (including the stale-completed-row and pending-not-zero regressions) and for the late-event guard. One existing test changed behavior deliberately: priming the direct path's overlay consumes the throttle's leading edge, so the first real percentage now lands on the trailing edge. I folded that into one test covering the full lifecycle rather than dropping the assertion. Practical effect is that a download finishing in under 500ms shows only the spinner.

`pnpm test` (9194 passed), `pnpm lint` and `pnpm format:check` all clean on this branch.

Two things left alone, both deliberate: the queue panel still renders `0` where the cover renders indeterminate, since changing `transferManager`'s formula would alter the panel's semantics beyond this fix; and the indeterminate spinner still animates on e-ink, consistent with every other `loading-spinner` in the app rather than adding a one-off. I have not verified the e-ink overlay on a device.

PR #5737, opened 2026-08-16
feat(wordlens): add the en-vi gloss pack, sync incrementally

chrox:
## Summary

Adds Vietnamese as a Word Lens hint language for English books (`en-vi`), and makes `pnpm wordlens:sync` upload only the packs that changed.

`vi-en` is deliberately **not** added. Vietnamese words are multi-syllable with spaces inside them ("học sinh"), so the planner's whitespace tokenizer would gloss syllables rather than words. A `vi-en` pack needs a segmenter first, like ja/ko/th. Nothing in code gates this: the pack simply does not exist, which is how every unbuilt pair already behaves.

### en-vi pack

WikDict publishes no `en-vi.sqlite3` (404), so this pair uses the script's existing kaikki `build` mode, which reads the Vietnamese `translations` off each English Wiktionary entry. No build-script changes were needed for the pack itself.

- 9,759 entries, 10,813 inflections, 652 KB
- Lemmatized through the `en-en` table like every other English-source pack, so difficulty is gated by the lemma's rank: `whispered -> whisper`, and the derivational `reluctantly -> reluctant`
- CC-BY-SA-4.0, attributed in the pack `meta` and `ATTRIBUTION.md`

Coverage is thinner than its peers (9,759 vs 14,769 for `en-es`), concentrated away from the 4k-14k rank band. That is Wiktionary's English to Vietnamese translation coverage, not a build problem. Vietnamese Wiktionary's English section would add ~14.8k more words from the top-50k frequency list, but merging it needs a new build mode plus cleanup for trailing periods and "past participle of X" form-of glosses, so it is left as a follow-up.

### Incremental sync

The sync re-uploaded all 15 packs (~20 MB) on every run, even when one pair changed. The published manifest already carries each pack's `sha256`, so the script now fetches it and uploads only what is new or changed. Verified against the live CDN manifest: this change uploads `en-vi.json` alone, 652 KB instead of 20 MB.

- Unreachable manifest (first sync, offline) falls back to a full upload; `--force` does the same on demand, for when the manifest is current but an object was deleted from the bucket
- The manifest is no longer published when a pack upload failed. The script always claimed the manifest goes last so it never points at a missing pack, but it uploaded the manifest regardless of failures, which is the exact state that invariant exists to prevent
- `planSync` is pure and unit tested; the CLI entry point is now behind the direct-execution guard `build-wordlens-data.mjs` already uses, so importing it in tests does not run `main()`

## Testing

Tests were written before the pack and before `planSync`, and both failed first.

- `en-vi` added to `EN_SOURCE_PAIRS`, so it runs the four lemmatization-invariant checks
- New manifest test asserts `en-vi` is present and `vi-en` is not
- Six new `planSync` cases: changed sha, unseen pack, all-match, unreachable remote, `--force`, and a remote-only pack
- `pnpm test` 9205 passed / 16 skipped, `pnpm lint` and `pnpm format:check` clean

No `src-tauri/` or koplugin files changed, so the Rust and Lua checks do not apply.

## Follow-up

Packs are served from R2, not bundled, so `WORDLENS_R2_BUCKET=<bucket> pnpm wordlens:sync` still needs to run after merge before clients can see `en-vi`.

## Summary by CodeRabbit

* **New Features**
  * Added English-to-Vietnamese Word Lens support with published pack metadata.
  * Added documentation for generating Vietnamese glosses and supported language direction.

* **Improvements**
  * Word Lens synchronization now uploads only changed packs and safely publishes manifests after successful uploads.
  * Added support for forced and fallback full synchronization.

* **Tests**
  * Expanded coverage for synchronization scenarios and Word Lens pack availability.


xxyzz:
Hi @chrox, please notice the kaikki.org file you're currently downloading is not supposed to be used, users should download the "raw data" files. For English Wiktionary, [this file](https://kaikki.org/dictionary/raw-wiktextract-data.jsonl.gz) should be download. But it has all languages so you'll need to filter the English words by checking the "lang_code" field.

And they have [Vietnamese Wiktionary data](https://kaikki.org/viwiktionary/raw-wiktextract-data.jsonl.gz) so you could use that.

chrox:
> please notice the kaikki.org file you're currently downloading is not supposed to be used

Any reasons?

xxyzz:
It's a file generated for the website itself not meant to be used by others, the reason is explained on the website after the download link https://kaikki.org/dictionary/English/ and the bold "DEPRECATED" link to the wiktextract issue https://redirect.github.com/tatuylonen/wiktextract/issues/1178

chrox:
Thanks, we will fix it later.

PR #5765, opened 2026-08-17
fix(opds): drop the webview Origin header on native requests

### LINKED ISSUE #5698: Komga OPDS 403
### Bug Description

Komga OPDS always 403's.
No logs on Komga side, which means no actual auth attempt happened, and Readest's OPDS support is utterly broken.

### Steps to Reproduce

1. Connect to Komga

### Expected Behavior

Work

### Reproduction File

_No response_

### Operating System

Linux, Android

### OS Version

Android 17

### Readest Version

Whatever the newest version is, deleted the app as this doesn't work, would reinstall AFTER it is fixed.

### Custom CSS

_No response_

NoaHimesaka1873:
Remove CORS origin for OPDS requests as this causes problem for Komga w/ HTTPS. It's always CORS, the most cursed dark magic of the HTTP.

Fixes #5698, which was wrongfully closed since @chrox was testing on local, non-HTTPS environment which can't reproduce this issue. Reproduction of the issue REQUIRES HTTPS setup as it stems from CORS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OPDS requests to prevent unwanted `Origin` headers on supported platforms.
  * Preserved explicitly configured custom `Origin` headers.
  * Applied consistent header handling to authentication checks and streamed OPDS pages.

* **Tests**
  * Added coverage for platform-specific, custom-header, and authentication probe scenarios.

### DIFF
--- a/apps/readest-app/src/__tests__/utils/opds-req.test.ts
+++ b/apps/readest-app/src/__tests__/utils/opds-req.test.ts
@@ -234,6 +234,16 @@ describe('opdsReq', () => {
       expect(res.status).toBe(200);
     });
 
+    it('does not add an Origin header to web-platform requests', async () => {
+      fetchMock.mockResolvedValue(makeResponse({ status: 200, body: '<feed/>' }));
+
+      await fetchWithAuth('https://komga.example.com/opds/v2/catalog', 'alice', 's3cret', true);
+
+      const init = fetchMock.mock.calls[0]![1] as RequestInit;
+      const headers = init.headers as Record<string, string>;
+      expect('Origin' in headers).toBe(false);
+    });
+
     it('retries with Digest auth when the server issues a Digest challenge', async () => {
       fetchMock
         .mockResolvedValueOnce(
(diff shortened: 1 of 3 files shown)


chrox:
Two things worth noting, because the distinction changes how we document
and reproduce this.

**The header is the cause; CORS is not doing anything to us.** CORS is enforced
by a *browser*, on behalf of a page, to decide whether that page is allowed to
read a cross-origin response. Its job is protecting a user's ambient
credentials from script they never authorized. Readest's native requests do not
pass through a browser at all: `@tauri-apps/plugin-http` hands them to Rust's
reqwest, which has no CORS, no preflight, and no origin checking. So there is
no CORS being applied to us here.

What happens is narrower. plugin-http emulates browser `fetch` semantics, and
part of that emulation is stamping an `Origin` header on every request
(`commands.rs:290-301`). Something on your side (your NGINX) then applies CORS *rules* to
that header and refuses. The bug is that a browser origin is sent from a client
that is not a browser. Removing it is correct not because CORS is cursed, but
because `tauri://localhost` is a meaningless claim for a native client to make.
Every other native OPDS client sends no Origin either.

**HTTPS is not the variable.** The Origin is derived from the webview's own URL,
not from the target, so it is stamped identically over HTTP and HTTPS. What
decides whether you see the 403 is whether something in front of the catalog (your NGINX)
inspects `Origin` and rejects an unrecognized value. The reverse proxy is the
load-bearing detail, and it is not in the issue's Steps to Reproduce ("1.Connect to Komga"). 
"Nginx returns 403, Komga has no access log" means the request never reached Spring, so Spring's
CORS processor cannot be what refused it. The patch's comment currently
attributes the 403 to "Spring-based servers such as Komga", which I have
corrected on the branch.

chrox:
Vanilla nginx does not look at Origin at all. It has no CORS logic built in. So a 403 keyed on that header means something was added on top. Ranked by likelihood:

1. A copy-pasted CORS snippet with a deny-by-default map. This shape is all over self-hosting guides:
map $http_origin $cors_origin {
    default "";
    "~^https://(app|read)\.example\.com$" $http_origin;
}
# ...
if ($cors_origin = "") { return 403; }
1. Any unrecognized Origin becomes a hard 403 before proxy_pass runs, which matches the reported symptom exactly: nginx answers, Komga never sees the request.
2. ModSecurity / OWASP CRS in blocking mode. A tauri:// scheme in Origin is the kind of protocol anomaly the CRS flags, and it returns 403 by default.

Is it a bad configuration? Mostly yes, for a specific reason: returning 403 based on Origin is a category error. CORS is a browser-side read restriction. The server's job is to advertise policy in Access-Control-Allow-Origin and let the browser enforce it. Rejecting the request outright buys no security, because any non-browser client can send whatever Origin it wants or none at all. It blocks nothing an attacker would do, and breaks every legitimate non-browser client. Real cost, no benefit.

@NoaHimesaka1873 Please check your NGINX configuration and update your understanding of what CORS is and for. And that's why I cannot reproduce it.

NoaHimesaka1873:
@chrox No CORS-related configuration on NGINX side, I've discovered that it's Spring's (which Komga uses) CORS handling, which leaves no logs on Komga.
I found out that it rejects any request which isn't allowed by the configuration, and I've set `KOMGA_CORS_ALLOWED_ORIGINS` variable, so that's causing it! (and it returns `Invalid CORS request`)

PR #5768, opened 2026-08-17
fix(tts): allow chapters sharing sentences with earlier packs to download

### LINKED ISSUE #5767: TTS: download marks a chapter as failed even though its audio was fully downloaded
### Bug Description

When downloading a book for offline TTS playback, individual chapters intermittently end up in the failed state with the message "Download incomplete", even though most or all of the chapter's audio actually made it onto the device and is playable. This tends to happen in books with a lot of repeated or short sentences, such as dialogue-heavy novels. The download appears to run quickly and "do nothing", then fails again with the same message every time. The chapter stays failed permanently and can only be worked around by clearing the download and starting over, with no guarantee the same chapter will not fail again.

### Steps to Reproduce

Download chapters with known repeated sentences

### Expected Behavior

If a chapter's audio is present and usable, downloading it should report success and make the chapter available for offline playback, so it can be listed and played back offline like every other successfully downloaded chapter. Repeated sentences that also appear in other chapters should not cause a chapter to fail or block it from being marked as downloaded.

### Reproduction File

_No response_

### Operating System

Android

### OS Version

Android 16, Android 14

### Readest Version

debug build of  89821136fb368ecedca2194718b59842aec35c08

### Custom CSS

_No response_

gfreitash:
closes #5767

### Problem
Downloading a book for offline TTS playback intermittently marks individual chapters as failed with "Download incomplete", even though most or all of that chapter's audio is present on the device and playable. A failed chapter could never succeed on retry: retrying re-ran quickly and "seemingly did nothing", then failed identically. The failure hit books with many repeated short sentences, especially dialogue-heavy novels.

### Root cause
The offline cache is content-addressed by sentence key (a hash of provider, language, voice, pitch, and text), so identical sentences share one entry. When an earlier chapter packed a sentence, that entry became pack-backed with its audio moved into the earlier chapter's pack file. A later chapter containing the same sentence referenced that already-packed entry, and compaction's completeness check demanded the entry still carry loose audio, so the later chapter could never form its own pack and failed forever, even though every sentence was cached. The audio was all there; it just lived inside another chapter's pack. The two secondary bugs were an empty-manifest chapter never being reported as packed (so cover and blank pages always failed) and warmSentence having no retry (a single transient network failure dropped a sentence and failed a chapter).

### Fix
Compaction's completeness check now only requires the entry row to exist rather than loose audio. When packing a section, pack-backed entries are resolved through the pack that owns them and a copy is embedded into the new section's pack, so each pack stays a self-contained, playable MP3 and repeated sentences are copied locally pack-to-pack rather than re-synthesized or re-fetched. Empty-manifest sections are now reported as packed so cover and blank chapters complete, and warmSentence reuses the existing synthesis retry path so a transient network blip no longer fails a chapter.

Pack deletion is now robust in the other direction as well: when eviction or
self-heal removes a pack, entries still referenced by another section through
`manifest_marks` are transferred to a surviving current pack of that section
(fingerprint-gated, pinned-first) instead of being destroyed, keeping pinned
chapters playable offline. Entries with no surviving candidate detach to a
loose miss and re-synthesize on next read — the same bounded recovery as the
existing "Download incomplete" path, explicitly accepted per the review
thread. Pins and the `packed` status flag are now fingerprint-aware, so stale
packs of a pinned section are reclaimable while `imported-%` packs stay
protected.

### Verification
Added four regression tests covering the shared-key eviction transfer
(shared sentence survives a later pack's eviction), stale-pack reclamation
for pinned sections, corrupt-pack self-heal transferring instead of
destroying, and imported-pack survival. Added earlier tests for a section
sharing a sentence key with an already-packed section and for empty-manifest
sections reported as packed; all fail against the old behavior and pass with
the fix. The pack-compaction suite (31 tests) and pnpm lint are green.

## Summary by CodeRabbit

- **New Features**
  - Text-to-speech downloads can now be cancelled while audio is being prepared.
  - TTS cache metadata is now read and validated across supported storage options.

- **Bug Fixes**
  - Improved recovery when cached audio is missing, corrupted, or removed.
  - Preserved shared audio entries during cache cleanup and pack eviction.
  - Improved retention of imported and pinned audio packs.
  - Improved handling of empty or rebuilt cache manifests.
  - Sections with incomplete audio are now reported as skipped instead of completed.
  - Added clearer warnings when downloads finish with incomplete sections.

### DIFF
--- a/apps/readest-app/src/__tests__/services/tts-book-cache-store.test.ts
+++ b/apps/readest-app/src/__tests__/services/tts-book-cache-store.test.ts
@@ -19,6 +19,11 @@ const packState = vi.hoisted(() => {
         if (!data) throw new Error('missing pack');
         return data.slice(offset, offset + length).buffer as ArrayBuffer;
       }),
+      readSidecar: vi.fn(async (name: string) => {
+        const data = files.get(name);
+        if (!data) return null;
+        return JSON.parse(new TextDecoder().decode(data));
+      }),
       remove: vi.fn(async (name: string) => {
         files.delete(name);
       }),
--- a/apps/readest-app/src/__tests__/services/tts-downloader.test.ts
+++ b/apps/readest-app/src/__tests__/services/tts-downloader.test.ts
@@ -90,6 +90,10 @@ describe('TTSDownloader', () => {
     // All three were attempted; two succeeded.
     expect(warmer.warmed).toHaveLength(3);
     expect(result.synthesized).toBe(2);
+    // A section with a failed sentence is reported skipped, not completed: the
+    // manifest gap means it has not fully downloaded.
+    expect(result.completed).toEqual([]);
+    expect(result.skipped).toEqual([0]);
     // A section with a failed sentence is still compacted (the pack simply
     // won't form until the gap fills), so a later retry can complete it.
     expect(warmer.compacts).toBe(1);
--- a/apps/readest-app/src/__tests__/services/tts-pack-compaction.test.ts
+++ b/apps/readest-app/src/__tests__/services/tts-pack-compaction.test.ts
@@ -28,6 +28,11 @@ class FakePackFs implements TTSPackFs {
     if (!data) throw new Error(`ENOENT: ${name}`);
     return data.slice(offset, offset + length).buffer as ArrayBuffer;
   }
+  async readSidecar(name: string): Promise<TTSPackSidecar | null> {
+    const data = this.files.get(name);
+    if (!data) return null;
+    return JSON.parse(new TextDecoder().decode(data)) as TTSPackSidecar;
+  }
   async remove(name: string): Promise<void> {
     this.files.delete(name);
   }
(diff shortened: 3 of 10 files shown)


gfreitash:
I turned this into a draft while I study this point CodeRabbit made:
> The change fixes repeated-sentence downloads and retry behavior, but shared audio can still become unavailable for an earlier offline chapter if a later pack containing the same sentence is deleted. Merge should wait for pack membership to be preserved independently per pack, or for this bounded data-availability risk to be explicitly accepted.

Turns out this is a much bigger issue than it appears. I will elaborate on it soon

gfreitash:
Follow-up: the deletion risk CodeRabbit flagged is real — it's the mirror
of the bug this PR fixes. Pack *creation* is now safe for shared sentences,
but pack *removal* still wasn't: a sentence key has one entry row, so LRU
eviction — and self-heal of a corrupt pack file — deleted that row for
every chapter, even though the bytes still lived in another chapter's
pack. A pinned chapter could lose audio it had offline.

429beaa86 closes it with no schema change. The cache already records every
section that uses a sentence (`manifest_marks`), so on pack removal the
entry is repointed to a surviving pack of any of those sections — the audio
is copied pack-to-pack, never re-downloaded. If no other pack holds the
sentence, the entry falls back to the existing "Download incomplete" repair
state and re-synthesizes on the next read; that bounded case is the
accepted half of the note, and it's now covered by tests. Pins became
fingerprint-aware too, so stale packs of a pinned section can be reclaimed
with the same transfer.

Ready for review.

gfreitash:
I've found some regressions related to clearDownloads leaving leftovers, I will work on them later. I will use this build for some time to see if I can identify other regressions.

chrox:
Pushed `61c921542` with six hardening fixes on the new code paths, each written test-first (the four new tests fail on the previous head):

1. **compact() excludes detached entries** (`audio IS NULL AND pack_id IS NULL`) from completable sections. A section healed by transfer-with-no-candidate previously stayed eternally completable: every debounce cycle loaded the whole section's audio blobs, hit the detached row, warned `pack audio missing`, and failed, forever until re-synthesis.
2. **warmSentence returns false when the abort signal cancels synthesis.** `#synthesizeWithRetry` resolves `undefined` on abort rather than throwing, so a cancelled warm recorded a mark and reported success for a sentence that never synthesized.
3. **clearDownloads GC passes cleared keys as one `json_each(?)` array.** A fully downloaded long book binds one host parameter per distinct sentence key and can exceed SQLite's 32766 cap, making Clear downloads throw exactly when the cache is largest.
4. **The transfer UPDATE re-validates each candidate pack still exists** inside the transaction. Candidate selection and sidecar reads run outside it (file I/O), so an interleaved eviction could repoint entries at a pack row that no longer exists.
5. **#readPackedAudio detaches entries whose pack row vanished.** A dangling `pack_id` was a permanent silent miss that neither eviction branch counted; now it degrades to a loose miss and repopulates on the next synthesis.
6. **readSidecar candidates with non-numeric offset/length are dropped.** `JSON.stringify` silently drops `undefined`, so a corrupt sidecar committed `pack_offset NULL` on a live row, another permanent miss.

**Device verification (Xiaomi 13, release build of this branch)** with a purpose-built EPUB where three chapters share five identical sentences plus a heading-only chapter: Download all reaches 4 of 4 (the shared-sentence chapters and the empty chapter all complete); offline playback works with wifi and data disabled, both the pack-MP3 timeline and the live per-sentence cache path (CDP console confirms every chunk scheduled from cache, no synthesis attempts); Clear all leaves 0 of 4 with no partly-downloaded leftovers; Download all again returns to 4 of 4.

**One open issue found on device, likely for a follow-up**: an empty or symbol-only section flips from Downloaded back to not-downloaded once live TTS visits it. The live speak path registers a different (empty) mark list than the downloader's enumeration did, the manifest fingerprint changes, and the now fingerprint-gated `packed` check correctly reports the old pack stale. It self-heals on the next Download all (re-registers and repacks instantly), and pre-PR it was masked because any pack counted as packed. The clean fix is making the two enumeration paths agree for sentence-less sections rather than loosening the fingerprint gate.

Issue #4668, opened 2026-06-19
FR: Editing annotations (notes) in place


edo-jan:
In order to read an existing annotation (specifically, a text note as implemented in 0.11.10) one would need to click on the little bubble indicating the note. This action displays the text note content in an overlay floating window. To edit the content one would need to go over the following steps:
1. Click on the note icon (opens the note content viewer)
2. Click on the note content viewer (opens the annotation pane on the left)
3. Select the correct annotation from the pane (not a trivial task sometimes if there are several)
4. Mouse over the annotation to reveal the "edit icon" (pencil)
5. Click the "edit" icon (notebook pane on the right is open)
6. Make your edits and save

**Describe the solution you'd like**
Enable editing the note in place (in the note content viewer, transforming it into content editor by a simple click on the edit icon). 

> A clear and concise description of what you'd like to happen.
1. Click on the note icon (opens the note content viewer)
2. Click on edit icon at the bottom of the note content viewer to switch to the edit mode
3. Make edits and save  

**Describe alternatives you've considered**
This feature further extends #3870 partially implemented in #4647 

<img width="947" height="369" alt="Image" src="https://github.com/user-attachments/assets/38d841c3-cec2-4477-adea-9541fbadc0ea" />

libbybar:
Hi! I'm planning to work on this issue.

I've looked through the existing inline editing flow in `BooknoteItem` and the read-only popup in `AnnotationNotes`. My current plan is to extract the shared note-saving logic and reuse the existing inline editing flow in the popup instead of duplicating it.

Does that sound like the direction you had in mind?

edo-jan:
@libbybar - Great timing! I was actually planning to update this post, as I’ve discovered a couple of related UX issues that could be addressed at the same time.

To give some context, the issue depends on the user's starting point. Here is a breakdown of the current workflow and where it breaks down.

✅ **The Current Workflow (Creating an Annotation)**
This part works well:

1. User selects text and chooses "Highlight" or "Annotate" from the context menu.
2. If "Annotate" is selected, the note editor opens in the right pane (Notebook).
3. User enters the note and clicks "Save" to close the pane.

**❌ The Problem (Editing an Existing Annotation)**
When trying to edit an existing note, the workflow becomes highly confusing due to two different routes:

**Route A: Via the tiny annotation icon**

1. User clicks the tiny annotation icon, which opens a read-only note in a popup.
[...]

PR #5780, opened 2026-08-18
feat(reader): add inline note editing

libbybar:
## Summary

Adds in-place editing for existing annotation notes directly from the note popup, addressing #4668.

The existing inline editing flow in `BooknoteItem` was first refactored into shared, reusable pieces so the popup can use the same save behavior without duplicating logic.

## What changed

* Extracted shared inline text editing state into `useInlineTextEditor`.
* Extracted shared note persistence into `useSaveBooknoteNoteText`.
* Added a pure `updateBooknoteNoteText` utility for immutable note updates.
* Updated `BooknoteItem` to use the shared editing flow without changing existing behavior.
* Added inline Edit / Save / Cancel support to annotation note popups.
* Made the popup edit action always visible instead of hover-only, improving discoverability and touch accessibility.
* Preserved existing note bubble behavior when adding or clearing note text.

## Validation

* 38/38 relevant tests passing after rebasing onto the latest `upstream/main`.
* Type-check passes.
* Biome lint passes for all touched files.
* Manual verification:

  * create a note
  * edit it from the popup
  * clear its text to remove the note
  * close and reopen the book and confirm the persisted state

## Follow-up

I also investigated the two related routing/state issues discussed in #4668. I kept this PR focused on the original in-place popup editing request so the change remains easy to review; those can be handled separately.

## Summary by CodeRabbit

- **New Features**
  - Added inline editing for annotation and bookmark notes.
  - Save or cancel changes directly from note displays.
  - Annotation notes support clearer navigation, mobile dismissal, timestamps, sanitized Markdown, and recent-update sorting.
  - Successful edits refresh note bubbles and save settings automatically.

- **Bug Fixes**
  - Improved handling of blank, missing, deleted, or unsuccessful note updates.
  - Prevented failed edits from changing displayed notes or saved data.

- **Tests**
  - Expanded coverage for note editing, saving, sorting, cancellation, formatting, and persistence.

### DIFF
--- a/apps/readest-app/src/__tests__/app/reader/components/sidebar/BooknoteItem.test.tsx
+++ b/apps/readest-app/src/__tests__/app/reader/components/sidebar/BooknoteItem.test.tsx
@@ -16,7 +16,13 @@ const mocks = vi.hoisted(() => {
     setNotebookEditAnnotation: vi.fn(),
     addAnnotation: vi.fn(),
     saveConfig: vi.fn(),
-    updateBooknotes: vi.fn(() => ({ booknotes: state.booknotes })),
+    // Mirrors the real store: `updateBooknotes` writes back whatever array
+    // it's called with (production code now returns a new array from
+    // `updateBooknoteNoteText` instead of mutating the existing one).
+    updateBooknotes: vi.fn((_key: string, booknotes: typeof state.booknotes) => {
+      state.booknotes = booknotes;
+      return { booknotes: state.booknotes };
+    }),
   };
 });
 
--- a/apps/readest-app/src/__tests__/app/reader/hooks/useInlineTextEditor.test.ts
+++ b/apps/readest-app/src/__tests__/app/reader/hooks/useInlineTextEditor.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it, vi } from 'vitest';
+import { act, renderHook } from '@testing-library/react';
+import { useInlineTextEditor } from '@/app/reader/hooks/useInlineTextEditor';
+
+describe('useInlineTextEditor', () => {
+  it('starts outside edit mode with an empty draft', () => {
+    const { result } = renderHook(() => useInlineTextEditor(vi.fn()));
+
+    expect(result.current.inlineEditMode).toBe(false);
+    expect(result.current.draftText).toBe('');
+  });
+
+  it('entering edit mode seeds the draft with the given initial text', () => {
+    const { result } = renderHook(() => useInlineTextEditor(vi.fn()));
+
+    act(() => result.current.startEdit('existing note'));
+
+    expect(result.current.inlineEditMode).toBe(true);
+    expect(result.current.draftText).toBe('existing note');
+  });
+
+  it('typing updates the draft while in edit mode', () => {
+    const { result } = renderHook(() => useInlineTextEditor(vi.fn()));
+    act(() => result.current.startEdit('old'));
+
+    act(() => result.current.setDraftText('new'));
+
+    expect(result.current.draftText).toBe('new');
+  });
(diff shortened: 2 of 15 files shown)


chrox (commented):
Thanks for the clean split into `useInlineTextEditor` / `useSaveBooknoteNoteText` / `updateBooknoteNoteText` — the popup editor reuses the sidebar save path exactly, and I could not find a behavior change in `BooknoteItem`.

I reviewed the code and verified the feature end to end on web (Chrome) and on Android (Xiaomi 13, release APK built from this branch): tap bubble → Edit → soft keyboard opens with the card still fully visible above it → type → Save updates the card, the sidebar and the stored config, and the text survives a reload. Clearing the text removes the bubble and the popup falls back to the annotation toolbar; the highlight stays, as intended.

**CI: `test_web_app (1)` was failing because of this PR**, not flakiness. `annotation-popup-layout.browser.test.tsx` mocks `annotatorUtil` with a factory that only exports `getHighlightColorLabel`. `AnnotationPopup → AnnotationNotes → AnnotationNoteItem → useSaveBooknoteNoteText` now imports `applyNoteBubbleTransition` / `decideNoteBubbleTransition` from that module, and a browser-mode `vi.mock` is a strict ES module, so the whole test file failed to import ("does not provide an export named 'applyNoteBubbleTransition'"). I pushed a one-commit fix to your branch that stubs both names in the mock factory (`1e64c00a8`); the browser test passes again locally (3/3), and the full unit suite + lint are green.

On the CodeRabbit comments, for the record:
- "Clearing a note leaves an empty card with no way back" is not accurate: `Annotator.tsx` rebuilds `annotationNotes` from `config.booknotes` filtered to non-blank notes, so the cleared note unmounts and the popup shows the toolbar instead (verified on device). No change needed.
- Keeping `Cancel`/`Save` as two `TextButton`s is fine here — it matches the sidebar `BooknoteItem` editor, and the popup card already gets the e-ink border from `.popup-container`.
- The `role='none'` card and "save closes edit mode before persisting" are both pre-existing patterns carried over from the old `AnnotationNotes` / `BooknoteItem`; happy to leave them out of this PR.

One unrelated thing I noticed while testing (pre-existing on `main`, not caused by this PR): an existing note's bubble is not drawn on initial load or after a section re-render — `onCreateOverlay` only re-adds `style` annotations, and the bubble only comes from the `progress` effect. I'll track that separately.

chrox:
Rebased the branch onto current `main` (force-pushed, so please `git fetch` + `git reset --hard origin/feature/inline-note-editing` — or your equivalent — before committing more here).

The conflict came from #5805 (`841b3639b`), which landed first and changed the same note-card markup to render the note through the shared `parseNoteMarkdown` (Marked + KaTeX + DOMPurify) instead of a bare text node. Resolution:

- `BooknoteItem.tsx`: kept main's `parseNoteMarkdown` import, dropped the now-unused `applyNoteBubbleTransition` / `decideNoteBubbleTransition` imports (they live in `useSaveBooknoteNoteText`).
- `AnnotationNotes.tsx`: kept this PR's delegation to `AnnotationNoteItem`.
- `AnnotationNoteItem.tsx`: ported main's render — `noteHtml = useMemo(() => parseNoteMarkdown(note.note))` and a `prose prose-sm max-w-none` `dangerouslySetInnerHTML` block in place of `{note.note}` — so the popup card still shows Markdown/KaTeX exactly like the sidebar.
- `AnnotationNotesMarkdown.test.tsx` (from #5805) now renders through `AnnotationNoteItem`, so it got the same store/translation mocks as `AnnotationNotesSurface.test.tsx`.

Targeted tests, lint, and the popup browser test pass locally; CI will confirm.

PR #5908, opened 2026-08-27
feat(tts): lyric-style sentence view in the Read Aloud player (#5755)

### LINKED ISSUE #5755: FR: FEATURE REQUEST- TTS UI/UX Improvement

Image

This is one of the best looking UI of tts. It really helps to focus one the text while listening to the novel. Its kind of a mixture of readest paragraph reading and reading ruler.

_No response_

### Additional Context

_No response_

chrox:
Closes #5755.

Draws the chapter as a scrollable transcript in the Read Aloud player: the sentence being spoken sits centred and lit, the rest dim around it, and the sheet follows the voice the way a music player follows a song. Scope is the lyric highlighting from the issue, not the rest of the referenced app's player.

### When it is offered

Gated on the engine actually aligning audio to the text (`TTSController.supportsLyrics()` = `mediaClock && textHighlight !== false`):

| Engine | Lyric sheet |
| --- | --- |
| Edge TTS (`BufferedTTSClient`) | yes |
| EPUB 3 Media Overlays with per-sentence timing | yes |
| Paired audiobook (chapter-level timing only) | no, keeps the cover player |
| Native / Web Speech (no media clock) | no, keeps the cover player |

A section with nothing to transcribe, or a whole book imported as one section (`LYRIC_MAX_LINES`), falls back the same way. Because capabilities change with the voice, `supportsLyrics` is state synced by `syncClientCapabilities()` rather than a render-time probe, so switching to a book's own narrator redraws the player.

### Seeking

Dragging the sheet parks it and raises a row over the centre line: page number on one side, play button on the other. Playback does not move until that button is pressed, and an abandoned drag slides back to the voice after four seconds. The page number goes through the same CFI to location path the footer uses, so it matches the number the reader left behind.

### Buffering

While an utterance has been handed to the engine but no audio is out yet, the seek button spins and the transport buttons (player sheet, and both mini player styles) wear a ring. A ring rather than a swapped icon, so a reader can still pause mid-fetch. `isBuffering()` is driven by the first event the `speak()` iterator yields, which for both buffered clients is the first audible chunk; `getChunkPosition()` is not usable for this, since the web player reports a position for chunks scheduled in the future.

### Notes

- The sentence list is the `SectionTimeline` the scrubber already builds, so there is no new enumeration and no second source of truth.
- Line geometry is measured per layout and re-measured whenever the content box changes. A narrower window re-wraps every sentence and moves every centre while the scroller's own height stays put, so watching the viewport alone left the follow scroll a line or two out.
- Long moves land outright and neighbouring sentences glide; e-ink never animates. Sentences taller than the window top-align instead of centring, so their opening words are not scrolled past.

### Verification

`pnpm test`, `pnpm lint` and `pnpm format:check` all green. Checked in Chrome against Moby-Dick with Edge TTS: the sheet follows the spoken line, a wheel scroll raises the seek row, pressing play moved playback back onto that exact line, and the spinner and ring both appear while a far uncached sentence synthesises and clear when audio arrives. Cached sentences correctly show no indicator. Not yet checked on a device or on e-ink.

## Summary by CodeRabbit

- **New Features**
  - Added a lyric-style transcript view to Read Aloud, highlighting the currently spoken sentence.
  - Added scrolling, page lookup, and playback controls for starting narration from a selected lyric.
  - Added buffering indicators to playback controls, including e-ink-friendly behavior.
  - Added support detection and graceful fallback when transcripts are unavailable.

- **Tests**
  - Added comprehensive coverage for lyric rendering, navigation, playback, buffering, scrolling, and transcript handling.

### DIFF
--- a/apps/readest-app/src/__tests__/components/tts/TTSLyricsView.test.tsx
+++ b/apps/readest-app/src/__tests__/components/tts/TTSLyricsView.test.tsx
@@ -213,6 +213,30 @@ describe('TTSLyricsView', () => {
     expect(defaults.onGetLyricPage).toHaveBeenCalledWith(3);
   });
 
+  test('drops a carried-over pick when playback crosses into a new chapter', () => {
+    const { rerender } = renderView({ activeIndex: 0 });
+    dragTo(3);
+    expect(screen.getByLabelText('Play')).toBeTruthy();
+
+    // New chapter, new ordinals: pressing play on index 3 of the old sheet
+    // would seek to whatever sentence sits there now.
+    rerender(
+      <TTSLyricsView {...defaults} lines={['Fresh chapter.', 'Second line.']} activeIndex={0} />,
+    );
+    expect(screen.queryByLabelText('Play')).toBeNull();
+  });
+
+  test('drops a committed line when the chapter changes before its audio lands', async () => {
+    const { rerender } = renderView({ activeIndex: 0 });
+    dragTo(2);
+    fireEvent.click(screen.getByLabelText('Play'));
+    expect(document.querySelector('.loading-spinner')).toBeTruthy();
+
+    rerender(<TTSLyricsView {...defaults} lines={['Fresh chapter.']} activeIndex={0} buffering />);
+    // Otherwise it spins against an ordinal the new chapter will never report.
+    await waitFor(() => expect(document.querySelector('.loading-spinner')).toBeNull());
+  });
+
   test('the play button reads from the dragged line, not the spoken one', () => {
     renderView({ activeIndex: 0 });
     dragTo(2);
--- a/apps/readest-app/src/__tests__/components/tts/TTSLyricsView.test.tsx
+++ b/apps/readest-app/src/__tests__/components/tts/TTSLyricsView.test.tsx
@@ -1,0 +1,260 @@
+import React from 'react';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { act, cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react';
+
+vi.mock('@/hooks/useTranslation', () => ({
+  useTranslation: () => (key: string, opts?: Record<string, unknown>) =>
+    opts ? Object.entries(opts).reduce((s, [k, v]) => s.replace(`{{${k}}}`, String(v)), key) : key,
+}));
+
+const dispatchToast = vi.fn();
+vi.mock('@/utils/event', () => ({
+  eventDispatcher: {
+    dispatch: (name: string, detail: unknown) => dispatchToast(name, detail),
+    on: vi.fn(),
+    off: vi.fn(),
+  },
+}));
+
+import TTSLyricsView from '@/app/reader/components/tts/TTSLyricsView';
+
+// jsdom has no layout engine: every offset is 0, so the geometry the view
+// measures has to be supplied. Lines are 40px tall, stacked under the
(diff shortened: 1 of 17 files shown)


chrox:
Thanks, all three were real. Verified each against the code before changing anything; fixed in the follow-up commit.

**1. Reset lyric seek state when the transcript changes** (`TTSLyricsView.tsx`): confirmed. `seekIndex` and `pending` survived a chapter turn, so the play button could seek to whatever sentence happened to sit at that ordinal in the new chapter, and a carried-over commit would spin until its 15s watchdog rather than resolving. Added an effect keyed to `lines` that clears `seekIndex`, `pending`, `page`, `seekingRef` and the idle timer. Two regression tests, both confirmed to fail with the effect disabled.

**2. Handle lyric retrieval errors as unavailable content** (`useTTSLyrics.ts`): confirmed, and reachable, since `getLyrics()` awaits `ensureTimeline()`, which dynamically imports `foliate-js/tts.js`, so a chunk load failure rejects. The rejection was unhandled and `unavailable` stayed false, leaving the lyric layout over an empty list. Now caught, clearing `lyrics` and setting `unavailable`, so it falls back to the cover player. Regression test added, confirmed to fail without the catch.

**3. Prevent a stale timeline from being committed during lyric loading** (`TTSController.ts`): confirmed. `doc` was captured before the await but `#timelineSectionIndex` was read from live state after it, so a chapter turning inside the enumeration would file the old document's sentences under the new section's index. Every `#timelineSectionIndex === #ttsSectionIndex` guard would then wave it through, and `seekToLyric` could hand a cross-document range to `#resumeAt`. The section index is now pinned before the await and the build is discarded if either the index or the document moved; `registerSectionManifest` and the duration hydration use the pinned index too. This mirrors the guard `#hydrateTimelineDurations` already had.

Note this one predates the PR (`ensureTimeline` has backed the scrubber for a while); the PR widened the exposure by adding `seekToLyric` as a consumer.

I did **not** add the regression test you suggested for #3. Reproducing it deterministically needs the section to change synchronously inside the enumeration, and there is no public synchronous path that mutates `#ttsSectionIndex` (`initViewTTS` no-ops once a section is loaded, and every other caller of `#initTTSForSection` is private and awaits first). The only way to write it would be a test-only seam in production code, which is a worse trade than the three-line guard. Flagging that plainly rather than adding a test that does not exercise the race it is named for.

`pnpm test`, `pnpm lint` and `pnpm format:check` are green.

chrox:
The out-of-diff one is valid too, fixed in the follow-up.

**Queue the section reload when a lyric request is active** (`useTTSLyrics.ts`): confirmed. The `loadingRef` guard dropped the reload outright. The in-flight request then resolves against whatever the controller's live section is by the time `ensureTimeline()` returns, so the sheet was left showing the previous chapter's lines, or falling back to the cover if the new pinning guard discarded the build. It did self-heal, but only on the next `tts-position` event, which is a sentence away.

`loadLyrics` now records the request and drains it in a loop once the active fetch settles. The loop is bounded by the number of changes that actually queued: each pass clears the flag before fetching, so only a change arriving during that fetch runs another pass.

Regression test added (unresolved first fetch, section change, then release) and confirmed to fail without the queueing.

`pnpm test`, `pnpm lint` and `pnpm format:check` green.

chrox:
Heads up: this merged just before the last review fix could be pushed onto it. The comment above describes a fix that is **not** in this PR; it now lives in #5909 (cherry-picked cleanly onto the merged main).

Issue #5296, opened 2026-07-24
Move book-related settings to reader view


dastarruer:
All Font, Layout, and some Theme settings (reading ruler, highlight colors and code highlighting) all only apply in the reader view. Thus, being able to change them in the global settings panel makes no sense, as if the user is in the library view, they can't see the changes visually. 

Likely, a better solution would be to move these to the right sidebar, like how kindle handles it:

<img width="1752" height="918" alt="Image" src="https://github.com/user-attachments/assets/8ece4dfd-fcdb-4c76-91d5-2654dff9eea5" />

It would give an actual purpose to the 'Font and Layout' button in the ribbon and reduce complexity in the global settings panel.

chrox:
Kindle is largely a really bad example of UI/UX, it hardly can keep text and icons align well. We will copy more from Apple Books.

dastarruer:
I guess another solution would be to move all non reader settings into a separate page instead of putting them in the modal. It might actually require less work to do that instead.

dastarruer:
> I guess another solution would be to move all non reader settings into a separate page instead of putting them in the modal. It might actually require less work to do that instead.

But @chrox, how does this sound? This is likely simple enough that even I could do it, and out of the two ideas I’ve presented, it seems the best. I like the current floating modal, so keeping it and moving everything else to a dedicated page could perhaps work.

chrox:
> I like the current floating modal, so keeping it and moving everything else to a dedicated page could perhaps work.

Yes, this might work, at least for the Settings - Behavior - Device, or most of the Behavior settings. At the moment, the settings are already too much for this tab. But we might need a list first to confirm the migration settings.

chrox:
<img width="516" height="978" alt="Image" src="https://github.com/user-attachments/assets/96e2cc49-90b1-470b-9fe0-37f77a304b1d" />

<img width="810" height="732" alt="Image" src="https://github.com/user-attachments/assets/fa463fb0-10b1-47f7-978f-894361c9230b" />

The quick settings popup is actually needed for the desktop apps in Readest, currently it's only available for the mobile Readest apps.

dastarruer:
And on mobile the popup is also very bare-bones, so it may be perfect for this.

chrox:
> And on mobile the popup is also very bare-bones, so it may be perfect for this.

Yes, that's true.

dastarruer:
[...]

PR #5933, opened 2026-08-29
feat(settings): show which scope the Settings dialog writes

### LINKED ISSUE #5932: FR: Show which scope the Settings dialog is editing

Settings can edit the defaults every book shares, or the values of the open book. A checkmark called Global Settings, in the ⋮ menu, decides which. The dialog does not show which is active.

I did not know the choice existed. I opened a book, opened Settings, and changed the font. I assumed I was changing that book. I changed the defaults for every book, and found out much later.

The two states look the same. A reader can use Readest for a year and never find that the dialog has a scope.

Put the answer in the dialog, where a reader will see it. This needs no new control. The checkmark can stay.

Open the ⋮ menu and read the checkmark before each change. That works only if you know the scope exists. The other workaround is to never use book scope.

### Additional Context

A pull request follows.

Possibly related: #5296, which proposes moving book settings to the reader view rather than naming the scope.

bmfrosty:
Closes #5932.

Readest keeps reader settings in two places: the global defaults, and the
overrides of one book. `viewSettings.isGlobal` selects which of the two the
Settings dialog writes. That flag stays behind a checkmark in the Settings Menu,
and the dialog does not show which scope is active.

The default is global. A font that a reader sets with a book open thus goes to
all books, and not to the book on the screen. A reader who flips the checkmark
gets the opposite result: the change stays with that one book. The dialog
announces neither state, so a reader cannot find out later why one book is
different.

This PR adds a banner that shows the active scope. It changes no storage, adds
no state, and writes nothing new.

## The banner

The banner is in the dialog header, and not in the body. The body scrolls, but
the header stays on the screen.

| Where                               | Reads                                     |
| ----------------------------------- | ----------------------------------------- |
| A book in global scope              | **Global Settings**                       |
| A book in book scope                | **This Book — Overrides Global Settings** |
| A panel with no per-book values     | **Always Global Settings**                |
| The library, where there is no book | **Global Settings**                       |

Every panel shows a banner. Integrations and AI hold no per-book values, so the
banner says **Always** there. A missing banner would be ambiguous, because it
could also mean a render failure. The word **Always** also tells the reader why
the ⋮ toggle does not change that banner.

The banner is read-only. The ⋮ checkmark changes the flag.

### Colour

Blue shows global, and amber shows one book. Global is the usual state, so it
must not look like a warning. The two colours also separate well for deutan and
protan vision. `themeVariables` sets `--color-info` and `--color-warning` from
one `STATE_COLORS` table for each theme, so a palette cannot change them.

Colour is never the only signal, and a browser test measures this.
`scope-banner-eink.browser.test.tsx` imports the class strings of the banner and
renders them against the compiled `globals.css`. On a usual screen the two tints
differ, and the inline-start bar is 4px. On e-ink `eink-bordered` gives both
scopes one colour and a 1px bar. The icon and the words then carry the scope.
CI runs that lane on the first shard.

Neither tint gives 3:1 contrast on the light themes. The border and the icon do
the work, and the tint is decoration.

## One resolver

`isSettingsScopeGlobal` serves the banner, the Settings Menu and the save path,
so the three cannot disagree. It replaces two `?? true` sites and matches them
for every value the type allows, including the null `viewSettings` a view holds
through a reload. It diverges only on a falsy non-nullish flag — `0` or `''` —
which the type forbids, and which now reads global rather than book scope.

`PANEL_SCOPE` is a `Record` keyed by `SettingsPanelType`, not an array, so the
compiler refuses a new panel until this file classifies it.

## Where the banner is not the whole truth

The banner makes one claim for a whole panel. Some rows do not obey that claim.
This PR adds none of them. The banner is only the first thing that states a
scope which those rows can contradict.

- **Some rows are always app-wide.** Behavior has thirteen, Theme has ten, and
  TTS has four. In book scope the banner says **This Book** above them. Theme
  Mode and Theme Color are the first two controls under the banner.
- **Four rows are always per-book.** They pass `skipGlobal`. In global scope the
  banner says **Global Settings** above them. In the library these rows write to
  no store at all.
- **Some rows write to the book and read from the global.** They are UI
  Language, Highlight Opacity, E-Ink Mode and Color E-Ink Mode.
- **A panel can show the other store's values.** Font and Layout fill their rows
  from the open book, in either scope. The banner is still correct about the
  write.
- **"Overrides" can be untrue.** `serializeConfig` keeps only a value that
  differs from the global. A value moved away and then back leaves no override.
- **Five sub-pages replace the panel body.** The header and the banner stay, and
  those pages write app-wide. Two deep links open the dictionary page directly.
- **A reload makes the banner flip and return.** A new view holds null settings
  for the full load, and the resolver reads null as global.

Each panel must fill and guard its rows from the scope in force to correct
these. That is sixteen files, and it is not this PR.

## Accessibility

The banner is a `role='status'` live region. The scope can change from the ⋮
menu, and that menu announces nothing when it closes.

The text also depends on the panel. A move from Font to Integrations thus
announces the banner on top of the tab change. On Integrations and AI the text
does not depend on the flag, so a ⋮ toggle there announces nothing. The sighted
case has the same gap: the checkmark moves, and the banner does not.

The two states differ in three ways: the icon, the words, and the tint. Each one
gives the scope alone.

## Size

43 files: 4 source, 4 test, 34 locale, and `DESIGN.md`. Each locale gets two new
strings.

`DESIGN.md` §6 gets a **Status banner (dialog header)** subsection. §13 asks for
a subsection when the system gets a new archetype. §6 holds the dialog header
archetypes, and it numbers nothing. Move the subsection to §4 or §5 if you
prefer the letter of §13.

## Translations

Each new string is built around that locale's own text for "Global Settings", so
the banner and the ⋮ menu name the same thing. Six locales inflect that label,
because a case ending cannot stand as a separate word.

**I read only English.** Treat all 34 as unverified. Corrections are welcome,
and a wrong string costs one more commit.

On a narrow phone the book state truncates to **This Book**. The icon and the
colour still separate the two states.

## Tests

The unit tests read the banner in each of its three states, and in the library.
They read the words, the icon and the colour separately, because on e-ink the
tint goes and the icon must still carry the scope.

The resolver has its own tests: no settings to read, and a flag that is not a
boolean. A behaviour test also pins the save path to the resolver.

Four tripwires read the source. They guard the lines that no render test
reaches: the dialog rendering the banner as a direct child of the header with
this book's key; the banner, the ⋮ menu and the save path all calling the
resolver; `PANEL_SCOPE` staying an exhaustive `Record`; and the banner building
its class list from the exported strings. These tests use `readFileSync`, so
watch mode does not re-run them. Only a full run finds a broken connection.

I proved each test: I made the change it guards, and confirmed the test fails.

## Verification

Rebased onto `main` at `f6146c217`. Types, lint and format are clean.

This environment fails several hundred tests on `main` alone, so the comparison
is the failing set and not the count. I ran `main` and this branch in one
checkout, one after the other. The two failing sets are identical, file for
file. This branch adds 17 passing tests to that lane, and 3 to the browser lane.

## Summary by CodeRabbit

* **New Features**
  * Added clearer settings scope options: “Apply to All Books” and “Apply to This Book.”
  * The selected scope is now indicated directly in the settings menu.

* **Bug Fixes**
  * Prevented unnecessary saves when selecting the already active settings scope.


bmfrosty:
Thanks — I have applied the translation feedback.

**Applied as suggested:** `de`, `fr`, `sv`, `pl`, `he`, `hi`, `id`, `ms`.

**Applied with one change:** `es`. I took the grammar and sentence-case fix, but kept a verb that does not mean the global values are destroyed:

```
This Book — Overrides Global Settings: Este libro — prevalece sobre la configuración global
Always Global Settings:                Siempre se aplica la configuración global
```

`anula` reads as "annuls" or "voids". The banner means this book's values take precedence while the global ones remain, so `prevalece sobre` is the meaning I want.

**Not applied:** `bo`. The comment asks for confirmation from a Tibetan translator and offers no replacement, so I have left the string and flagged it here.

**Still unfixed, and worth knowing:** the capitalisation fault is wider than the review found. 19 locales embed the ⋮ menu label with its title case mid-sentence — the 10 reviewed plus `el`, `hu`, `it`, `nl`, `pt`, `pt-BR`, `ru`, `sl`, `tr`, `uk`, `uz`, `vi`. I have not touched those, because I read only English and did not want to guess at 9 more languages. Native corrections are welcome, in this PR or a follow-up.

chrox:
Thanks for this, and for the issue that goes with it. The write-up, the tripwire tests, and the honesty about what the banner can't promise are all well above the bar I usually get, so I want to be equally direct about where I've landed.

**Short version: I think the problem is real but smaller than a permanent banner, and I'd rather fix the part of it I can actually evidence.**

### Sizing

#5932 is currently the only report of scope confusion in the tracker, with no comments and no reactions, and #5296 has one upvote. Global-by-default also matches what Kindle, Apple Books and Play Books do with typography, so a reader sitting in the default state holds the mainstream mental model. And the mistake is undone by one gesture. That's not nothing, but it isn't permanent-chrome sized.

The simplicity cost is what I keep coming back to. The banner renders on every panel, in every scope, and in the library. In the default state, where nearly every session lives, it says "Global Settings", which restates the default. An indicator that is always on stops being read; it only carries information when it marks the unusual state. On mobile it takes ~31px permanently from a sheet that `snapHeight` already caps at 70%.

### But something is broken, and it's narrower than the framing

`DialogMenu.tsx:67` sets `buttonClass='lg:tooltip'`, and `MenuItem.tsx:124` passes the text through the native `title` attribute. Native `title` never fires on touch, and `lg:` starts at 1024px. So `Apply to All Books` / `Apply to This Book` — the only copy in the app that states the consequence — is invisible on every phone and tablet. The menu just reads "Global Settings" with a checkmark.

That is your "I did not know the choice existed", and it's a labelling bug rather than a missing surface. What I'd take instead:

```
⋮ →  ✓ Apply to All Books
     ○ Apply to This Book
```

Both strings already exist and are translated in all 34 locales. No new UI, no new i18n keys, no new design-system vocabulary.

If you'd like to rework the PR that way, I'll review it promptly. If you'd rather not, that's completely fair and I'll pick it up from here.

### Two things worth recording either way

**The tint pair doesn't survive measurement.** `bg-info/15` and `bg-warning/10` come out 1.02–1.09:1 against each other, and the `border-s-4` bar misses the 3:1 floor in DESIGN.md §7 on every light theme (1.49 on `default-light`, 1.11 on `gray`). On dark themes the bar is fine but the two tints collapse to 1.02:1. `STATE_COLORS` being theme-invariant is the cause rather than the safeguard — one fixed blue/amber pair can't hold across 11 base surfaces running from `#ffffff` to `#002b36`. Worth knowing before reaching for those tokens again; this PR would be their first use in `src/`.

**Two tripwires in `scopeWiring.test.ts` will fire on innocent edits.** The stripper at :60 doesn't handle escaped quotes, so a single `\'` anywhere in `SettingsDialog.tsx` desyncs the brace count and fails with a message that blames the banner — 21 files in the app already contain one. And :82 pins the local variable name `viewSettings`, so inlining it to the same shape the other two rows use breaks the test.

For what it's worth, the rest holds up: the resolver is behaviour-identical to both `?? true` sites for every value the type allows, `PANEL_SCOPE`'s classification is correct, the `eink-bordered` claim is true, and all 20 new tests pass along with `tsc` and Biome.

One last thought — #5296 proposes moving book-scoped settings into the reader view, which dissolves the flag rather than explaining it. If that's where this ends up, signposting the flag now is work we'd later delete.

chrox:
Heads up — rather than leave you to do the rework, I've pushed it to this branch myself (`302b566f6..6119acbb4`). It's a new commit on top of yours, not a force-push, so your original is intact in the history and you can drop mine with `git reset --hard 302b566f6` if you'd rather take it a different way.

What the branch is now, at +95/−11 across 3 files:

- `DialogMenu.tsx` — the single "Global Settings" checkmark becomes the two exclusive rows, using the `toggled` prop the menu already uses for mutually exclusive choices elsewhere (e.g. library `ViewMenu.tsx:199`). No new strings, no new surface.
- `helpers/settings.ts` — **your `isSettingsScopeGlobal` resolver, kept as-is**, along with its tests. Folding the two `?? true` sites into one resolver is a genuine improvement and stands on its own merits; it's the part of this PR I wanted regardless of what happened to the banner.
- The banner, `PANEL_SCOPE`, the three new test files, the DESIGN.md subsection and the 34 locale entries are reverted.

`pnpm test` is green (10,434 passed), `pnpm lint` and `tsc` clean.

I've credited you as co-author on the commit. If you disagree with any of this, say so and I'll happily revisit — you did the work of finding the problem, and the diagnosis in your issue is what made the actual fix findable. The `lg:tooltip` blind spot is not something I would have gone looking for on my own.

chrox:
Correction to my last comment: I said your `isSettingsScopeGlobal` resolver was kept as-is. It isn't any more — `04e28a53b` removes it and restores the inline `?? true`. You should hear the reasoning rather than just find it gone.

Two things decided it. The resolver's stated justification was that three readers must not disagree — the banner, the ⋮ menu and the save path. Two of those three were the banner, so with the banner gone it's back to the two call sites that existed before this branch, both already spelling the rule identically.

And its distinguishing logic turned out to be unreachable. `ViewSettings.isGlobal` is declared `boolean` at `types/book.ts:471` — not optional, not nullable — so `typeof chosen === 'boolean' ? chosen : true` can never take the false branch for any value the type permits. The corrupt `0` / `''` case it defends is one the type forbids. Without that branch the helper is `?? true` with an extra hop, against 17 files in the app that spell it inline.

Your reasoning for the guard wasn't unreasonable, for the record — `book_configs` does arrive over sync from other devices and from KOSync/koplugin, so malformed values aren't pure fantasy. I just don't think a guard on one field at one read site is where that belongs; it wants validation at the sync boundary, applied to the whole config rather than this one flag.

So the branch is now a single file, +18/−8, all of it in `DialogMenu.tsx`. That does mean none of your code remains, which I'd rather say plainly than let you discover. What remains is your issue and your diagnosis, and those were the hard part — the `lg:tooltip` blind spot is not something I would have gone looking for without #5932 pointing at the scope in the first place. I've kept you as co-author on the fix commit.

Full suite green (10,431 passed), lint and `tsc` clean.

bmfrosty:
Thanks for the response.

I came across the fact that there were two different scopes while I was playing with another feature I was thinking about and was confused by some inconsistencies in my testing. I'm just back home and had been thinking when I was out about another way to display this. In the header bar of the settings, adjacent to ⋮, having a 🌐 or 📖 to indicate the current scope would be way less intrusive size-wise and would only take a little horizontal space. If one wanted, one could make it interactive to change scopes.

On the same lines, the scope change could fade in a short toast using the strings that already exist — Apply to All Books or Apply to This Book — so the change confirms itself in words. Readest already has the toast surface, so that part is a dispatch rather than anything new. The 🌐 / 📖 pair could also sit beside those same strings in the ⋮ menu and in the toast, which would let the header bar icon teach its own meaning: someone who notices it and wonders what it is can find out by using it once, and after that it is a one-glyph answer they can read at a glance.

PR #5964, opened 2026-08-30
feat(reader): add selectable OCR overlays for scanned books

### LINKED ISSUE #173: OCR compatibility
**Does your feature request involve difficulty completing a task? Please describe.**
OCR needs to be implemented, so can be or not hard depending on the library used.

Like I said on Reddit, I thought it could be interesting to have OCR for manga. The idea behind is Bilingual Manga, a website where you can select text and switch to English. Only main idea here is to allow to select, so manga readers could check the words (using Yomichan or any integrated on the app, if any).

**Additional context**
Sorry for tagging here too. I thought it wouldn't be lost if it was here.

WhiteHades:
Closes #173.

Adds OCR for CBZ and scanned PDFs, processed on your device. Models download on demand. Japanese manga uses Paddle recognition with Manga OCR refinement and Tesseract fallback. Heavily inspired by Mokuro's text detection, line cropping and hover overlays.

Hover or select text to use Readest's existing dictionary and translation popups. Includes vertical text, current page priority, a small background buffer and one progress toast. OCR results stay in the session and are never saved or synced.


WhiteHades:
@chrox what say you - should OCR stay session only with on demand models, and is Tesseract fine as the baseline?

chrox:
> @chrox what say you - should OCR stay session only with on demand models, and is Tesseract fine as the baseline?

Yes to all three.

PR #6001, opened 2026-09-01
READ-9: shrink hidden Android navigation buttons to 8px

### LINKED ISSUE #5990: Unexpected page jumps when tapping the bottom corners on Onyx Boox Go 7
### Bug Description

I noticed another issue that does not occur on my Pixel 9 Pro and seems to happen only on my Onyx Boox Go 7.

Even when “Paginate by tapping” is disabled, tapping the bottom-right area of the page causes the book to jump 10–15 pages backwards. Since the book is RTL, I would normally expect the right side to correspond to the previous page, but the jump is much larger than a single page.

I think the app may be navigating to the previous section instead of simply moving to the previous page.

Similarly, tapping the bottom-left area causes the book to jump 10–15 pages forward, seemingly to the next section.

This is particularly noticeable because the bottom part of the screen is easy to touch accidentally while reading, so these unexpected jumps can happen unintentionally.

### Steps to Reproduce

disable "tap to paginate"
Bottom-right tap → jumps approximately 10–15 pages backwards (on a rtl book)
Bottom-left tap → jumps approximately 10–15 pages forwards
The behavior appears to correspond to navigating between sections rather than individual pages (even if the "Tap to paginate" is active)
The issue occurs on the Onyx Boox Go 7
I cannot reproduce it on my Pixel 9 Pro

Onyx Boox Go 7 BW

Android 13
Firmware: 4.2-rel x
Readest: 0.12.6
WebView: 151.0.7922.200

Pixel 9 Pro

Android 17
Readest: 0.12.1
WebView: 151.0.7922.199
Works as expected

Thanks a lot, @chrox!
I always wish you all the best! : )

### Expected Behavior

When “Paginate by tapping” is disabled, tapping the bottom corners should not cause a large jump to another section.

Alternatively, there should be a separate option to disable this bottom-corner navigation behavior entirely, since accidental taps in this area are very easy to make while reading.

[dead souls - persian.zip](https://github.com/user-attachments/files/31644609/dead.souls.-.persian.zip)

### Reproduction File

_No response_

### Operating System

Android

### OS Version

13

### Readest Version

Readest: 0.12.6 (WebView: 151.0.7922.200)

### Custom CSS

_No response_

chrox:
## Summary

- Reduce the invisible Android page and section navigation controls from 16×16 px to 8×8 px, minimizing accidental bottom-corner section jumps.
- Keep visible controls and all non-Android behavior at the existing 80×80 px size.
- Update the existing browser-spec expectation to match the new Android dimensions.

## Verification

- `git diff --check` passes.
- Unit tests were not run, per the issue request.

## Pre-Landing Review

No issues found.

## Design Review

No issues found. The smaller touch area applies only to invisible Android controls and directly addresses accidental activation on e-ink devices.

## Scope

Scope check: clean. No iOS size, layout, or behavior changes.

Fixes #5990
Closes READ-9

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced the hidden Android page-navigation controls’ touch area to minimize interference with text selection.
  * Preserved overflow handling while tightening the navigation hit-area behavior.

### DIFF
--- a/apps/readest-app/src/__tests__/components/page-navigation-buttons.browser.test.tsx
+++ b/apps/readest-app/src/__tests__/components/page-navigation-buttons.browser.test.tsx
@@ -63,7 +63,7 @@ describe('PageNavigationButtons Android hit areas', () => {
     for (const label of navigationLabels) {
       const button = screen.getByRole('button', { name: label });
       const bounds = button.getBoundingClientRect();
-      expect([bounds.width, bounds.height], label).toEqual([16, 16]);
+      expect([bounds.width, bounds.height], label).toEqual([8, 8]);
 
       const hitTarget = document.elementFromPoint(bounds.left + bounds.width / 2, bounds.top - 1);
       expect(button.contains(hitTarget), label).toBe(false);
--- a/apps/readest-app/src/app/reader/components/PageNavigationButtons.tsx
+++ b/apps/readest-app/src/app/reader/components/PageNavigationButtons.tsx
@@ -39,7 +39,7 @@ const PageNavigationButtons: React.FC<PageNavigationButtonsProps> = ({
     (hoveredBookKey === bookKey || isDropdownOpen) && viewSettings?.showPaginationButtons;
   const navigationButtonSize =
     !isPageNavigationButtonsVisible && appService?.isAndroidApp
-      ? 'h-4 w-4 overflow-hidden'
+      ? 'h-2 w-2 overflow-hidden'
       : 'h-20 w-20';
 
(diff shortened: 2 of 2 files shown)


shahram7:
Hi Hung,
Thanks a lot.
just out of curiosity, what is the purpose of these hidden navigation buttons? Are they necessary? 
BR, Shahram

chrox:
It’s intended for screen-reader accessibility. On Android, the buttons need to receive pointer events, which may indicate an issue specific to Android WebView.

PR #6027, opened 2026-09-02
feat(settings): add a Gamepad Support toggle

### LINKED ISSUE #5979: FR: Gamepad Toggle / Remap

Hello! I love Readest so far, but after installing the flatpak on my Steam Deck, the built-in gamepad support conflicts with Steam Input remapping - causing double inputs, or simply undesired control.

Amusingly, the useGamepad.ts method of remapping via `dispatchKey` also causes the SteamOS overlay to receive double inputs while the app is open, but that's probably on their end. :P

I think it'd be great to have user-facing options to:
* Remap gamepad actions (via existing Keyboard Shortcuts menu?)
* Enable/disable Gamepad support entirely

I imagine for most use cases, the toggle option alone would be fine!

I searched for any kind of environment variable to control/disable the native browser Gamepad API, but to no avail.

### Additional Context

Thank you!

chrox:
Closes #5979.

## The problem

Readest replays a connected controller's buttons and sticks as key events in the reader, and there was no way to turn that off. On a handheld whose own remapper already binds the same buttons to keys, every press lands twice: the remapper's key plus Readest's own.

The reporter hit this with the Flatpak on a Steam Deck, where Steam Input is already mapping the D-pad and sticks, so one button press turns two pages. They also noticed the SteamOS overlay picking up double inputs while Readest is open.

## The change

Settings > Behavior > Device > **Gamepad Support**, on by default so nothing changes for anyone who is not affected.

| | |
|---|---|
| `types/settings.ts` | new `gamepadEnabled: boolean` on `SystemSettings` |
| `services/constants.ts` | `gamepadEnabled: true` in `DEFAULT_SYSTEM_SETTINGS` |
| `settings/ControlPanel.tsx` | the switch row, last item of the Device list |
| `services/commandRegistry.ts` | registry entry so settings search finds it |
| `reader/components/ReaderContent.tsx` | gates `useGamepad` and `useAndroidGamepadConnection` |

No migration is needed: `loadSettings` spreads `DEFAULT_SYSTEM_SETTINGS` under the saved settings, so existing profiles read `true`.

Turning it off takes effect immediately. `saveSysSettings` writes through the store, the primitive selector in `ReaderContent` re-evaluates, and `useGamepad`'s cleanup cancels the animation frame loop and removes the `gamepadconnected` / `gamepaddisconnected` listeners. Nothing has to be reloaded.

## Scope

Only the toggle, which is what the issue said would cover most cases. Remapping is left out: once Readest stops competing for those buttons, Steam Input and the desktop remappers already own that job.

## Verification

- New unit test `ControlPanelGamepad.test.tsx` (written first, confirmed failing): the row reflects the saved preference, and toggling it persists `gamepadEnabled`. `useGamepad`'s existing test already covers listeners and polling being torn down when `enabled` flips false.
- `pnpm test` 888 files / 10679 tests passed, 16 skipped.
- `pnpm lint` (tsc + biome) and `pnpm format:check` clean.
- In the browser, with a synthetic `gamepadconnected` event against a spy: setting off, no handler runs; toggled on from the reader's own settings dialog with no reload, the handler runs; toggled off again, silent. The off state survives a page reload, and searching settings for "gamepad" returns the row.
- All 34 locales translated.

## Summary by CodeRabbit

* **New Features**
  * Added a Gamepad Support setting under Settings > Behavior > Device.
  * Gamepad support is enabled by default and can be turned off to prevent duplicate inputs.
  * Added search support for the setting using gamepad, controller, joystick, and related terms.
  * Added localized labels and guidance across supported languages.

* **Tests**
  * Added coverage verifying the setting reflects saved preferences and persists changes.

### DIFF
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -2538,5 +2538,7 @@
   "As {{alias}}": "باسم {{alias}}",
   "Keep this screen on so nearby devices can find you.": "أبقِ هذه الشاشة قيد التشغيل حتى تتمكّن الأجهزة القريبة من العثور عليك.",
   "Devices disappear from the list when their screen turns off; keep the screen on to stay visible.": "تختفي الأجهزة من القائمة عند إيقاف تشغيل شاشتها؛ أبقِ الشاشة قيد التشغيل لتبقى مرئيًا.",
-  "{{name}}'s Readest": "Readest الخاص بـ {{name}}"
+  "{{name}}'s Readest": "Readest الخاص بـ {{name}}",
+  "Gamepad Support": "دعم ذراع الألعاب",
+  "Navigate with a connected controller": "التنقل باستخدام ذراع ألعاب متصل"
 }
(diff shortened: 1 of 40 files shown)


alexstrout:
Just wanted to leave a quick note that this works perfectly - thank you! :heart:

PR #6057, opened 2026-09-04
feat(tts): skip inline parenthetical readings

### LINKED ISSUE #5954: FR: Option to skip inline parenthetical readings/glosses in TTS (plain-text ruby not covered by #1608)
### Does your feature request involve difficulty for you to complete a task? Please describe.

#1334 / #1608 fixed double reading for ruby annotations marked up with proper `<ruby>`/`<rt>` tags. However, many ebooks — especially older digitizations and self-converted texts — do not use ruby markup at all. Instead, the reading is written inline in parentheses right after the base word, as plain text:

> 彼は憂鬱（ゆううつ）な気分だった。

Here 憂鬱 and （ゆううつ） are the same word: the parenthetical kana is purely a visual reading aid. Since it is plain text rather than an `<rt>` element, the ruby fix does not apply, and TTS reads the word twice in a row ("yūutsu yūutsu"). To an English speaker it would sound like the text said "NASA (NASA)" — the identical word spoken twice back to back.

Books converted from Aozora Bunko sources also commonly use 《…》 for the same purpose (e.g. 憂鬱《ゆううつ》), with the same result.

This convention appears very frequently — in some books on nearly every page — so the double reading significantly degrades the TTS listening experience. Similar inline gloss conventions exist in other CJK-language ebooks as well. The underlying "double reading" problem is documented by W3C as a known TTS accessibility concern: https://w3c.github.io/ruby-t2s-req/

### Describe the solution you'd like

An option in TTS settings to skip inline parenthetical readings, for example:

- A toggle like "Skip parenthetical annotations in TTS", or
- A smarter heuristic: when a segment enclosed in （）/()/《》 immediately follows a CJK word and consists only of kana or Han characters, skip it during TTS playback (similar in spirit to how `<rt>` is skipped since #1608).

The visual text should remain unchanged — this should only affect what is sent to the TTS engine.

### Describe alternatives you've considered

- Editing the epub to strip the parenthetical readings, but this modifies the book and has to be repeated for every book.
- Living with the double reading, but it breaks immersion in books that use this convention heavily.

chrox:
## Summary

- Show a Japanese-UI-only opt-in **Skip Parenthetical Readings** TTS setting for plain-text ruby written as `憂鬱（ゆううつ）`, `憂鬱(ゆううつ)`, or `憂鬱《ゆううつ》`.
- Remove only conservative Han-plus-kana/Han reading annotations at the final synthesized TTS timeline and SSML boundary, preserving the visible book text and SSML marks.
- Apply the setting to new, active, and reattached background TTS sessions while leaving recorded narration unchanged.
- Translate the setting copy across all supported locales.

Closes #5954

## Test Coverage

```text
CODE PATHS                                      USER FLOWS
[+] Default/config                              [+] Settings UI
  └── [★★★ TESTED] opt-in defaults false          ├── [★★ TESTED] renders unchecked
                                                   ├── [★★ TESTED] persists false → true
                                                   └── [GAP] true → false/reset persistence
[+] Inline annotation filter                   [+] Start/reattach TTS
  ├── [★★★ TESTED] （kana）, (kana), 《kana》       ├── [★★ TESTED] new controller receives false
  ├── [★★★ TESTED] Han gloss                      ├── [★★ TESTED] adopted controller receives false
  ├── [★★★ TESTED] conservative exclusions        ├── [GAP] enabled value on new/adopted sessions
  ├── [★★★ TESTED] cross-element SSML              └── [GAP] toggle updates active session
  ├── [★★★ TESTED] malformed XML fallback
  └── [GAP] multiple annotations plus exact
             valid-SSML/mark preservation
[+] TTSController                              [+] Synthesized playback
  ├── [★★ TESTED] final SSML boundary disabled     ├── [★★ TESTED] option off speaks annotation
  ├── [★★ TESTED] final SSML boundary enabled      └── [★★ TESTED] option on skips annotation
  ├── [GAP] enabled timeline text filtering
  └── [GAP] setting change invalidates/rebuilds timeline

COVERAGE: 23/28 paths tested (82%)
QUALITY: ★★★ edge/error coverage for parsing; ★★ behavior coverage for UI/controller wiring
```

Tests: 999 → 1001 (+2 files). Coverage gate: PASS (82%, target 80%).

## Pre-Landing Review

No issues found.

## Design Review

Design Review (lite): no issues found. The setting uses the existing boxed-list and switch primitives, including the established e-ink treatment.

## Eval Results

No prompt-related files changed; evals skipped.

## Scope Drift

Scope Check: CLEAN. The branch contains only the requested TTS behavior, setting, tests, and translations.

## Plan Completion

No relevant plan file detected; completion audit skipped.

## Verification Results

No plan verification section was available; verification used the repository test, lint, translation, and build commands below.

## Documentation

Documentation is current. Existing documentation does not enumerate individual TTS settings and remains accurate.

## Test plan

- [x] Full Vitest suite: 901 files passed, 10,857 tests passed; 4 files/16 tests skipped
- [x] Settings visibility: shown for Japanese UI locales and hidden otherwise
- [x] TypeScript and Biome lint: 2,338 files checked
- [x] Translation validation: no untranslated placeholders
- [x] Production Next/Tauri-target build and static export completed
- [x] Repository pre-push hook passed format, lint, and the full test suite

Generated with OpenAI Codex.

## Summary by CodeRabbit

- **New Features**
  - Added a Japanese-language text-to-speech option to skip parenthetical kana or Han readings following Han text.
  - The preference is saved with reader settings and applies when starting, resuming, or updating speech playback.

- **Localization**
  - Added translations for the new speech controls across supported languages.

- **Tests**
  - Added coverage for the setting, playback behavior, annotation handling, and default configuration.

### DIFF
--- a/apps/readest-app/public/locales/ar/translation.json
+++ b/apps/readest-app/public/locales/ar/translation.json
@@ -2547,5 +2547,8 @@
   "{{alias}} wants to send you {{count}} book(s)_two": "يريد {{alias}} إرسال كتابين إليك",
   "{{alias}} wants to send you {{count}} book(s)_few": "يريد {{alias}} إرسال {{count}} كتب إليك",
   "{{alias}} wants to send you {{count}} book(s)_many": "يريد {{alias}} إرسال {{count}} كتابًا إليك",
-  "{{alias}} wants to send you {{count}} book(s)_other": "يريد {{alias}} إرسال {{count}} كتاب إليك"
+  "{{alias}} wants to send you {{count}} book(s)_other": "يريد {{alias}} إرسال {{count}} كتاب إليك",
+  "Speech": "النطق",
+  "Skip Parenthetical Readings": "تخطي القراءات بين قوسين",
+  "Do not speak kana or Han readings shown after Han text": "عدم نطق قراءات الكانا أو الهان الظاهرة بعد نص الهان"
 }
--- a/apps/readest-app/public/locales/bn/translation.json
+++ b/apps/readest-app/public/locales/bn/translation.json
@@ -2307,5 +2307,8 @@
   "BookOrbit Sync": "BookOrbit সিঙ্ক",
   "Store your library, reading progress, and highlights in your Readest account. You can change this any time in Settings.": "আপনার লাইব্রেরি, পড়ার অগ্রগতি এবং হাইলাইটগুলি আপনার Readest অ্যাকাউন্টে সংরক্ষণ করুন। আপনি যেকোনো সময় সেটিংসে এটি পরিবর্তন করতে পারেন।",
   "{{alias}} wants to send you {{count}} book(s)_one": "{{alias}} আপনাকে {{count}}টি বই পাঠাতে চায়",
-  "{{alias}} wants to send you {{count}} book(s)_other": "{{alias}} আপনাকে {{count}}টি বই পাঠাতে চায়"
+  "{{alias}} wants to send you {{count}} book(s)_other": "{{alias}} আপনাকে {{count}}টি বই পাঠাতে চায়",
+  "Speech": "কথন",
+  "Skip Parenthetical Readings": "বন্ধনীর মধ্যের পাঠ এড়িয়ে যান",
+  "Do not speak kana or Han readings shown after Han text": "হান পাঠ্যের পরে দেখানো কানা বা হান পাঠ উচ্চারণ করবেন না"
 }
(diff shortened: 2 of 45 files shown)


54wedge:
I remember this PR https://github.com/readest/readest/pull/2952. It could be useful for this case. I can’t find this feature anywhere. Hmmm

chrox:
Writing a correct and reliable TTS proofreading rule for this would be cumbersome, so we implement it this way instead.

PR #6075, opened 2026-09-05
fix(tts): stop a redundant PDF render detaching the ranges TTS reads

### LINKED ISSUE #6071: TTS on PDF: paragraphs skipped, no highlight, pages loop — a second un-awaited render detaches the live TTS ranges
### Bug Description

Read Aloud on a PDF is broken from the first page turn onward. Three symptoms, all from one cause:

- **Paragraphs are skipped** — most of each page is silently passed over.
- **No word/sentence highlight** while speaking.
- **Pages loop**: when a page goes fully dry, TTS reads one paragraph, turns the page, reads one paragraph, and keeps going that way to the end of the book.

Reported on iOS; **reproduced on Android** (Xiaomi 2211133C, Android 16, WebView 153, Readest 0.12.6), so this is shared web code and affects every platform. EPUBs are unaffected. The reporter tried several PDFs and all behaved the same.

### Steps to Reproduce

1. Import any text PDF (`APerfectSpy-JohnleCarre.pdf`, 315 pages, attached below).
2. Open it and start Read Aloud.
3. Let it read past the end of the first page.

### Expected Behavior

TTS reads every paragraph of each page with the spoken sentence highlighted, then advances — as it does in an EPUB.

### Actual Behavior

After the first page turn, most blocks come back with empty text. The controller skips them, nothing highlights, and a page whose blocks are all empty causes an immediate advance — producing the one-paragraph-per-page loop.

### Reproduction File

`APerfectSpy-JohnleCarre.pdf` (1,862,911 bytes, md5 `ada274e2e2cbb621f2497833fe6c3937`). Any text PDF reproduces it.

### Operating System

Reported on iOS. Verified on Android (Xiaomi 2211133C / Android 16 / Readest 0.12.6).

### Root cause

**Each PDF page is rendered twice per page turn, and the second render destroys the DOM that the freshly-built `TTS` already holds `Range`s into.**

`packages/foliate-js/pdf.js:462` — every PDF render wipes the text layer:

```js
const container = doc.querySelector('.textLayer')
container.replaceChildren()
```

`packages/foliate-js/fixed-layout.js:217` — the renderer re-renders on any resize, **discarding** the render promises:

```js
#observer = new ResizeObserver(() => this.#render())
```

`#render()` returns a `renderPromises` array precisely so callers can wait for the async text-layer rebuild — `fixed-layout.js:808-809` does it correctly (`await Promise.all(renderPromises)`). The `ResizeObserver` does not: it is fire-and-forget. (`fixed-layout.js:331` already notes that this observer's callback "can land in the same tick as a page turn".)

So a page turn during TTS goes:

1. `TTSController.forward()` → `#initTTSForNextSection` → `#initTTSForSection` → `onSectionChange` → `renderer.goTo` → `goToSpread` → `#render` → pdf `render()` → `.textLayer.replaceChildren()`. This one **is** awaited.
2. `#initTTSForSection` (`TTSController.ts:709`, `:754`) then builds `new TTS(doc, …)` over that freshly rendered `.textLayer` and materialises its sentence `Range`s.
3. **~150 ms later the `ResizeObserver` fires** (the page turn changed layout) and re-renders the same page. `replaceChildren()` runs again and detaches every node the live TTS ranges point at.
4. `getPDFSentenceBlocks` (`packages/foliate-js/tts.js:156-213`) captured those text nodes up front, so from here on the blocks yield empty text → empty SSML.
5. Empty SSML makes `TTSController.ts:1498-1520` skip the block; when a whole page comes back empty it advances to the next section. `#nossmlCnt` resets to 0 on any non-empty block (line 1522), so the 10-strike stop at line 1500 never trips and it runs to the end of the book.

Dead ranges also explain the missing highlight — there is nothing left in the document to anchor to.

#### Device trace

Instrumented over CDP on the Xiaomi, patching `Element.prototype.replaceChildren` inside the PDF iframe and wrapping `view.tts.next()`. One page turn (page 15), times in ms:

```
 998  WIPE     page 15 textLayer wiped (104 children)   <- awaited render, from renderer.goTo
1019  NEW TTS  built over page 15 (104 children)
1021  next ->  "16Pym was a handsome man, boyi…"        <- blocks materialise fine
1021  next ->  "In his early fifties he was in…"
1021  next ->  "But the best thing about him i…"
1147  WIPE     page 15 textLayer wiped AGAIN            <- ResizeObserver, un-awaited
6711  next ->  ""    (empty from here on)
6712  next ->  ""
6713  next ->  ""
```

Stack of the first wipe:

```
pdf render() >> Promise.all >> #render >> goToSpread >> renderer.goTo
  >> onSectionChange >> #initTTSForSection >> #handleNavigationWithoutSSML >> forward
```

Stack of the second wipe — no synchronous caller, i.e. a `ResizeObserver` callback:

```
pdf render()
```

Same pattern captured independently on pages 6, 14 and 17.

### Notes

- **Separate latent bug found while investigating.** `packages/foliate-js/fixed-layout.js` implements no `primaryIndex` at all (`grep -c` → 0); `paginator.js:1689` does, and `apps/readest-app/src/types/view.ts:49` declares it non-optional (`primaryIndex: number`), so TypeScript never flags the `undefined`. Both TTS consumers silently fall back to `contents[0]` (`TTSController.ts:536-537`, `useTTSControl.ts:484`), and `getContents()` (`fixed-layout.js:1723`) returns every iframe whose parent is not `visibility:hidden` — i.e. both pages of a spread. With `spread: none` only one content is live, so this does not fire and it is **not** the cause of this bug; with a two-page spread `contents[0]` would be the wrong page, and `useTTSControl.ts:492`/`:508` would then fire `view.goTo(cfi)` on every sentence mark. Worth fixing separately.
- CBZ never gets TTS at all: `packages/foliate-js/comic-book.js:138` builds sections with no `createDocument`, and `TTSController.ts:688` bails when that is missing.

chrox:
Closes #6071. Depends on readest/foliate-js#91 (pinned here at `8481b9a`; needs a re-pin once that merges).

## The bug

Read Aloud on a PDF skipped most paragraphs, highlighted nothing, and once a page came back fully empty it read one paragraph, turned the page, and looped that way to the end of the book. Reported on iOS; it is shared web code and reproduces everywhere.

## Root cause

Each PDF page is rendered **twice** per page turn:

1. The awaited render, from `forward()` → `#initTTSForSection` → `onSectionChange` → `renderer.goTo` → `goToSpread` → `#render` → pdf `render()`.
2. `#initTTSForSection` then builds `new TTS(doc, …)` over that text layer and materialises its sentence `Range`s.
3. ~130 ms later `fixed-layout.js`'s `#observer = new ResizeObserver(() => this.#render())` fires — the page turn changed layout — and re-renders the same page. It discards the `renderPromises` array that `#render()` returns for exactly this purpose (`fixed-layout.js:808-809` awaits it properly).
4. `pdf.js`'s `container.replaceChildren()` runs again and detaches every node those ranges point at.

The blocks then yield empty text, which `TTSController.ts:1498-1520` skips; `#nossmlCnt` resets on every non-empty block (line 1522), so the 10-strike stop at line 1500 never trips and it runs to the end of the book. Dead ranges also explain the missing highlight — there is nothing left to anchor to.

## The fix

foliate-js skips a render when the page, zoom, page colours and OS font scale are all unchanged (readest/foliate-js#91). Fixing it there rather than in the `ResizeObserver` protects the text layer from *every* redundant render trigger, and skips the rasterisation too.

## Verification

- **New unit test** — a node captured from the first render must still be `isConnected` after an identical second `onZoom`. Red before the fix, green after, plus guard-rails that a zoom change and a `pageColors` change still rebuild. Needs a real iframe document: `render` bails on `document.implementation.createHTMLDocument` because `defaultView` is null.
- **Browser A/B on the real bundle** — unfixed, every page wiped twice with identical content (`had=2368` → `had=2368`, ~150 ms apart); fixed, once per page, and a node captured at render time survived 4 consecutive page turns.
- **Xiaomi 2211133C (Android 16)**, rebuilt and installed, replaying the reporter's own PDF — 310 non-empty blocks, **zero** empty, a clean page turn, word highlight tracking down the page. The old build gave 4 blocks then empties.
- `pnpm test` (10,889 pass) and `pnpm lint` clean.

## Found but not fixed here

`fixed-layout.js` implements no `primaryIndex` (`grep -c` → 0) while `paginator.js:1689` does, and `apps/readest-app/src/types/view.ts:49` declares it non-optional, so TypeScript never flags the `undefined`. Both TTS consumers silently fall back to `contents[0]` (`TTSController.ts:536-537`, `useTTSControl.ts:484`), and `getContents()` returns both pages of a spread. Harmless at `spread: none`; on a two-page spread `contents[0]` is the wrong page and `useTTSControl.ts:492`/`:508` would fire `view.goTo(cfi)` on every sentence mark. Worth a separate fix.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for PDF rendering to verify that unchanged content preserves its existing text layer.
  * Added coverage confirming that text layers are rebuilt when zoom or page-color settings change.

### DIFF
--- a/apps/readest-app/src/__tests__/foliate-pdf-redundant-render.test.ts
+++ b/apps/readest-app/src/__tests__/foliate-pdf-redundant-render.test.ts
@@ -1,0 +1,193 @@
+// readest/readest issue #6071: a redundant PDF re-render detaches the live TTS ranges.
+//
+// Every PDF render tears the text layer down (`container.replaceChildren()`) and
+// builds it again. fixed-layout's `#observer = new ResizeObserver(() => this.#render())`
+// re-renders on any layout change -- including the one a page turn itself causes --
+// and discards the render promises, so a second render lands ~130ms after the first
+// with nothing about the output changed.
+//
+// By then TTSController has already built its `TTS` over the first render's text
+// layer and materialised the sentence `Range`s. The second `replaceChildren()`
+// detaches every node those ranges point at, so the blocks yield empty text: TTS
+// skips paragraphs, nothing highlights, and a page whose blocks are all empty makes
+// the controller advance -- the reported "one paragraph per page" loop.
+//
+// Traced on a Xiaomi 2211133C (Android 16, Readest 0.12.6): every page was wiped
+// exactly twice, ~130ms apart, at an identical scale and host box.
+//
+// `render` must therefore skip the rebuild when nothing that affects its output
+// (the page, the zoom, the page colours, the OS font scale) has changed.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Text of the page the fake pdf.js text layer builds, one span per entry.
+const PAGE_TEXT = ['Magnus Pym got out of his elderly country taxicab. ', 'No one to flap about.'];
+
+// Counts how many times a text layer was actually built.
+let textLayerBuilds = 0;
+
+vi.mock('@pdfjs/pdf.min.mjs', () => {
+  class PDFDataRangeTransport {
+    requestDataRange!: (begin: number, end: number) => void;
+    onDataRange = vi.fn();
+    constructor(
+      public length: number,
+      public initialData: unknown,
+    ) {}
+  }
+
+  // Stands in for pdfjsLib.TextLayer: fills the container the way the real one
+  // does, so the test can hold a node from it and check it stays connected.
+  class TextLayer {
+    #container: HTMLElement;
+    constructor({ container }: { container: HTMLElement }) {
+      this.#container = container;
+    }
+    async render() {
+      textLayerBuilds++;
+      for (const str of PAGE_TEXT) {
+        const span = this.#container.ownerDocument.createElement('span');
+        span.textContent = str;
+        this.#container.append(span);
+      }
+    }
+  }
+
+  class AnnotationLayer {
+    async render() {}
+  }
+
+  const makePage = () => ({
+    getViewport: ({ scale }: { scale: number }) => ({ width: 600 * scale, height: 800 * scale }),
+    render: () => ({ promise: Promise.resolve(), cancel: vi.fn() }),
+    streamTextContent: vi.fn(async () => ({})),
+    getTextContent: vi.fn(async () => ({ items: [] })),
+    getAnnotations: vi.fn(async () => []),
+    cleanup: vi.fn(),
+  });
+
+  const getDocument = vi.fn(() => ({
+    promise: Promise.resolve({
+      numPages: 3,
+      getPage: vi.fn(async () => makePage()),
+      getMetadata: vi.fn(async () => ({ metadata: undefined, info: {} })),
+      getViewerPreferences: vi.fn(async () => null),
+      getOutline: vi.fn(async () => null),
+      getPageLabels: vi.fn(async () => null),
+      getDestination: vi.fn(),
+      getPageIndex: vi.fn(),
+      destroy: vi.fn(),
+    }),
+    destroy: vi.fn(),
+  }));
+
+  (globalThis as unknown as { pdfjsLib: unknown }).pdfjsLib = {
+    GlobalWorkerOptions: {},
+    PDFDataRangeTransport,
+    getDocument,
+    TextLayer,
+    AnnotationLayer,
+  };
+  return {};
+});
+
+const fakeFile = {
+  size: 1,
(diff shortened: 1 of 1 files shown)


chrox:
Re-pinned to `8ffd2cf` now that readest/foliate-js#91 is merged into `main`. Verified the merged file is byte-identical to the build that was device-tested on the Xiaomi.

Issue #4906, opened 2026-07-03
Currently unable to properly detect daily updates


ybbluesky:
when the ci(release): attest release and nightly build artifacts (https://github.com/readest/readest/pull/4851) up,
Currently unable to properly detect nightly build updates

PR #6108, opened 2026-09-07
fix(ci): bundle the audio libraries the CEF AppImage dlopens

chrox:
TTS played through with working highlighting but nothing was audible; every new AppImage also asked for a root password on first launch. This branch fixes both by doing what @Samueru-sama and @FabianLars pointed at in the thread: let quick-sharun deploy Chromium's dependencies itself, and take the pinned-pkgforge route of tauri-apps/tauri#12491 instead of the fork.

## Root cause

Chromium dlopens `libpulse.so.0`, `libspeechd.so.2` and the NSS modules by soname; only libraries CEF *links* against were in the bundle. Whatever is missing is resolved from the host at run time and fails against the bundled glibc 2.35 on any newer distro — no audio path at all (the ALSA fallback dies through the same host plugin), and NSS aborts the app on first use.

quick-sharun's `DEPLOY_CHROMIUM` rule already covers NSS and, via `DEPLOY_PIPEWIRE` → `DEPLOY_PULSE`, pulse. None of it reached the bundle because `@tauri-apps/cli-cef@3.0.0-alpha.26` fetches [FabianLars' fork](https://github.com/tauri-apps/tauri/blob/feat/cef/crates/tauri-bundler/src/bundle/linux/appimage/sharun_cef.rs#L68) of the script, whose HEAD [3e280d1b](https://github.com/FabianLars/Anylinux-AppImages/commit/3e280d1b2270fecfcb2c2c823b490c782ecf1277) drops the `eval` that rebuilds the deployment array: every entry keeps the quotes `_save_array` puts on it, fails `[ -f ]`, and every `DEPLOY_*` extra is silently skipped. The fix that keeps both halves is `eval set -- "$TO_DEPLOY_ARRAY" "$(_save_array "$@")"`. The CLI re-downloads the fork on every build with no way to point it elsewhere.

## Changes

- **Build the CEF tauri CLI from readest/tauri on the Linux legs** (readest/tauri@c62e9f90, the model the workflow used until #6074): `sharun_cef.rs` fetches pkgforge at `facb95e8` verified against #12491's SHA-256, runs the tooling with bash and streamed output (the two halves of the #4906 hang), passes the deploy list as arguments, keeps whitespace out of the AppDir name, and adds no `fix-namespaces` hook. Rev-pinned; a marker skips the rebuild while rust-cache keeps `~/.cargo/bin`. `scripts/tauri.mjs` runs `cargo tauri` when `TAURI_CEF_CARGO=1`.
- **Staging reduced to the two gaps no rule covers**, passed to quick-sharun as arguments via `appimage.files`: Debian keeps the NSS modules in `<triple>/nss/` where the Arch-layout globs miss them, and nothing deploys `libspeechd`.
- **No root-password prompt** (#6110, merged here): a `00-` AppRun hook stubs the two helpers the bundler's hook prompts through; CEF runs without the sandbox feature, so the namespaces it asked for were never used.
- Post-build guard checks the full set lands, including `lib.path` covering `lib/pulseaudio`.

## Verification

Built locally with the patched CLI. quick-sharun deployed, with no staging involved: `libpulse`, `libpipewire`, `alsa-lib/libasound_module_*_pulse.so`, the NSS modules, mesa/dri, gconv, gio and GTK IM modules — 397 libraries, hooks `00-no-userns-prompt`, `01-path-mapping-hardcoded`, `05-vulkan-check`. The AppImage ran headless for the full window with no `FATAL`/`GLIBC_`/`nss_error`/`zenity`/`pkexec` output and the GPU process stayed up; under its own loader and `lib.path`, `libpulse`, `libpulsecommon`, `libsoftokn3` and `libGL` resolve from `<bundle>/lib`. Before the change, the same probes on the shipped `2026090706` bundle failed with `GLIBC_2.38 not found` and a 440 Hz tone through a staged libpulse was the only way to reach PipeWire.

PR CI cannot exercise this (it needs a full CEF build); the first nightly after merge is the CI validation. `STRACE_MODE` stays off — re-enabling dlopen discovery is its own change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux AppImage audio support by bundling required PulseAudio libraries.
  * Improved Web Speech text-to-speech support by bundling the required Speech Dispatcher library.
  * Prevented unnecessary user-namespace setup prompts when launching the Linux AppImage.
  * Expanded build verification to confirm required audio libraries, runtime components, and startup files are packaged correctly in nightly and release builds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux AppImage audio support by ensuring required PulseAudio libraries are packaged.
  * Improved Web Speech text-to-speech support by including the required Speech Dispatcher library.
  * Prevented unnecessary user-namespace setup prompts when launching the Linux AppImage.
  * Improved reliability of Linux AppImage builds and verification for required runtime libraries and startup components.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -132,18 +132,13 @@ jobs:
         with:
           key: nightly-${{ matrix.config.os }}-${{ matrix.config.release }}-${{ matrix.config.arch }}-cargo
 
-      # The published @tauri-apps/cli-cef (3.0.0-alpha.26, pinned in
-      # apps/readest-app/scripts/tauri.mjs) bundles the AppImage with a fork of
-      # quick-sharun whose deployment array is broken — see the staging step
-      # below — and re-downloads that fork on every build with no way to point
-      # it elsewhere. The Linux legs build the CLI from readest/tauri instead,
-      # where sharun_cef.rs fetches a pinned, checksummed pkgforge revision and
-      # runs the tooling with bash and streamed output (the fix behind the
-      # STRACE_MODE hang, tauri-apps/tauri#12491). Pinned to a rev, not the
-      # branch tip, so a release cannot change behaviour under us; rust-cache
-      # keeps ~/.cargo/bin, and the marker file skips the ~6 minute rebuild
-      # while the pin is unchanged. TAURI_CEF_CARGO=1 in the build step makes
-      # scripts/tauri.mjs run `cargo tauri` with the same command line.
+      # The npm cli-cef bundles the AppImage with a broken quick-sharun fork
+      # and cannot be pointed elsewhere. Build the CLI from readest/tauri
+      # instead: its sharun_cef.rs pins a pkgforge revision (the one
+      # tauri-apps/tauri#12491 uses) and runs the tooling with bash and
+      # streamed output. Rev-pinned; the marker skips the rebuild while
+      # rust-cache keeps ~/.cargo/bin. scripts/tauri.mjs runs `cargo tauri`
+      # when TAURI_CEF_CARGO=1.
       - name: install the CEF tauri CLI from readest/tauri (linux only)
         if: matrix.config.release == 'linux'
         run: |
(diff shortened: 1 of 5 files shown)


Samueru-sama:
`quick-sharun` already has a `DEPLOY_ELECTRON` mode that implies `DEPLOY_PIPEWIRE`

https://github.com/pkgforge-dev/Anylinux-AppImages/blob/65f3de5a190234b201686d0083f5c77ed481993f/useful-tools/quick-sharun.sh#L1098

(and `DEPLOY_PIPEWIRE` also triggers `DEPLOY_PULSEAUDIO`).

[Both also imply a minimal set of alsa plugins get added](https://github.com/pkgforge-dev/Anylinux-AppImages/blob/65f3de5a190234b201686d0083f5c77ed481993f/useful-tools/quick-sharun.sh#L1175-L1189)

Pipewire really isn't needed for audio tbh, only needed for the screenshare portal to work. https://github.com/pkgforge-dev/Stoat-AppImage/issues/5

chrox:
Thanks — and agreed, that is exactly where this should end up: the CEF bundler already sets `DEPLOY_CHROMIUM=1`, so `DEPLOY_PIPEWIRE` → `DEPLOY_PULSE` and the minimal alsa plugin set come with it for free, and hand-staging libpulse should not be necessary at all.

The catch is the same one as on 9266f9c4: `@tauri-apps/cli-cef@3.0.0-alpha.26` does not fetch pkgforge's script, it fetches [FabianLars' fork](https://github.com/tauri-apps/tauri/blob/feat/cef/crates/tauri-bundler/src/bundle/linux/appimage/sharun_cef.rs#L68), which is 3851 lines against your 4970 and still expands the deployment array [without the `eval`](https://github.com/FabianLars/Anylinux-AppImages/blob/3e280d1b2270fecfcb2c2c823b490c782ecf1277/useful-tools/quick-sharun.sh#L1345-L1349) — so `* Deploying pipewire` prints and nothing is deployed. What that looks like in the shipped AppImage: `libasound.so.2` is present because it is a `NEEDED` of libcef.so, `libpulse.so.0` is absent, and inside the bundle's own runtime

```
libpulse.so.0   <bundle>/lib/libc.so.6: version `GLIBC_2.38' not found
                (required by /usr/lib/x86_64-linux-gnu/pulseaudio/libpulsecommon-17.0.so)
snd_pcm_open("default") -> No such file or directory   # conf hook -> libasound_module_conf_pulse.so, same failure
```

so there is no audio path at all. Hence the stopgap; the PR description says to drop it once tauri-apps/tauri#12491 lands and the tooling deploys these itself.

Also noted on pipewire — I am not bundling it. Only `libpulse.so.0` for output, and `libspeechd.so.2` for Chromium's Web Speech voices, which as far as I can see no rule in quick-sharun deploys, so that one probably has to stay even after the switch.

Samueru-sama:
> The catch is the same one as on [9266f9c](https://github.com/readest/readest/commit/9266f9c46e1f6eca14205c638ae74d0a0222d25d): `@tauri-apps/cli-cef@3.0.0-alpha.26` does not fetch pkgforge's script, it fetches [FabianLars' fork](https://github.com/tauri-apps/tauri/blob/feat/cef/crates/tauri-bundler/src/bundle/linux/appimage/sharun_cef.rs#L68), which is 3851 lines against your 4970 and still expands the deployment array [without the `eval`](https://github.com/FabianLars/Anylinux-AppImages/blob/3e280d1b2270fecfcb2c2c823b490c782ecf1277/useful-tools/quick-sharun.sh#L1345-L1349)

uh oh

cc @FabianLars https://github.com/FabianLars/Anylinux-AppImages/commit/3e280d1b2270fecfcb2c2c823b490c782ecf1277

Something is very wrong here because I used the same trick [here](https://github.com/Samueru-sama/simple-appimage-sandbox/blob/bc5802c49ab7d0721aafaea600fe9ebe9f3d1ca4/sas.sh#L195-L207)

And I tested it sandboxing the `xfce4-terminal -x` to run a long complicated `sh -c` where there was multiple layers of quotes and spaces and it handled it correctly.

And we have deployed apps that contain spaces in the libdir before as well, [like this one](https://github.com/pkgforge-dev/Stirling-PDF-AppImage/blob/355eb19ae8f7ff3005514afe85cb8f9beed3ad76/make-appimage.sh#L15)

image

FabianLars:
> cc @FabianLars

Right I wanted to reach out to you about that but forgot. I know my commit is shit but I don't get bash and this was a time sensitive hotfix for a customer. If you have a better idea then I am all ears, maybe Lucas' Claude commits fixed that as well, didn't check.

I can get you the build logs in probably 3 weeks when I'm back in office 🙃

chrox:
@FabianLars no worries — and thanks for saying so. The better idea is a one-liner: quote the positional arguments through the same helper the array already uses,

```sh
eval set -- "$TO_DEPLOY_ARRAY" "$(_save_array "$@")"
```

That keeps both halves. Reproduction in dash with `libextra.so` in the array and `My App/libfoo.so` positional:

| | array entries | positional path with a space |
|---|---|---|
| `eval set -- "$TO_DEPLOY_ARRAY" "$@"` (pkgforge) | ok | split into `My` and `App/libfoo.so` |
| `set -- $TO_DEPLOY_ARRAY "$@"` (3e280d1b) | kept as `'…/libextra.so'`, quotes and all — every `DEPLOY_*` extra skipped | ok |
| `eval set -- "$TO_DEPLOY_ARRAY" "$(_save_array "$@")"` | ok | ok |

`_save_array` already escapes embedded single quotes, so paths with quotes survive as well.

@Samueru-sama that is the trick from your sas.sh — the array side was always fine; it's `"$@"` inside the `eval` that splits, which is what 3e280d1b was fixing. The same line applies to pkgforge's `_deploy_libs`.

Since the CLI re-downloads the fork's `main` on every build, landing that line fixes every tauri CEF AppImage on its next build. Until then we're doing what Lucas' branch does — building the CLI with the bundler pointed at pkgforge pinned to facb95e8 + checksum, bash and streamed output (tauri-apps/tauri#12491's approach) — and once that build confirms the tooling deploys NSS and libpulse by itself, the hand-staged entries here go away, leaving only the two things no rule covers: Debian's `<triple>/nss/` layout and libspeechd.

FabianLars:
> quote the positional arguments through the same helper the array already uses,

Does it make a difference that the issue I had wasn't directly the positional arguments? The arguments did not have spaces but were relative paths that had spaces when made absolute (dunno if the script does that, only on my phone so can't check). I can't remember if it also collected libs (eg the cef files) with spaces in their names but I assume so considering this issue is in the collect libs logic.

chrox:
@FabianLars @Samueru-sama Sorry for the noise from the Fable 5 model. I’ll admit that I don’t have much experience with AppImage and the bundler at all. This fix was primarily developed by the Claude Code agent; I only verify if the AppImage produced by the latest Ubuntu GitHub Actions workflow runs successfully on my Ubuntu 24.04.

Samueru-sama:
Oh, I think I see the problem now, I cannot run `eval` directly on the positional array, I have to save the positional array as well before merging them. (which sas does correctly).

Will fix it tomorrow asap. Thanks for the observation @chrox

Samueru-sama:
https://github.com/pkgforge-dev/Anylinux-AppImages/commit/53a05bc37f0d5241fe7ce3c7daee20fd93e26750

PR #6112, opened 2026-09-07
ci(linux): build the CEF tauri CLI from readest/tauri with pkgforge pinned

chrox:
Stacked on #6108 (retarget to `main` once it merges). This is the direction @Samueru-sama and @FabianLars pointed at in that thread: let quick-sharun deploy Chromium's dependencies itself instead of hand-staging them, and take the pinned-pkgforge route of tauri-apps/tauri#12491 rather than the fork.

## Why a self-built CLI

`@tauri-apps/cli-cef@3.0.0-alpha.26` bundles the AppImage with FabianLars' fork of quick-sharun, whose HEAD drops the `eval` that rebuilds the deployment array, so every `DEPLOY_*` extra is silently skipped. It re-downloads that fork on every build and offers no way to point elsewhere: no env var, `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR` only rewrites `github.com` URLs, and the cache seed the workflow used before #6074 is dead because the `if !exists` guard is commented out. There is no newer `cli-cef` on npm, and #12491 has not reached `feat/cef`.

So the Linux legs build the CLI from readest/tauri, the way the workflow did until #6074. readest/tauri@c62e9f90 (`feat/cef-pkgforge-bundler`, +107/−26 on `sharun_cef.rs`) does what #12491's bundler does, on top of the CEF-aware one:

- fetches pkgforge-dev/Anylinux-AppImages at `facb95e8` verified against #12491's SHA-256, cached under the revision, and pins `ANYLINUX_LIB_SOURCE` to the same rev;
- runs the script with bash and streams its output — the two halves of the #4906 hang (dash never creates the process group the script signals; the captured pipes stayed open in surviving CEF children);
- passes the deploy list as arguments and keeps whitespace out of the AppDir name, since that revision's `eval` splits positional paths on it;
- drops `ADD_HOOKS=fix-namespaces.hook`: CEF never asks for user namespaces without the sandbox feature, so the hook only produced the root-password prompt (#6110's shim stays and is now a no-op);
- lifts the Windows-only cfg on `HashAlgorithm::Sha256` (#12491's `http_utils.rs` change).

Pinned to a rev, not the branch tip; rust-cache keeps `~/.cargo/bin` and a marker skips the ~6 min rebuild while the pin holds. `scripts/tauri.mjs` runs `cargo tauri` when `TAURI_CEF_CARGO=1`, same command line as the npm CLI.

## What the tooling now does on its own

Built locally with that CLI (`--bundles appimage`, `STRACE_MODE=0` as in CI). The AppDir quick-sharun produced, with **no** staging step involved:

- audio: `libpulse.so.0`, `libpipewire-0.3.so.0`, `alsa-lib/libasound_module_{conf,ctl,pcm}_pulse.so`, `pulseaudio/` on `lib.path` — via `DEPLOY_CHROMIUM` → `DEPLOY_PIPEWIRE` → `DEPLOY_PULSE`, exactly as Samueru-sama described;
- NSS: `libsoftokn3.so`, `libfreeblpriv3.so`, `libnssckbi.so`, `libnssdbm3.so`;
- GL: mesa EGL/GLX, `dri/`, `gbm`, `vdpau` (and the host's NVIDIA libs, since this box has them);
- also gconv, gio modules, gvfs, gdk-pixbuf loaders, GTK IM modules, p11-kit, spa/pipewire modules; 397 libraries in all;
- hooks: `00-no-userns-prompt`, `01-path-mapping-hardcoded`, `05-vulkan-check` — no fix-namespaces.

The resulting AppImage, run headless with a fresh `$HOME`: alive for the whole window, zero `FATAL`/`GLIBC_`/`nss_error`/`zenity`/`pkexec` lines, and the GPU process does **not** exit. Inside the bundle's own runtime (`LD_DEBUG=libs` under its loader and `lib.path`):

```
calling init: <bundle>/lib/libpulse.so.0
calling init: <bundle>/lib/pulseaudio/libpulsecommon-17.0.so
calling init: <bundle>/lib/libsoftokn3.so
calling init: <bundle>/lib/libGL.so.1
```

## What the staging step keeps

Down to the two gaps no rule covers, passed to quick-sharun as arguments the way the script itself prescribes rather than copied in:

- Debian keeps the NSS PKCS#11 modules in `<triple>/nss/` and the globs only know Arch's flat layout, so the runner copies them up a level for `"$LIB_DIR"/libsoftokn3.so*` to match (this box, Ubuntu 25.04, ships them flat, which is why the local build found them unaided; jammy does not);
- `libspeechd.so.2`, which nothing deploys and without which the Web Speech engine has no voices.

The libpulse and NSS `appimage.files` entries are gone; the guard step still checks the full set lands.

## Not in this PR

`STRACE_MODE` stays off. The hang it works around is fixed by the bash/streamed-output change, but turning dlopen discovery back on is its own change with its own validation. Cannot be exercised by PR CI — it needs a full CEF build — so the first nightly after merge is the CI validation.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -132,6 +132,31 @@ jobs:
         with:
           key: nightly-${{ matrix.config.os }}-${{ matrix.config.release }}-${{ matrix.config.arch }}-cargo
 
+      # The published @tauri-apps/cli-cef (3.0.0-alpha.26, pinned in
+      # apps/readest-app/scripts/tauri.mjs) bundles the AppImage with a fork of
+      # quick-sharun whose deployment array is broken — see the staging step
+      # below — and re-downloads that fork on every build with no way to point
+      # it elsewhere. The Linux legs build the CLI from readest/tauri instead,
+      # where sharun_cef.rs fetches a pinned, checksummed pkgforge revision and
+      # runs the tooling with bash and streamed output (the fix behind the
+      # STRACE_MODE hang, tauri-apps/tauri#12491). Pinned to a rev, not the
+      # branch tip, so a release cannot change behaviour under us; rust-cache
+      # keeps ~/.cargo/bin, and the marker file skips the ~6 minute rebuild
+      # while the pin is unchanged. TAURI_CEF_CARGO=1 in the build step makes
+      # scripts/tauri.mjs run `cargo tauri` with the same command line.
+      - name: install the CEF tauri CLI from readest/tauri (linux only)
+        if: matrix.config.release == 'linux'
+        run: |
+          set -euo pipefail
+          rev=c62e9f90c1f5759a4e3795cc56b0b9a1273a5773
+          marker="$HOME/.cargo/bin/.cargo-tauri-rev"
+          if [ -x "$HOME/.cargo/bin/cargo-tauri" ] && [ "$(cat "$marker" 2>/dev/null)" = "$rev" ]; then
+            echo "cargo-tauri at $rev already installed"
+            exit 0
+          fi
+          cargo install tauri-cli --git https://github.com/readest/tauri --rev "$rev" --locked --force
+          echo "$rev" > "$marker"
+
       - name: install dependencies (ubuntu only)
         if: contains(matrix.config.os, 'ubuntu') && matrix.config.release != 'android'
         run: |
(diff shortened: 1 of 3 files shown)


chrox:
Folded into #6108.

Issue #5911, opened 2026-08-28
Clean install followed by WebDAV full sync can overwrite existing book groups


Zack9476:
### Bug Description

After reinstalling Readest on my BOOX device, some existing book group assignments were lost during the initial synchronization.
Before reinstalling, all books had already been organized into groups, and the group information was available on my other devices.
After the BOOX synchronization, the affected group assignments were also removed from my other devices.

### Steps to Reproduce

1.Maintain an existing library synchronized through WebDAV.
2.Organize all books into groups and complete synchronization.
3.Uninstall Readest from the BOOX device.
4.Reinstall Readest.
5.Sign in to the Readest account before configuring WebDAV.
6.At this point, Readest Cloud is used by default and begins restoring/downloading some books.
7.Shortly afterward, configure and enable WebDAV.
8.Select “Send and Receive”.
9.Enable “Full Sync” and start synchronization.
10.Check the groups on the BOOX device.
11.Allow the PC and phone to synchronize again.

### Actual result

Some books lose their group assignments on the newly installed BOOX device.

After the other devices synchronize, the same group assignments are also removed from those devices.

I inspected the WebDAV library.json after the problem occurred:

181 live book records

161 books still contain group information

20 books contain neither groupId nor groupName

19 affected records received sequential updatedAt timestamps within approximately eight seconds

The timestamps strongly suggest that the affected records were rewritten during one initial batch restoration or download operation.

Therefore, this is not only a BOOX UI issue. The ungrouped state was written into the shared WebDAV index and then propagated to the other devices.

### Expected Behavior

Existing group assignments should be preserved during full synchronization and when restoring the library on a newly installed device.

A device with stale or missing group information should not overwrite newer valid group assignments from another device.

### Possible cause

On a clean installation, Readest Cloud is used as the default library synchronization provider before WebDAV is configured.

In this case, Readest Cloud initially restored some older book records that apparently did not contain the latest group assignments. Downloading or restoring those books appears to update the book-level updatedAt timestamp, even when their metadata is incomplete or stale.

When WebDAV was subsequently enabled and a full two-way synchronization was performed, these recently updated but ungrouped local records were considered newer than the correctly grouped WebDAV records.

Because group membership is resolved using the same book-level updatedAt timestamp, the missing groupId and groupName values could overwrite the valid WebDAV values. The resulting ungrouped records were then downloaded by the other devices.

[...]

PR #6116, opened 2026-09-07
fix(sync): repair missing local covers from the file-sync remote

### LINKED ISSUE #5931: S3 Full Sync does not restore covers for books previously restored by Readest Cloud
### Bug Description

When reinstalling Readest, I found an issue with book covers when switching from the default Readest Cloud sync to S3 file sync.

My normal setup uses S3 for book files. I do not upload book files to Readest Cloud, so Readest Cloud only contains the library/book metadata.

After reinstalling Readest and signing in, Readest Cloud sync runs first and restores the books to the local library. At this point, the books exist as metadata-only shelf entries:

- Book metadata is present
- Book files are not available locally
- Covers are not available locally

I then configure my existing S3 storage and run a **Full Sync**.

The S3 storage already contains the corresponding book files and covers, but Full Sync does not restore the missing covers.

Image

### Steps to Reproduce

1. Have an existing Readest library synced to Readest Cloud.
2. Do not upload the actual book files to Readest Cloud.
3. Store the complete library in S3, including book files and covers.
4. Uninstall Readest or clear the app data.
5. Reinstall Readest.
6. Sign in to the Readest account.
7. Readest Cloud restore the library.
8. The books appear in the library, but without their actual covers or local book files.
9. Configure the existing S3 file-sync provider.
10. Run **Full Sync**.
11. Observe that the book covers are still not restored.

### Expected Behavior

After configuring S3 and running Full Sync:

- Existing metadata-only library entries should be matched with the corresponding books in S3.
- Missing local covers should be downloaded from S3.
- The actual EPUB/PDF files should remain remote until the user chooses to download them.
- Once a cover has been downloaded, the library UI should refresh immediately without requiring the user to open the book first.

In other words, the expected state after Full Sync should be:

- Book metadata: local
- Cover: local
- Book file: remote/on-demand

## Actual Behavior

After Readest Cloud has already restored the book metadata:

- S3 Full Sync does not restore the missing covers.
- The books remain displayed with generated placeholder covers.

## Possible Cause

From looking at the current file-sync logic, this may be related to how existing local library entries are handled.

After Readest Cloud restores the library, the affected books already exist in the local book map by hash.

Therefore, when S3 sync runs, these books are no longer treated as remote-only books that need to be added to the local library.

## Suggested Fix

I think the cover state should be handled independently from whether the book entry already exists in the local library.

Currently, a book restored by Readest Cloud may already exist in the local library by hash, but that does not necessarily mean that its local resources are complete.

For example:
Local library entry: exists
Local cover: missing
Local book file: missing

S3 library entry: exists
S3 cover: exists
S3 book file: exists
In this state, S3 Full Sync should not treat the existing local library entry as meaning that the cover is already available.
1. Full Sync should repair missing covers independently
During Full Sync, for every live book that exists in the S3 index, Readest could check whether the local cover actually exists.

2. Existing local entries should not block S3 resource recovery
A book being present in allBooksMap only proves that the local database contains a book entry.
It does not prove that:
cover.png exists
book file exists
config exists
So allBooksMap.has(hash) should not prevent S3 from repairing missing resources.

3. Cover download success should update state only after the file is actually saved

### Reproduction File

_No response_

### Operating System

Android

### OS Version

Android 16

### Readest Version

0.12.6

### Custom CSS

_No response_

jadhavgaurav:
## Problem

Fixes #5931.

When Readest Cloud holds only the library metadata and the book files live in an S3/WebDAV file-sync remote, a fresh install restores the books as metadata-only shelf rows: no book file, no cover. Running **Full Sync** against the provider that *does* have the covers never restores them, so the affected books keep showing generated placeholders.

## Root cause

Once the row exists locally, its presence in `allBooksMap` was treated as proof that the local resources were complete:

- **Discovery** only materialises hashes the library doesn't already have (`if ((!local || revivesLocalTombstone) && !rb.deletedAt)` and `if (!allBooksMap.has(entry.name))`), so it never looks at these rows and its cover pull never runs.
- **Metadata reconciliation** only fires when the indexed copy is strictly newer, which it usually isn't.
- The **push pass** calls `pushBookCover`, which finds no local cover and resolves to `no-source`. Nothing ever pulls.

So the cover exists on the remote, the row exists locally, and no code path connects them. As the issue puts it: being in the library only proves the *row* is there, not the cover.

## Fix

Full Sync now runs a pull-side **cover repair** pass (after deletion propagation, before the push pass): for every live row in the remote index whose local cover file is missing, pull the remote `cover.png`, save it, stamp `coverDownloadedAt`, and persist the row via `store.updateBookMetadata` so the bookshelf regenerates the cached cover URL immediately instead of showing the placeholder until the next reload.

It is gated on `fullSync && canPull`:

- **Receive Only heals too** — the repair does not depend on the push loop, which `'receive'` never enters.
- **The incremental path stays O(changed)** — Full Sync is the audit/repair pass in this engine.
- The push pass's HEAD then matches the freshly written local copy, so nothing is bounced back up.

Deliberately unchanged: book files stay remote/on-demand, and `'send'` (Send Only) still never pulls.

### Cost

One GET per cover-less row on Full Sync only; a row with a local cover costs nothing. A row the remote has no cover for either pays a single 404 GET per Full Sync.

## Tests

`src/__tests__/services/sync/file/engine-cover-repair.test.ts` (6 cases; the first written against unmodified `main` and confirmed failing — `saveBookCover` was called 0 times):

- Full Sync pulls a remote cover the local row is missing, and writes the row back.
- A cover the remote does not have costs one GET and writes nothing back.
- `'send'` never pulls a cover down.
- A book that already has its cover locally is untouched.
- Receive Only Full Sync repairs the cover (fails when the repair lives in the push loop).
- Incremental sync leaves the repair to Full Sync (fails if the `fullSync` gate is removed).

The #5911 field-repair test now gives its fake store a local cover so it keeps asserting what it meant: an index-field repair never pulls a cover that is already on the device.

## Checks

- `pnpm test` — clean
- `pnpm lint` (tsc + biome) — clean

## Note

Verified from source against the sync engine's test harness, not on a physical Android device; the reporter's S3 setup is what the fake provider models.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed full synchronization so missing local book covers are restored when available remotely.
  * Preserved send-only behavior by preventing remote covers from being downloaded in that mode.
  * Avoided unnecessary downloads when a local cover is already present.
  * Covers unavailable remotely are not saved locally.
  * Incremental synchronization continues to leave cover repair to full synchronization.
  * Added synchronization reporting for the number of covers restored.

### DIFF
--- a/apps/readest-app/src/__tests__/services/sync/file/engine-cover-repair.test.ts
+++ b/apps/readest-app/src/__tests__/services/sync/file/engine-cover-repair.test.ts
@@ -1,0 +1,183 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+
+import type { Book, BookConfig } from '@/types/book';
+import { FileSyncEngine } from '@/services/sync/file/engine';
+import type { FileSyncProvider } from '@/services/sync/file/provider';
+import type { LocalStore } from '@/services/sync/file/localStore';
+import type { RemoteLibraryIndex } from '@/services/sync/file/wire';
+
+/**
+ * Cover repair for rows that already exist locally (#5931).
+ *
+ * A book restored by Readest Cloud lands in the library as a metadata-only
+ * row: no book file, no cover. Being in `allBooksMap` used to be treated as
+ * proof that the local resources were complete, so the file-sync discovery
+ * pass skipped it and the push pass only ever tried to upload a cover this
+ * device didn't have. These tests pin the repair: when the remote holds a
+ * cover the local row is missing, sync pulls it down.
+ */
+
+const COVER_PATH_SUFFIX = 'cover.png';
+
+const makeBook = (overrides: Partial<Book> = {}): Book => ({
+  hash: 'h1',
+  format: 'EPUB',
+  title: 'Restored Book',
+  sourceTitle: 'Restored Book',
+  author: 'Author',
+  createdAt: 1,
+  updatedAt: 100,
+  ...overrides,
+});
+
+const makeIndex = (books: Book[]): RemoteLibraryIndex => ({
+  schemaVersion: 1,
+  updatedAt: 100,
+  books,
+  uploadedHashes: books.map((b) => b.hash),
+});
+
+/**
+ * Provider routed by path. `remoteCover` null models a remote that never
+ * received a cover for this hash, so both the HEAD probe and the GET 404.
+ */
+const makeProvider = (
+  index: RemoteLibraryIndex | null,
+  remoteCover: ArrayBuffer | null,
+): FileSyncProvider => ({
+  rootPath: '/',
+  readText: vi.fn(async (path: string) =>
+    path.endsWith('library.json') && index ? JSON.stringify(index) : null,
+  ),
+  readBinary: vi.fn(async (path: string) =>
+    path.endsWith(COVER_PATH_SUFFIX) ? remoteCover : null,
+  ),
+  head: vi.fn(async (path: string) =>
+    path.endsWith(COVER_PATH_SUFFIX) && remoteCover
+      ? { size: remoteCover.byteLength, etag: 'cover-etag' }
+      : null,
+  ),
+  list: vi.fn(async () => []),
+  writeText: vi.fn(async () => {}),
+  writeBinary: vi.fn(async () => {}),
+  ensureDir: vi.fn(async () => {}),
+  deleteDir: vi.fn(async () => {}),
+});
+
+const makeStore = (overrides: Partial<LocalStore> = {}): LocalStore => ({
+  loadConfig: async (): Promise<BookConfig> => ({ updatedAt: 50, booknotes: [] }),
+  saveBookConfig: async () => {},
+  loadBookFile: async () => null,
+  resolveLocalBookPath: async () => null,
+  saveBookFile: async () => {},
+  prepareLocalBookPath: async () => '/local/path',
+  // The defining state of the bug: the row is in the library, the cover file
+  // is not on disk.
+  loadBookCover: async () => null,
+  saveBookCover: async () => {},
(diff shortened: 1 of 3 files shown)


chrox:
Thanks @jadhavgaurav — the diagnosis is exactly right, and I've pushed a follow-up commit on top of yours rather than sending you back and forth.

**What changed and why.** The repair lived inside the `canPush` push loop, so a device set to **Receive only** (a real option in the file-sync form, and exactly the mode a secondary device restores in) never ran it — a Receive Only + Full Sync still reported `coversDownloaded: 0`. That's the CodeRabbit finding, and I confirmed it with a probe before touching anything.

The follow-up moves the repair to the pull side as its own Full Sync pass, right after deletion propagation: for every live indexed row whose local cover is missing, GET the remote cover, save it, stamp `coverDownloadedAt`, and persist through `updateBookMetadata` (your write-back logic is kept as is). Gated on `fullSync && canPull`, so:

- Receive Only heals too.
- The incremental path stays O(changed) — Full Sync is the audit/repair pass in this engine.
- The push pass's HEAD then matches the freshly written local copy and never bounces it back up, so `PushBookFileResult.remoteExists` is no longer needed and is dropped.

**Cost.** Same as before for the default strategy: one GET per cover-less row. The one difference is a row the remote has no cover for either now pays a single 404 GET per Full Sync instead of nothing — I've flipped that test's assertion to pin "one probe, no write-back" rather than "no GET", since the alternative was two code paths for the same repair.

**Tests.** Your four cases are kept (one re-asserted as above), plus a Receive Only case and an incremental-path guard. I mutation-checked the guard: removing the `fullSync` gate makes exactly that test fail.

PR #6168, opened 2026-09-10
fix(reader): keep sliders mirrored when a book turns the layout RTL

### LINKED ISSUE #6157: Bug: Text options controls are reversed in RTL EPUBs
### Bug Description

In RTL EPUBs (like Arabic), the controls in the text options panel opened from the bottom behave in the opposite direction.

### Steps to Reproduce

1. Open an Arabic EPUB (I attached one to test on)
2. Open the text options panel from the bottom.
3. Adjust a left/right control, such as moving the pointer or slider.

Arabic epub for test: https://drive.google.com/file/d/1ULKUPwggjjedzNSt8eDP8m4HZGNJSDax/view?usp=drivesdk

### Expected Behavior

**Actual result:**
Moving the control to the left moves the pointer/value to the right, and moving it to the right moves it to the left.

**Expected result:**
The control direction should match the user's gesture. Moving left should move the value/pointer left, and moving right should move it right, regardless of the EPUB being RTL.

Recorded video: https://drive.google.com/file/d/1DaRwtAzDF_QKZKok0OSl6ZSeelWGwMkK/view?usp=drivesdk

### Reproduction File

_No response_

### Operating System

Android

### OS Version

16

### Readest Version

0.12.8

### Custom CSS

_No response_

jadhavgaurav:
## Problem

In an RTL book, the controls in the text options panel move against the drag:
dragging a slider left moves the thumb and value right, and dragging right
moves them left.

## Cause

`Slider` resolved the surrounding text direction in JS — a one-shot ancestor
walk for `dir='rtl'` in a `useEffect` with an empty dependency array — and
mirrored its own track and thumb from the result, anchoring them to `right`
instead of `left`.

That value went stale. The footer bar that hosts these sliders renders
`dir={viewSettings?.rtl ? 'rtl' : 'ltr'}`, and `viewSettings.rtl` is not known
at mount: `FoliateViewer` derives it from the first document's writing
direction (plus `bookDoc.dir` and `getDirFromUILanguage()`) on the `load`
event, then writes it back with `setViewSettings`. The panels are always
mounted — they are only translated out of view — so the one-shot check runs
while the bar is still `ltr` and never runs again.

The native `<input type="range"></input>` inside the slider follows the inherited
direction and mirrors itself. Confirmed in Chromium: from 50, ArrowLeft gives
49 under `dir='ltr'` and **51** under `dir='rtl'`. So after the switch the input
counted up leftwards while the fill and thumb still grew from the left. The two
disagreed, and the control read as reversed.

## Change

Place the fill and the thumb with logical properties — `inset-inline-start` —
so the browser mirrors them from the same inherited direction the native input
reads. There is nothing left to resolve, nothing to go stale, and the component
no longer sets its own `dir`.

`transform: translateX(-50%)` has no logical form and would have to flip sign
by hand, so the thumb is centered with a negative `margin-inline-start`; the
bubble is exactly `heightPx` wide, so the two are equivalent.

Net **−11 lines** in `Slider.tsx`: the `isRtl` state, the ref, the ancestor walk
and the `dir` attribute are all gone.

## Tests

Direction coverage moves to `slider-rtl-direction.browser.test.tsx`, because
jsdom has no layout — a mirrored box and an unmirrored one are indistinguishable
there, and assertions on the inline `left`/`right` only read back what the
component had just written. The new cases measure real boxes in Chromium. Two
of the four fail against `main`:

- the surrounding direction turning RTL **after** mount
- ArrowLeft under RTL — the mirrored input raises the value, and the thumb must
  still travel left. On `main` it travelled right, which is the reported bug
  stated as an invariant.

`Slider.test.tsx` keeps the geometry pins (now on `insetInlineStart`) and gains
a guard that nothing in the component names a physical side or sets its own
`dir`.

## Verification

- `pnpm test` — 11078 passed, 16 skipped, 0 failed
- `pnpm test:browser` — 469 passed
- `pnpm lint` (tsc + biome) — clean
- **Live in the reader**: with a book open, flipping the footer bar's `dir` to
  `rtl` the way `setViewSettings` does mirrors all four real sliders exactly —
  thumb 81.8px from the left becomes 81.8px from the right, fill re-anchors —
  **with no React re-render**, since the mirroring is now pure CSS inheritance.

Not verified on a device with an Arabic EPUB.

## Note on the original submission

@jadhavgaurav's diagnosis of the mechanism was correct and is what made this
fixable — the stale one-shot read, and the native input mirroring underneath it.
The original fix re-ran that ancestor walk on every render. This replaces it
with the CSS equivalent, which also covers the case where the direction changes
without the slider re-rendering. One detail in the original write-up was off:
`viewSettings` is never undefined when the footer renders (`BooksGrid` returns
`null` until it exists); the post-mount switch comes from `FoliateViewer`
computing `rtl` on document load.

Fixes #6157

## Summary by CodeRabbit

* **Bug Fixes**
  * Slider thumb and fill positioning now remains synchronized with the native range control in both left-to-right and right-to-left layouts.
  * Switching layout direction after the slider loads now correctly updates visual positioning and keyboard behavior.

* **Tests**
  * Added browser coverage for mirrored positioning and dynamic direction changes.

### DIFF
--- a/apps/readest-app/src/__tests__/components/Slider.test.tsx
+++ b/apps/readest-app/src/__tests__/components/Slider.test.tsx
@@ -43,6 +43,63 @@ describe('Slider', () => {
     expect(thumb.style.left).toBe('calc(75% - 11px)');
   });
 
+  it('mirrors the track when the surrounding direction turns right-to-left after mount', async () => {
+    // The reader's footer bar renders dir='ltr' until the book's view settings
+    // load, then switches to 'rtl' for a right-to-left book. The native range
+    // input follows that change, so the track and thumb have to follow it too.
+    const { container, getByRole, rerender } = render(
+      <div dir='ltr'>
+        <Slider label='Reading Progress' initialValue={100} heightPx={44} />
+      </div>,
+    );
+    getByRole('slider');
+    expect(container.querySelector<HTMLElement>('.slider-thumb')!.style.left).toBe(
+      'calc(100% - 22px)',
+    );
+
+    rerender(
+      <div dir='rtl'>
+        <Slider label='Reading Progress' initialValue={100} heightPx={44} />
+      </div>,
+    );
+
+    await waitFor(() =>
+      expect(container.querySelector<HTMLElement>('.slider-thumb')!.style.right).toBe(
+        'calc(100% - 22px)',
+      ),
+    );
+    expect(container.querySelector<HTMLElement>('.slider-thumb')!.style.left).toBe('');
+    expect(container.querySelector<HTMLElement>('.slider-fill')!.style.right).toBe('0px');
+  });
+
+  it('restores the left edge when the surrounding direction turns back to left-to-right', async () => {
+    const { container, getByRole, rerender } = render(
+      <div dir='rtl'>
+        <Slider label='Reading Progress' initialValue={100} heightPx={44} />
+      </div>,
+    );
+    getByRole('slider');
+    await waitFor(() =>
+      expect(container.querySelector<HTMLElement>('.slider-thumb')!.style.right).toBe(
+        'calc(100% - 22px)',
+      ),
+    );
+
+    rerender(
+      <div dir='ltr'>
+        <Slider label='Reading Progress' initialValue={100} heightPx={44} />
+      </div>,
+    );
+
+    await waitFor(() =>
+      expect(container.querySelector<HTMLElement>('.slider-thumb')!.style.left).toBe(
+        'calc(100% - 22px)',
+      ),
+    );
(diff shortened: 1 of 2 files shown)


JackyHe398:
Sorry I have encountered a similar issue recently and I have discovered a tiny problem.
Did your PR considered the also OPF rtl tag or only the html div rtl tag? Cuz I have seen a few books where the page direction can be differ across the book, which the method you describe might result in the slider flipping when we turn a page. The logic might also need to consider the tag from OPF, then fall back to html tag if not present, in order not to mess up with a higher level tag. 

I have a similar PR #6197 in draft pending a upstream merge. You might want to take a reference.

PR #6188, opened 2026-09-12
feat(about): show the real WebView build number instead of the UA-reduced stub

k6G52m4Dz75W:
## Summary

The version shown in **About** and on the error page — and the `webview.engine` / `webview.version` Sentry tags — come from parsing the User-Agent string. On Chromium engines, UA Reduction freezes the minor/build/patch tokens in that string, so a Windows WebView2 user sees:

```
Version 0.12.8 (Edge 152.0.0.0)
```

…no matter which actual runtime build they are on. The real build number is only available through the User-Agent Client Hints `fullVersionList` (high-entropy). This PR resolves it everywhere the version is shown or reported, and tightens the parsing on the Rust side to match.

## Changes

**Version resolution (`utils/ua.ts`)**

- `getWebViewFullVersion()` — resolves the engine's entry in the Client Hints `fullVersionList`; returns `null` on WebKit engines (no Client Hints; their UA already carries the real version) and is SSR-guarded.
- `parseWebViewInfoAsync()` — the existing UA-parsed label with its trailing version upgraded to the real build; falls back to the sync label verbatim when Client Hints are unavailable.
- `withWebViewFullVersion(ua, build)` — the single definition of "splice the real build into a UA string", matched to the brand's own token (a WebView2 UA carries both a reduced `Chrome/` and an `Edg/` token; the rewrite must land on the one the brand version came from, not the first match).

**Consumers**

- `AboutWindow` and the error page render through a shared `useWebViewInfo` hook: the label starts at the UA-parsed value and settles to the full build after the Client Hints round-trip, with mounted-guarded state updates. The copy-to-clipboard in About automatically carries the full build.
- The copied version string is now locale-neutral and self-identifying: `Readest 0.12.8 (WebView2 152.0.4191.66)` — previously it pasted a localized "Version …" prefix into issues and never named the app.
- `nativeAppService.init()` reports the UA through `withWebViewFullVersion` before invoking `set_webview_info`.

**Rust (`sentry_config.rs`)**

- `parse_webview_info` keeps the **full** version string (it previously truncated to the major, which would have made the UA rewrite pointless) and recognizes the `Edg/` token as its own **`WebView2`** engine — a WebView2 UA also carries a reduced `Chrome/` token that must not win, and "Edge" read as the user-installed Edge browser while the component's actual name is WebView2 (matching the `WebView <version>` naming already used for the Android/iOS system WebViews).
- `ua_token_version` rejects malformed tokens: empty, leading dot (`Chrome/.5`), or trailing dot (`Chrome/140.`) never reach the `webview.version` tag.

## Platform behavior

| Platform | Before | After |
|---|---|---|
| Windows (WebView2) | `Edge 152.0.0.0` (frozen stub) | `WebView2 152.0.4191.66` (real build via Client Hints) |
| Android (system WebView) | `WebView 124.0.6367.219` | same build, now also cross-checked against `fullVersionList` (available and matching — verified) |
| iOS / macOS (WKWebView) | `WebView 605.1.15` | unchanged — no Client Hints on WebKit; fallback is verbatim |
| Linux | UA-parsed (WebKitGTK UA tokens are frozen upstream) | Chromium/CEF runtimes get Client Hints; WebKitGTK unchanged |

Copied bug-report string (all platforms): `Readest 0.12.8 (WebView2 152.0.4191.66)`.

## Verification

- New unit tests: Client Hints brand match / missing API / no matching brand / rejected `getHighEntropyValues`, label upgrade and WebKit fallback, brand-token splice targeting (`Edg` not the reduced `Chrome`), no-op on tokenless UAs; Rust tests assert the full builds, the `WebView2` engine branch winning over the reduced `Chrome/` token, and malformed-token rejection (27 frontend + 19 Rust tests green; tsc and biome clean).
- Android emulator (system WebView 124): `getHighEntropyValues(['fullVersionList'])` resolves and reports `Android WebView | 124.0.6367.219`, matching the UA token (Android WebView UAs are not reduced); About label and copy string verified on device.
- Windows dev build (WebView2 152): About shows `WebView2 152.0.<build>` with the real build from Client Hints; the `.0.0` stub is gone.

The diff went through four internal review rounds (15 findings, all resolved — including two real bugs the first round caught in its own Sentry path: a token-targeting mismatch and a Rust parse that truncated to the major, plus an SSR-safety regression introduced while extracting the hook); the fifth round returned zero findings.

## Summary by CodeRabbit

* **Improvements**
  * Browser and WebView information now shows more accurate, complete version numbers.
  * Windows WebView2 is identified with a dedicated “WebView2” label.
  * About and error screens use consistent browser information and version formatting.
  * Version details are enhanced when additional browser information is available, with a reliable fallback when unavailable.
  * Browser information remains safe and stable when details are missing or malformed.

### DIFF
--- a/apps/readest-app/src-tauri/src/sentry_config.rs
+++ b/apps/readest-app/src-tauri/src/sentry_config.rs
@@ -142,7 +142,7 @@ pub fn webview_info() -> Option<&'static (String, String)> {
 /// `None` if neither token family is present.
 pub fn parse_webview_info(user_agent: &str) -> Option<(String, String)> {
     if let Some(v) = ua_token_version(user_agent, "Edg/") {
-        return Some(("Edge WebView2".to_string(), v));
+        return Some(("WebView2".to_string(), v));
     }
     if let Some(v) = ua_token_version(user_agent, "Chrome/") {
         return Some(("Chromium".to_string(), v));
@@ -377,7 +377,7 @@ mod tests {
                   Chrome/138.0.0.0 Safari/537.36 Edg/138.0.3351.62";
         assert_eq!(
             parse_webview_info(ua),
-            Some(("Edge WebView2".to_string(), "138.0.3351.62".to_string()))
+            Some(("WebView2".to_string(), "138.0.3351.62".to_string()))
         );
     }
 
--- a/apps/readest-app/src-tauri/src/sentry_config.rs
+++ b/apps/readest-app/src-tauri/src/sentry_config.rs
@@ -159,11 +159,11 @@ fn ua_token_version(user_agent: &str, token: &str) -> Option<String> {
         .chars()
         .take_while(|c| c.is_ascii_digit() || *c == '.')
         .collect();
-    if version.is_empty() {
-        None
-    } else {
-        Some(version)
-    }
+    // Trim stray trailing dots (`Chrome/140.`) and require at least one digit
+    // so a malformed token never reaches the webview.version tag.
+    let version = version.trim_end_matches('.');
+    (!version.is_empty() && version.contains(|c: char| c.is_ascii_digit()))
+        .then(|| version.to_string())
 }
 
 /// C-ABI accessor for the compile-time Sentry DSN, used by the iOS native
--- a/apps/readest-app/src-tauri/src/sentry_config.rs
+++ b/apps/readest-app/src-tauri/src/sentry_config.rs
@@ -129,28 +129,40 @@ pub fn webview_info() -> Option<&'static (String, String)> {
     WEBVIEW_INFO.get()
 }
 
-/// Parse the WebView engine and major version from a User-Agent string. Chromium
+/// Parse the WebView engine and version from a User-Agent string. Chromium
 /// WebViews (Android System WebView, Windows WebView2, Linux Chrome) carry a
-/// `Chrome/<v>` token; WebKit ones (iOS/macOS WKWebView, Linux WebKitGTK) carry
-/// `Version/<v>` and no `Chrome/`. Chrome is checked first because Android
-/// WebViews also include a legacy `Version/4.0`. `None` if neither is present.
+/// `Chrome/<v>` or `Edg/<v>` token; WebKit ones (iOS/macOS WKWebView, Linux
+/// WebKitGTK) carry `Version/<v>` and no `Chrome/`. Edge is checked first
+/// because a WebView2 UA carries both `Edg/` and a reduced `Chrome/` token,
+/// and Android WebViews include a legacy `Version/4.0` that must not win over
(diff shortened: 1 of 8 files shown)


chrox:
Thanks for this, and for the care that went into it. The self-review rounds, the Rust-side hardening, and the test coverage are all visible in the diff, and the premise is correct: UA Reduction really does freeze the build tokens, so `Edg/152.0.0.0` on a 152.0.4191.66 runtime is exactly the problem you describe.

I'm going to close it, though, because Tauri already exposes the real build directly and readest just never called it.

`tauri::webview_version()` is re-exported at `crates/tauri/src/lib.rs:200` (we're on wry 0.56.1) and returns the actual runtime version on every platform:

- Windows: `GetAvailableCoreWebView2BrowserVersionString` -> the real WebView2 runtime version
- macOS / iOS: the `com.apple.WebKit` bundle's `CFBundleVersion`
- Linux GTK: `webkit_get_major/minor/micro_version()`
- Android: queries the WebView package directly

That's one call inside `sentry_config.rs`, the file that already sets the tag, and it needs no frontend changes at all: no Client Hints round-trip, no UA rewrite, no brand table, no new hook, nothing added to the startup path. It's also strictly more capable than the Client Hints route, which by its own design can't help WebKit. Two caveats worth knowing if you pick it up: Windows appends a channel suffix on non-stable runtimes (`"138.0.3351.62 beta"`), and the Linux CEF build would still need the UA path since `webview_version()` reports WebKitGTK there.

Two things the review turned up in the current approach, so the close isn't arbitrary:

1. `utils/ua.ts` picks the engine **name** by UA branch order but the **version** by `Edg/` presence, and they disagree. On macOS Edge the `Macintosh && Chrome && AppleWebKit` branch (`:62`) wins the label while `clientHintsBrandFor` (`:105`) matches the Microsoft Edge brand, so `parseWebViewInfoAsync` returns `Chrome 138.0.3351.62`: an Edge build attributed to Chrome, a version string that never shipped. Since that label is also the copy-to-clipboard bug-report text, it's more misleading than the reduced stub it replaces. Reproduced in vitest.

2. `nativeAppService.ts:643` awaits `getWebViewFullVersion()` before `loadSettings()`. The `try/catch` handles a rejection but not a promise that never settles, and `EnvContext.tsx:44` documents that specific path as "no error, no UI, just a blank window for the rest of the session." Nothing in `init()` consumes the value, so it shouldn't gate startup.

Separately, on the About dialog: the `_('Version {{version}}')` -> `Readest ${version}` change drops the translated word for every non-English user, and the `<h2>Readest</h2>` directly above it already names the app. If a locale-neutral bug-report string is wanted, the display can stay localized and `handleCopyVersion` can build its own.

None of this is a knock on the work. If you'd like to do the ~10-line Rust version instead, I'd be glad to review it.

PR #6196, opened 2026-09-12
docker: pull minio images from quay.io, Docker Hub removed them

### LINKED ISSUE #6195: Bug: Self-hosted MinIO images (minio/minio, minio/mc) 404 — Docker Hub removed them
### Bug Description

docker/compose.yaml's minio and minio-setup services pull minio/minio and minio/mc from Docker Hub. Both repositories are gone from Docker Hub entirely — docker pull fails with a misleading auth-shaped error even with valid credentials and even for pinned historical release tags.

 Confirmed this isn't a rate-limit/login issue: hub.docker.com/r/minio/minio 404s and the registry API refuses anonymous manifest requests. Other projects hit the same wall and fixed it the same way (homebase-id/odin-core#1730, OpenCTI-Platform/opencti#18242). MinIO still publishes both images to quay.io/minio/minio and quay.io/minio/mc under the same tags, no login needed.

### Steps to Reproduce

1. Follow docker/README.md to set up the self-hosting stack
2. Run docker compose pull (or docker compose up -d)
3. minio and minio-setup fail with: Error response from daemon: pull access denied for minio/minio, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

### Expected Behavior

 docker compose pull/up -d should successfully pull every image in the stack and start it.

### Reproduction File

_No response_

### Operating System

Linux

### OS Version

N/A

### Readest Version

N/A

### Custom CSS

_No response_

thietala:
Fixes #6195 

## Summary by CodeRabbit

- **Chores**
  - Updated the MinIO server and setup services to pull images from Quay instead of Docker Hub.

- **Documentation**
  - Updated the Docker/Podman stack reference to reflect the new image registry locations for both MinIO services.


chrox:
Thanks for the fix.

Issue #5303, opened 2026-07-24
Minor issue with the header display


shuqiao4-png:
The images below show how the header appears when the top margin is set to -39 in both "slide" and "scroll" modes; you will notice that the header does not appear in the area occupied by the phone's notch. Reducing the top margin seems merely to adjust the transparent overlay area where the header was previously located, effectively increasing the overall size of the page area containing the main content.
 located.![Screenshot_2026-07-24-22-17-48-481_com.bilingify.readest.png](https://github.com/user-attachments/assets/982b4149-e69b-4925-9fe1-58ae0ad54fed)![Screenshot_2026-07-24-22-18-27-801_com.bilingify.readest.png](https://github.com/user-attachments/assets/68c95cc1-ba4e-47e2-8520-573d8fcb29cd)

chrox:
If I understand it clearly, what you want is lifting the header section info to the top notch area when you decrease the top margin, is this correct?

shuqiao4-png:
Yes, if possible, I would really like to be able to disable that top safe area. I apologize for this request; my OCD makes it impossible for me to tolerate that blank space. I tried turning off the header—which does eliminate that area during scrolling and makes the reading experience more comfortable—but the downside is that I can't tell which chapter I'm on unless I open the table of contents.😭

chrox:
The request is reasonable.

Issue #4977, opened 2026-07-06
Top bar interfering with text interaction


bincent0929:
## Bug Description

When trying to select at the top of the page, the interaction for the top bar sits above the text on the page, making it impossible to select the text. Even when trying to start highlight from another area towards the text at the top of the page, it still causes problems.

[Here’s a video of the issue](https://drive.google.com/file/d/1YIpmibhTcG6uBa_yS679db1jrx2exguy/view?usp=drivesdk)

## Expected behavior

A clear and concise description of what the expected behavior is.

The top bar’s activation margin **shouldn’t** overlap with where text is displayed on the screen. Text should **always** be selectable by the user if it’s on the screen.

## Technical inputs

Operating System: iPadOS 26.5

Readest
Version 0.11.17 (WebView 605.1.15)

You can tap the center of the screen to pull up the top bar in addition to the bottom bar. Should the app even have an activatable margin at the top for the top bar?

Issue #4089, opened 2026-05-08
Top toolbar of sidebar in reading mode inaccessible


vladrj:
## Bug  Description

When center-clicking while in reading mode, the device status bar appears - then when you proceed to open the left sidebar, its option icons are obscured by the device's status bar, making them inaccessible.

<img width="2915" height="148" alt="Image" src="https://github.com/user-attachments/assets/a620ba63-8fa2-46a3-b774-aaf6833b79da" />

## Expected behavior

The side bar should respect the bottom margin of the device's status bar, when it is visible, like main reading view does.

> **Main reading view behaviour:**
> Without device status bar

<img width="2944" height="308" alt="Image" src="https://github.com/user-attachments/assets/6d5eb326-9992-4285-b199-04a7004a8091" />

> With device status bar

<img width="2938" height="288" alt="Image" src="https://github.com/user-attachments/assets/e1075b78-7cef-435c-8a73-27a78d08ada3" />

## Technical inputs

> Device: Lenovo TB373FU (Landscape View)
> Operating System: Android 16 (WebView 147.0.7727.111)
> Readest Version: 0.10.6

Issue #624, opened 2025-03-18
FR: Issue with Scroll Direction in Vertical Writing Mode Books


doublex640:
Current Situation:
For vertical writing mode books (i.e., writing-mode: vertical-rl), the scroll direction is automatically set to vertical (up and down).

Suggestion:
Change the scroll direction to horizontal for page turning.

https://github.com/user-attachments/assets/08088f96-ed12-4ca4-9125-7c0b6bc1e143

PR #6197, opened 2026-09-12
Fix page-turn direction fallback and add missing applyMarginAndGap for vertical layout

JackyHe398:
Generally speaking this PR fix bugs for vertical layout books to make it more usable. 

fix: page-turn direction fallback -- The current page-turn direction was falling back to page-level determination, which if one book contains both vertical and horizontal layout, the page-flip logic for the two page will be different disregarding the page-progression-direction tag inside opf file. 
fix: add missing applyMarginAndGap for vertical layout -- the margin was not applied after calculating for vertical layout. The commit add the function call to enable the correct margin
test: add relative test cases 

for sample epub, please refer to the [attachment](https://github.com/user-attachments/files/32149457/01.zip)

## Summary by CodeRabbit

- **New Features**
  - Improved page-turn direction handling for right-to-left publications, including mixed writing directions and vertical layouts.
  - Preserved each section’s text direction while applying the appropriate book-level pagination direction.
  - Updated vertical reading layouts when display settings change, including margins, gaps, and content width.
  - Improved touch gestures in vertical layouts, including toolbar toggling and swipe behavior.

- **Tests**
  - Added coverage for RTL pagination, vertical writing modes, touch gestures, and inline sizing behavior.


JackyHe398:
The current PR requires [this](https://github.com/readest/foliate-js/pull/95) to be accepted in order to pass all tests

JackyHe398:
this fix should also fix [issue 6204](https://github.com/readest/readest/issues/6204)

chrox:
Thanks — the diagnosis is right, and both halves of it are real. I pushed a follow-up commit (`d43bb35`) to this branch and rebased readest/foliate-js#95 onto `main`. Summary of what changed and why.

### The book direction must not override the reader's own setting

`bookDoc.dir === 'ltr'` was hard-overriding `viewSettings.writingMode`, which is an explicit choice the reader made in Settings. Forcing **Vertical (RL)** on a book whose OPF says `page-progression-direction="ltr"` (common in InDesign exports) gave `rtl = false` with `vertical = true`, and from there:

- `viewPagination` stopped swapping left/right (`usePagination.ts:135`),
- `ReadingRuler` mapped columns left→right over right→left ones (`ReadingRuler.tsx:586-605`),
- the progress bar reversed,
- the paginator's vertical drag-follow `forwardSign` flipped turns backwards.

The precedence now lives in one place, `getPageProgressionRTL` in `libs/document.ts`, with a test: **user writing mode → spine → document → UI language**. `default` and a missing attribute both fall through to the document, as before.

### The columns had to turn around too

Pinning `#rtl` to the book made the paginator set `dir="rtl"` on the scroll container while the iframe's columns still ran left-to-right, so an LTR section of an RTL book opened on its **last** page and `expand()` collapsed it to one page. That case is reachable from your own sample — a `vertical-rl` + `horizontal-tb`/LTR mix in an `rtl` book. Fixed in foliate-js#95; the details are in a comment there.

### Margins and inline size

Good catch, and `viewSettings?.vertical` really was missing from the deps of the effect that owns `applyMarginAndGap` (`FoliateViewer.tsx:1078`). Moved the fix there rather than calling it imperatively from `docLoadHandler`, and gave `max-inline-size` its own effect — the old imperative version also set it for pre-paginated books, which `openBook` deliberately doesn't.

### Two things restored

- `src/__tests__/utils/insets.test.ts` lost 156 lines — the whole `getHeaderBandGeometry` (#5303), `getHeaderTriggerHeight` (#4977) and `getPanelTopInset` (#4089) suites — replaced by two unrelated `getViewInsets` cases. CI stayed green because the tests simply vanished. This looks like a bad merge rather than intent.
- `flake.lock` picked up your fork's auto-update through the `update_flake_lock_action` merge; reverted to `main`'s.

### Verification

`pnpm test` 11091 passing · `pnpm lint` clean · `pnpm format:check` clean · `pnpm test:browser` 60 files, 475 passing. The new `paginator-page-progression-direction.browser.test.ts` fails 3 of its 8 cases against unmodified `main`.

Not device-tested against `01.zip` yet — if you get a chance to run this branch against your sample, that would be the last thing I'd want to see before merging. foliate-js#95 needs to land first, then the submodule gets re-pinned to `main`.

chrox:
Merged `main` into this branch (it had gone conflicting) and re-pinned `packages/foliate-js` to readest/foliate-js `main` at `444be90`, the squash merge of foliate-js#95 — byte-identical to the `1286eb3` I pushed earlier.

Verified on Android (Xiaomi 13) with your `01.zip` book: the horizontal front matter and the vertical chapters now page in the same direction.

Locally on the merged tree: `pnpm test` 11104 passing, `pnpm lint` and `pnpm format:check` clean, paginator browser tests 24 files / 158 passing.

chrox:
Re the CodeRabbit finding on `FoliateViewer.tsx:721` (metadata language overwriting the OPF direction) — checked against the code, and declining it:

- **"An LTR metadata language can replace an RTL spine direction" can't happen.** `getBookDirFromLanguage` → `getDirFromLanguage` (`src/utils/rtl.ts`) only ever returns `'rtl'` or `'auto'`, never `'ltr'`, so an LTR-language book leaves `bookDoc.dir` exactly as the OPF set it.
- **The remaining case is an RTL-language book (ar/he/fa/…) whose spine declares `ltr`.** It gets `bookDoc.dir = 'rtl'` and pages right-to-left. That is unchanged from before this PR, where `bookDoc.dir === 'rtl'` was OR'd straight into `rtl`. Letting the spine win there would flip those books to left-to-right paging, a regression for the readers most likely to have a mis-declared spine, so this branch is deliberate.

The precedence this PR codifies (`getPageProgressionRTL`: user writing mode → spine → document → UI language) is untouched; the language hint sits in front of it only for RTL scripts, as it always has.

chrox:
Pushed `d0387af`: swipe up now shows **and** hides the toolbars on vertical books, same as on horizontal ones.

Two leftover gates from when vertical books paged along the block axis were blocking it in `useTouchEvent`. `onTouchEnd` skipped the swipe-up toggle for `vertical`, and `onTouchMove` hid the bars on *any* movement of a vertical book, so the next swipe up re-showed them. Vertical books turn pages with horizontal swipes now (#624), so both gates are gone. New cases are in `useTouchEvent.test.tsx`.

Verified on a Xiaomi 13 with `01 冰菓.epub` on a vertical-rl chapter: show → hide → show with the page untouched, and a horizontal swipe still turns the page and hides the bars.

PR #6229, opened 2026-09-16
fix(reader): open image-heavy sections in paged mode without freezing

### LINKED ISSUE #6155: BUG: An EPUB with a large HTML text file cannot be opened in paged mode in 0.12.8
### Bug Description

In the latest version (0.12.8), an EPUB with Duokan-format footnotes cannot be opened under paged mode, and the whole app would freeze. This issue occurs on both Windows and Android. To be more specific, this issue did not occur on 0.12.6, and this EPUB could still be opened and smoothly read in scrolled mode under the current version.
I tried to recreate the issue with another EPUB file with Duokan footnotes, and nothing happened, so I guess it's because the problematic EPUB puts almost all the text in a single html.

### Steps to Reproduce

Open the sample I provide here under paged mode.

### Expected Behavior

The EPUB file cannot be opened, and the app would freeze.

### Reproduction File

https://drive.google.com/file/d/1VQav8o2rTkDPi5HIheufg8JIHWpmz1hc/view?usp=drive_link

### Operating System

Windows

### OS Version

Windows 11 

### Readest Version

0.12.8

### Custom CSS

_No response_

### LINKED ISSUE #6212: BUG: Performance issue in 0.12.8 with EPUBs in paged mode
### Bug Description

In 0.12.8, Readest struggles (even fails) to open EPUBs that place large quantities of text in a single HTML under paged mode, which did not occur in previous versions. After I divided the large, single HTML into separate pieces, Readest is able to open the new EPUB file almost instantly.
More details could be found in #6155 . 

### Steps to Reproduce

Open the sample EPUB provided here.

### Expected Behavior

Readest would either freeze forever, or take minutes to open the file.

### Reproduction File

https://drive.google.com/file/d/1VQav8o2rTkDPi5HIheufg8JIHWpmz1hc/view?usp=drive_link

### Operating System

Windows

### OS Version

Windows 11

### Readest Version

0.12.8

### Custom CSS

_No response_

chrox:
Closes #6155, closes #6212.

## Symptom

A Duokan-authored EPUB marks every footnote with an inline ``. When such a book keeps a whole volume in one XHTML file, the section reaches thousands of them — and since 0.12.8 opening it in paged mode blocks the main thread for minutes. On Windows and Android it often never recovers. Scrolled mode stays usable because its layout is far cheaper to redo.

## Cause

The paginator's `setImageSize` clamps each replaced element to the page box, walking them one at a time: `removeProperty` the previous clamp, read the element's computed style, write the new clamp.

Since foliate-js#90 that read includes the element's own margins so the clamp can cap the *margin* box. The load-bearing detail: **`marginTop` / `marginBottom` resolve to used values and force layout; `marginLeft` / `marginRight` don't, and neither do `maxWidth` / `maxHeight`** — which is all the pre-#90 code read. So each iteration's read flushed a layout that the previous iteration's write had just dirtied: one full re-fragmentation of the section per image.

## Fix

foliate-js#96 — split the pass into clear → read all → write all, so the whole loop costs one layout. This PR re-pins the submodule and adds the regression test.

> **Merge order:** pinned at `8a21628` on the foliate-js PR branch. Re-pin to the merge commit once readest/foliate-js#96 lands, before merging this.

## Measurements

Chrome, reported book's biggest section (1989 images in a 1.1 MB section), through the real paginator:

| | before | after |
|---|---|---|
| open in paged mode | 27,542 ms | 525 ms |
| margin change | 105,441 ms | 544 ms |
| same book, images stripped (control) | 253 ms | 326 ms |

The control matters for #6212, whose title blames "large quantities of text in a single HTML". That isn't the mechanism — strip the images and the same 1.1 MB single-file book opens in ~250 ms with or without the fix. Section size only multiplies the per-element cost. An image-free book with a huge single section was never affected.

The slow margin changes reported in #6212's comments are the same cause, and worse than opening: on a re-render `removeProperty('max-width')` actually removes a previously-set value, so every iteration dirties layout, where on a first open many of those removals are no-ops.

## Verification

- End to end in the app (`pnpm dev-web`, Chrome): without the fix the renderer goes unresponsive and CDP `Page.captureScreenshot` times out; with it the book opens to page 875/1565 and Duokan footnote popups work.
- New `paginator-image-size-perf.browser.test.ts` paginates a 2000-image section: **23,996 ms before, 882 ms after**, budget 5,000 ms. Fixture is 25 KB (the filler text repeats, so it compresses hard).
- `pnpm format:check`, `pnpm lint`, `pnpm test` (11,098), `pnpm test:browser` (472) all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved paged-mode performance for footnote-heavy EPUBs, reducing freezes during pagination.
  * Improved image layout handling to reduce excessive reflow and improve margin-change responsiveness.

* **Tests**
  * Added a performance regression check for EPUBs containing large numbers of footnote images.
  * Added coverage to verify stable pagination, image sizing, and completion within an expected time limit.


chrox:
Re-pinned to `b9dcd03` now that readest/foliate-js#96 is merged. Its `paginator.js` is byte-identical to the PR-branch commit this branch pinned during review, and `pnpm test:browser` passes against the new pin (472 tests, 61 files). The merge-order note in the description is resolved — this is ready to go in on its own.

Issue #5658, opened 2026-08-12
OPDS books are erased on restart


chloeroform:
### Bug Description

At restart, books synchronized with OPDS are deleted.

This makes the application completely unusable for me. 🙁 

### Steps to Reproduce

1. Add an OPDS feed with Auto-Download.
2. Wait for book to sync.
3. Swipe off the app.
4. Restart.
5. Books are gone. They are never sync'd again.

I can't reproduce if I reinstall the app, this is maybe a 0.11 -> 0.12 migration issue.

### Expected Behavior

Don't delete the books.

### Operating System

iOS/iPadOS

### OS Version

iPadOS 26

### Readest Version

0.12.1

### Custom CSS

None

Issue #4479, opened 2026-06-05
Bug: OPDS loading and downloading issue


droM4X:
## Bug Description
Maybe related to my earlier issue (#4181), but I’m seeing new problems with this OPDS server.

As I mentioned before, this is an old public/state OPDS server, and it is highly unlikely that it will be changed server-side, so the reader app needs to be more tolerant.

In the web reader `(web.readest.com, Readest 0.11.2, Firefox 151.0)`, clicking almost anything just shows a loading state and then returns me to the same page. This happens on nearly every subpage.

Reader link: https://web.readest.com/opds?url=http%3A%2F%2Fbookserver.mek.oszk.hu%2Fabcrend.atom&id=1780693515850
OPDS server URL: https://bookserver.mek.oszk.hu

On Android `(Android 15, Readest 0.11.4, WebView 148.0)`, the behavior is slightly better: it does not freeze, and the books are listed correctly. However, when I try to download anything, a red notification appears saying that the download failed, including the book link. If I manually open the same link in a browser, it downloads correctly, so the link itself is not broken.

PR #6237, opened 2026-09-17
fix(opds): persist sync progress incrementally so a large first sync can converge

raman325:
## Problem

`syncCatalog` runs a single concurrency pool over **every** pending entry and writes subscription state only after all of them resolve:

```js
const downloadResults = await runWithConcurrency(allItems, DOWNLOAD_CONCURRENCY, downloadAndImport);
// ...only after ALL items resolve:
if (newBooks.length > 0) await onBooksImported?.(newBooks);
state.knownEntryIds = pruneKnownEntryIds([...]);
await saveSubscriptionState(appService, state);
```

A first sync of a large catalog runs for minutes and is a prime target for Android's low-memory killer. Because nothing is persisted until the end, a kill at item N discards **all N** completed imports, and the next run restarts from zero — so the sync can never converge, however often it is retried.

## Evidence

Against a self-hosted OPDS catalog of ~330 books, reverse-proxy access logs recorded **87 successful acquisition `GET`s returning 200** while the app's library ended up with **4 books**.

Device side showed the memory-kill signature rather than a crash:

```
I/lmkd: Reclaim '<pkg>' ... to free 39816kB rss
I/ChimeraAggressivePolicyHandler: executePolicy() - triggerSource: TRIGGER_SOURCE_LMKD
I/ActivityManager: Process com.bilingify.readest (pid ...) has died: fg TOP
```

No `AndroidRuntime` FATAL and an empty `adb logcat -b crash` buffer — an OOM kill, not a crash with a stacktrace. `DOWNLOAD_CONCURRENCY = 3` means three simultaneous download + `importBook` pipelines, each parsing a whole EPUB (files up to 28MB here).

## Change

Download in batches of `PERSIST_BATCH_SIZE` (10) and persist after each one, bounding what a mid-sync kill can discard to a single batch.

The #5658 ordering is deliberately preserved: imported books are still written **before** their entries are marked known, so a kill between those two writes costs a redundant re-download rather than losing library rows. Batch size is a named constant in `types.ts` alongside `DOWNLOAD_CONCURRENCY`, so it is easy to tune.

This does not change per-batch concurrency or add requests; it only moves the persistence boundary.

## Test

Adds a regression test asserting state is written **before** the final item completes. On the unbatched implementation it fails with:

```
AssertionError: expected 30 to be less than 30
```

With the change, it passes.

- `opds-auto-download.test.ts`: **14/14** (13 pre-existing + 1 new)
- Full app suite: **11,099 passed**, 16 skipped, 0 failures
- `tsc --noEmit`: clean
- `biome check` on changed files: clean

## Notes

This addresses the Android download half of #4479, which #4506 explicitly deferred ("That's a distinct problem in the download/acquisition path, not the XML parsing, and is left for a separate change"). It fixes the *convergence* failure; reducing peak memory during sync (e.g. adaptive concurrency) is a separate follow-up.

## Summary by CodeRabbit

* **Reliability**
  * OPDS catalog synchronization now saves subscription progress incrementally during large downloads.
  * If synchronization is interrupted, only the current batch of downloads may need to be retried.
  * Books successfully imported in earlier batches remain available even if a later batch fails.
  * Download progress continues to respect existing retry and failure-handling behavior.

* **Tests**
  * Added coverage for incremental progress, partial synchronization failures, and retry limits.


chrox:
Reviewed, and the core change is right — persisting per batch is the correct shape for this, and keeping the #5658 write order inside the loop was the right call. I pushed a follow-up commit (c64a4fc) with two fixes rather than round-tripping.

**1. Retry counts reset for any entry past the first batch (the serious one).**

```ts
const existingFailed = state.failedEntries.find((fe) => fe.entryId === item.entryId);
const attempts = (existingFailed?.attempts ?? 0) + 1;
```

Each batch ends with `state.failedEntries = updatedFailedEntries`, and `updatedFailedEntries` is seeded with only the **non**-retry-eligible failures. So from batch 2 onward this `find()` searches an array that has already dropped exactly the entries it is looking for, and `attempts` collapses to `1`.

Retry items are appended after the freshly-discovered ones, so they land past the boundary on any catalog with ~10 entries ahead of them. The entry then never reaches `MAX_RETRY_ATTEMPTS`, never enters `knownEntryIds`, and `isRetryEligible` keeps its backoff pinned at the first step instead of doubling — so a 404'd or paywalled book is re-downloaded on every 5-minute background check, forever, and sits in the failed-downloads dialog permanently. Ironic given the PR is about convergence.

Fix: snapshot the counts into a `Map` before the loop and read from that.

**2. A throw in a later batch discards the books earlier batches already committed.**

Before this PR, a throw out of `syncCatalog` meant nothing had been persisted — clean retry next run. Now batches `1..k-1` are on disk *and* marked known, but the exception takes `newBooks` with it, so `allNewBooks.push` never runs and `queueOPDSBookUploads` never sees them. Since their entryIds are in `knownEntryIds`, no later sync rediscovers them: those books never reach the cloud. `syncCatalog` now returns `{ newBooks, state, error }` and the caller rethrows after taking the partial result, which leaves the existing error handling untouched.

**3. Test tightening.** `expect(completedAtFirstSave).toBeLessThan(TOTAL)` passes for an implementation that flushes once at item 29, which bounds nothing. It now asserts one save per batch carrying the entries earned so far. The test also only checked *when* a save happened, never *what* was saved, which is why #1 sailed through it. Two regression tests added for the above, both verified red against the previous head. Also moved it below `freshStatePerLoad` — as the first `it` in the file it was pushing 30 ids into the shared mock state object the later tests read.

Gates on the pushed head: `pnpm test` 11101 passed / 16 skipped, `pnpm lint` and `pnpm format:check` clean.

**Not changed, worth knowing:** `onBooksImported` went from once per catalog to once per batch, and it rewrites the whole library file each time. A 300-book first sync is now ~30 full-library serialisations of a growing library instead of one. That is the price of durability and 10 looks like a reasonable point on the curve, but it is real extra IO during exactly the long sync you are protecting. Worth noting too that batching bounds *loss*, not *memory* — `allItems` and `newBooks` still grow to the full catalog size, so the kill itself is no less likely. Your "adaptive concurrency is a separate follow-up" framing covers that.

PR #6269, opened 2026-09-18
fix(reader): preserve style tag attributes in the style transformer

### LINKED ISSUE #6233: Reflowable EPUB2 book renders blank content pages (due to print-media `.noprint` class) and floated images sized incorrectly in paged mode
### Bug Description

Testing with a real-world reflowable EPUB2 book (InDesign/Sigil-exported, no EPUB3 nav document originally), I found two separate rendering issues in paged mode that do not occur in Calibre's viewer or in a plain browser (Chrome, opening the extracted XHTML directly):

**1. Content pages render completely blank**
Every content XHTML file in this book has `<body class="noprint">`, paired with an inline stylesheet in each page's `<head>`:
    <style type="text/css" media="print">.noprint { display:none }</style>
This rule is scoped to `media="print"` and should have no effect on screen rendering. However, in Readest's paged reading mode, every page rendered completely blank (including the cover) — only a hand-authored `nav.xhtml` (which lacked the `noprint` class) displayed correctly. Removing `class="noprint"` from every content page's `<body>` immediately fixed the blank-page issue. This suggests the paginator may be evaluating `@media print` styles (perhaps as part of pagination/column-width calculation), incorrectly hiding real content.

**2. Text-wrapped (floated) images render far too small in paged/column mode**
The book uses a common float-based text-wrap pattern for small inline figures:
    <div class="w30r"></div>
    /* CSS: */
    div.w30r { width: 30%; float: right; margin: .5em .5em 0 .5em; }
    img.w100 { width: 100%; max-height: 100%; }
This renders correctly (image at ~30% of the reading column width) in Calibre and in plain Chrome (opening the extracted XHTML directly). In Readest's paged mode, the same images render much smaller than expected — consistent with the `30%` width being resolved against a narrower per-column width from the multi-column pagination layout, rather than the full page/viewport width.

### Steps to Reproduce

1. Open the attached EPUB (reflowable EPUB2, float-based text-wrap images, `noprint` class on every page body) in Readest, in Paged mode.
2. Observe blank pages before removing `noprint`.
3. After removing `noprint` (content now visible), observe undersized floated images compared to Calibre / plain-browser rendering of the same XHTML+CSS.

### Expected Behavior

- `@media print`-scoped rules should never affect on-screen paginated rendering.
- Percentage widths on floated elements should resolve consistently with how a single-column/full-page browser view would render them, so book styling looks the same across reading modes.

### Reproduction File

_No response_

### Operating System

iOS/iPadOS

### OS Version

26.5

### Readest Version

0.12.8

### Custom CSS

_No response_

digibrander:
## Problem

Issue #6233 reports two related paginator bugs. This PR fixes the first one: a page whose `<head>` contains a print-only stylesheet, e.g.

```html
<style type="text/css" media="print">.noprint { display: none }</style>
```

paired with `<body class="noprint">`, renders completely blank in Readest's paged mode, even though the rule is scoped to `media="print"` and should never affect on-screen rendering.

## Root cause

`styleTransformer` (`apps/readest-app/src/services/transformers/style.ts`) rewrites every `<style>` block in a section's HTML by matching it with a regex, running its text content through `transformStylesheet`, and then replacing the whole match with a freshly built `<style>${transformed}</style>` tag. That rebuild drops any attributes the original opening tag had, including `media`. A `<style media="print">` block silently becomes an all-media `<style>` block, so a print-only `display: none` rule starts hiding content on screen too.

## Fix

Capture the opening tag's attributes in the match and replay them on the rewritten tag, so `media`, `type`, etc. survive the transform unchanged.

## Scope

Issue #6233 also reports a second, unrelated bug: percentage-width floated images sizing against the per-column width instead of the full page width in paginated mode. That's a separate root cause (likely CSS multi-column percentage-resolution behavior in the rendering engine, not this transformer), needs its own investigation, and isn't addressed by this PR. Scoping this PR to the print-media regression only, which is a clean, isolated, well-understood fix.

## Testing

- Added a regression test to `apps/readest-app/src/__tests__/services/transformers/transformers.test.ts` (`styleTransformer > preserves the media attribute on a <style> block`, plus a companion test for multiple style blocks with different attributes). Confirmed both fail against unmodified `main` (the `media` attribute is dropped) and pass after the fix.
- `pnpm test`: 11224 passed / 16 skipped / 0 failures.
- `pnpm lint` (tsc + biome): clean.
- `pnpm build-web`: succeeded.

Fixes #6233 (print-media blank-page half only — see Scope above).

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved `<style>` element attributes, including `type` and `media`, when stylesheet content is transformed.
  - Print-specific styles remain restricted to print output instead of applying universally.


jadhavgaurav:
Closing and reopening under the correct account (this was opened under the wrong GitHub identity due to a local auth mix-up on my end). The fix itself stands; resubmitting shortly as a fresh PR from the right account.

digibrander:
Closing to reopen under the correct GitHub account (this went out under a local auth mix-up). The fix itself stands and I'll resubmit shortly from the right account.

Issue #5585, opened 2026-08-08
Instant dictionary: Context menu unexpectedly opens after dismissing translation popup


EdEfremoff:
### Bug Description

In the latest version of the app, closing the Instant Dictionary translation popup causes the text context menu to trigger automatically.

<img width="1280" height="2772" alt="Image" src="https://github.com/user-attachments/assets/68fce602-0e47-4003-9f7d-e3a4152ef811" />

<img width="1280" height="2772" alt="Image" src="https://github.com/user-attachments/assets/b9043cb0-6acc-4f11-b90f-d2d2a781d621" />

<img width="1280" height="2772" alt="Image" src="https://github.com/user-attachments/assets/8bfc1b33-da02-46ef-ba12-972efaa8b915" />

### Steps to Reproduce

​Open the Quick Access menu and select Instant dictionary.
​Tap on any word to trigger the translation popup.
​Tap anywhere on the screen to dismiss/hide the translation popup.

### Expected Behavior

The translation popup disappears cleanly without opening any additional menus (matching the behavior of previous app versions).

### Operating System

Android

### OS Version

16

### Readest Version

0.12.1

ChuwuYo:
This conflicts with the fix for #5213. Perhaps we could add a setting to let users choose the selection toolbar behavior after closing a lookup popup.

chrox:
No need to add yet another option. For Instant Dictionary the text selection toolbar should not popup when the lookup popup is dismissed.

PR #6272, opened 2026-09-18
fix(annotator): hand the selection back when the instant dictionary closes

### LINKED ISSUE #6213: Maintain selection after quick action
Hi team,

I just noticed a change in behavior after the most recent update (I think at least) that doesn't fit well with my workflow. In the past, when I had a quick action on (particularly the dictionary one), the text would still stay selected after closing the dictionary. This was great because if I wanted to do something else with the text (ie, copy and paste it into Anki), I still could after reading the definition.

But now, if I want to do anything else with the text (highlight, copy, etc.) the only way is to turn the quick action off entirely and then reselect it.

If you think that this matches most people's workflows, I'd love to revert to the old behavior of maintaining selection after quick action. But if it was an intentional change, I'd like to suggest an additional setting like "Maintain selection after quick action?" with a checkbox in the behaviors tab so the user can pick either.

I've never contributed before here but I'd be happy to help prepare the pr if you'd like.

Thanks so much 🙏

chrox:
Closes #6213.

### What changed

#5730 made the instant dictionary quick action consume its selection — the word was tapped to be looked up, not selected — which fixed two things at once: iOS paints its selection grabbers and blue highlight above web content (i.e. on top of the popup), and dismissing no longer raised the selection toolbar (#5585).

It also took away the only route to the word afterwards. With a quick action armed, re-selecting the word just opens the dictionary again, so highlighting or copying it meant turning the quick action off entirely and re-selecting — exactly what #6213 reports.

The deselect now lasts only as long as the lookup:

- **`useTextSelector.restoreSelectionRange(range)`** puts the range back when the popup closes. The re-add is programmatic under `guardProgrammaticSelection`, so (a) the native grabbers stay away — the engine only draws them for a user-initiated selection, and the app's own handles take over — and (b) the `selectionchange` echo can't republish the selection. It deliberately does not call `setSelection`.
- **`Annotator`** marks the lookup with `instantLookupDeselectedRef` when it deselects, and `handleDismissPopupShowToolbar` hands the range back before deciding, so the dismiss lands on the toolbar like every other lookup (#5213).
- **`quickActionHandled`** is stamped on the handed-back selection and rides along through every republish of it (`{...selection}`). Without it, `handleHighlight` stamping `annotated` read as a fresh selection and re-opened the dictionary — highlighting the word you had just looked up bounced you straight back into the lookup. The once-per-gesture latch does not cover this: Android bridges every touch on the window, including the tap on the toolbar itself, and re-arms it.

The deselect *while the popup is open* is unchanged, so the handles-over-the-dictionary fix stays.

### Verification

Xiaomi 13, dev build of this branch:

1. Long-press a word with Instant Dictionary on → sheet opens, **no grabbers, no system selection menu over it**.
2. Dismiss → the word is selected again with the app's own handles, and the selection toolbar is back.
3. Copy works and spends the selection; Highlight applies and **does not re-open the dictionary** (this failed before the `quickActionHandled` guard — caught on device).
4. A fresh long-press still fires the quick action, so the flag doesn't leak to new selections.

Also:
- `pnpm test` — 11285 tests pass, including new coverage in `AnnotatorLookupSurfaces.test.tsx` (deselect while open, restore + toolbar on dismiss, republish must not re-open the lookup, failed restore dismisses clean) and `useTextSelector-nativeHandles.test.ts` (restore publishes nothing, its `selectionchange` is ignored, collapsed range refused). Both new guards are mutation-checked — removing either fails its test.
- `pnpm lint` clean.
- `e2e/tests/annotation.spec.ts` › `the instant dictionary hands the selection back when it closes (#6213)` passes in the Chromium lane; it replaces the #5585 spec's dismiss half and still pins "no live selection while the popup is up".

### Note on #5585

#5585 asked for the opposite of #6213 — a dismiss that raises no menu. Per maintainer decision this restores unconditionally rather than adding a setting; the toolbar returning on dismiss is now consistent with every other lookup route.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved instant dictionary behavior when dismissing a lookup: the selected word and annotation toolbar are restored correctly.
  - Prevented restored selections from reopening the dictionary lookup unexpectedly.
  - Ensured clean dismissal when selection restoration is unavailable.

- **Tests**
  - Added coverage for selection restoration, dictionary dismissal, and quick-action behavior.


ajdubovoy:
Thank you!

Issue #857, opened 2025-04-10
iOS app has "Application error" when start


atiati123:
## Bug Description

When I run the version 0.9.32 iOS app downloaded from App Store on iPhone with iOS 14.3, I get a black screen with follow error message: 

**Application error: a client-side exception has occurred while loading localhost (see the browser console for more information).**

I can't do anything further on this black screen but to quite the app. 

From App Store description I understand the app is compatible on iOS 13.0 or higher. 
I really love Readest and wish to enjoy the across platform reading experience. Hope this issue can be solved, thanks! 

## Technical inputs

Operating System: IOS 14.3
Readest Version: 0.9.32 IOS app

Issue #592, opened 2025-03-13
FR: Add an option of "keep navigation bar" for Android


doublex640:
It is recommended to add an option for Android, namely "whether to keep navigation bar".
To summarize, there are three forms:
1. Full screen display (current default mode). That is, hide the function bar above and below, and hide the Android navigation bar. The function bar can be displayed in the center of the single screen. (As shown in Figure 1)
Figure 1
![Image](https://github.com/user-attachments/assets/fdc2bd2f-9931-4f3f-9885-2fe660dbff66)

2. Full screen display (can be added). The Android navigation bar below is always displayed. (Figure 2 shows the effect)
Figure 2
![Image](https://github.com/user-attachments/assets/402ffe76-a420-44fd-b593-1494505351b4)

3. Non-full screen display (can be added). Always display the function bar and Android navigation bar.

PR #6301, opened 2026-09-19
feat(opds): stream audiobooks from OPDS catalogs, with a BookOrbit connector (#6224)

### LINKED ISSUE #6224: FR: Stream audiobooks from OPDS catalogs alongside ebooks (e.g. BookOrbit)

I use BookOrbit to host ebooks, comics/PDFs, and audiobooks, and Readest on iPhone and Mac. I would like to browse those titles in one catalog and listen to recorded audiobooks inside Readest, without maintaining a second audiobook server or switching to the server’s web player.

BookOrbit exposes audiobook entries through OPDS, but discovering an audio entry is not the same as being able to stream it in Readest. Readest’s documented Audiobookshelf integration covers streaming through that server’s API; this request is specifically for OPDS-based audiobook playback.

Support playable audiobook entries from authenticated OPDS catalogs, reusing Readest’s existing recorded-audio player where appropriate:

- Recognize supported audio acquisition links (for example MP3 and M4B/M4A where supported by the platform) and offer a Play/Stream action.
- Start playback without first importing the whole audio file as an ebook; support seeking where the server provides HTTP range requests.
- Use the catalog’s existing authentication safely for playback requests.
- Allow ebook/comic and audiobook entries to coexist in the catalog and, ideally, on the Readest bookshelf with clear media-type indicators.
- Preserve listening position in Readest. Chapter navigation and multi-file audiobooks would be useful follow-ups where suitable metadata is available.

Writing listening progress back to BookOrbit would be a separate, optional server-specific integration; I am not assuming OPDS itself provides progress sync.

Keep Readest’s catalog limited to ebooks/comics/PDFs and use BookOrbit’s web player for audio. Audiobookshelf integration is another supported approach, but requires maintaining a separate server alongside BookOrbit.

### Additional Context

A live BookOrbit 2.10.0 feed was inspected. Its MP3 entry currently uses a generic MIME type, for example (sanitized):

```xml
<link rel="http://opds-spec.org/acquisition"
      href="/api/v1/opds/123/download?fileId=456"
      type="application/octet-stream"
      title="MP3" />
```

Correct audio MIME types and any required streaming/manifest metadata may need changes on the BookOrbit side too. This request does not ask Readest to assume every generic binary download is audio.

Source inspection of Readest’s current OPDS page shows acquisition downloads going through the book importer, while its stream handler builds an OPDS-PSE page stream. This is a feed/source review, not a claimed end-to-end iOS audio playback test.

Related but different requests: #3025 (Audiobookshelf integration) and #5775 (web browser as a book source).

Thanks for considering this—one reading/listening app connected to one self-hosted library would be very useful.

chrox:
Closes #6224.

An OPDS catalog that offers audio alongside ebooks can now be played from Readest instead of only downloaded. Two paths land together, because the generic one alone cannot carry a real audiobook.

## Generic OPDS audio

Acquisition links are recognised as audio by MIME type, href extension **or** link title, which recovers catalogs that type every file as `application/octet-stream` but still label it "MP3".

OPDS publishes no durations, so each track's length is read from its own header: a 16 KB ranged read parses ID3v2, the first frame header and the Xing/Info frame. Verified against `ffprobe` on real files, matching to **0.00%** from 8 KB, in place of a whole-file download per track. A track whose duration cannot be established is dropped rather than shown as zero.

Playback goes straight from the device to the catalog and **never through Readest's servers**. A protected catalog therefore plays on native, which can attach credentials, and is refused on web rather than proxied. Credentials already stored for the catalog are reused, so nothing is asked for twice at play time.

## BookOrbit connector

BookOrbit's OPDS download route ignores `Range` (upstream readest/bookorbit#856, PR #857) and it publishes no OPDS-PSE (#592, PR #603), so its audiobooks cannot stream over OPDS at all. Its audiobook API can: one manifest request yields ordered assets, real durations and real chapters, and the asset route answers `Range` with 206.

The connector uses the server row the KOReader sync integration already fills in, rather than asking for the same host and credentials twice, and pushes position back so BookOrbit's own player resumes where Readest left off. BookOrbit stores a position *within one asset*, so it is mapped onto the global timeline in both directions; writes carry `baseRevision` and re-read rather than clobbering a position someone else moved.

Importing a BookOrbit audiobook **auto-pairs** it with the same book's EPUB, so read-along narration works with no wizard. Pairing maps top-level TOC entries only (a story's `I`/`II`/`III` sub-sections are not separate audio chapters) and trims front and back matter *solely* to reach an exact count, since a near miss narrates the wrong text.

## Why blobs rather than a URL

Audio is fetched natively and played from blobs, one track at a time, released when the track changes.

BookOrbit sends `Cross-Origin-Resource-Policy: same-origin`, so a webview refuses its audio from a URL whatever credentials are attached. I built a native `streamurl` scheme to relay the bytes with `Range`, and reverted it: under CEF a custom scheme cannot serve a range that starts anywhere but byte 0 (Chromium 40739128, the bug already behind `RemoteFile.fromNativePath`'s query-carried ranges). Measured against a live server:

```
Range: bytes=0-1023     -> 206, 1024 bytes      ok
Range: bytes=0-         -> 206, whole body      ok
Range: bytes=1000000-…  -> "Failed to fetch"    never reaches the handler
```

Serving whole bodies instead plays linearly but leaves the track unseekable, so resume clamps to 0. A blob is the only source that seeks there, which both resume and mid-chapter narration need. `MediaOverlayClient` gained one optional `loadTrack` hook so a multi-track narration source can carry server paths instead of URLs. No `src-tauri/` changes remain.

A follow-up will make this lazy via MediaSource (`audio/mpeg` + ranged reads reaches first sound in ~100 KB; proven in the desktop build). Not folded in here to keep this reviewable.

## Also fixed

- The OPDS catalog header row never applied its own `justify-start` and `gap-*`: the container was missing `flex`.
- Sorting the library by time remaining ran audiobook seconds through the page-count estimate, turning 7h of listening into 446h and floating every audiobook above 365h books.

## Verification

Unit tests throughout (`pnpm test` green; the mp3 prober is checked against ffprobe fixtures).

End to end in the desktop build against a live BookOrbit server, on a real 10-hour, 284 MB audiobook:

- OPDS import, play, **resume at the stored position** (120.7s) and **seek** to 1800s.
- Progress pushed back to the server (revision advanced, position mapped per asset).
- Read-along from the first mapped chapter (section 2 -> asset 0) **and** from a later one opened via the TOC (section 6 -> asset 4), each loading only its own track.

Start-up cost on that book went from 15.1s/284 MB (whole-book download) to 6.0s via header probing to 3.0s through the connector.

## Known, not addressed here

- m4b audiobooks do not play on the Linux build at all: the CEF runtime ships without AAC (`canPlayType('audio/mp4; codecs="mp4a.40.2"')` is `""`), so they fail with `MEDIA_ERR_SRC_NOT_SUPPORTED` even from a blob. Pre-existing and not specific to this feature; worth its own issue.
- BookOrbit types its OPDS audio links `application/octet-stream`. Worked around here by reading the link title; no upstream issue filed yet.
- A Tauri e2e test for the WebView/CORS facts is written but held back: the harness cannot run on this machine (pre-existing WebKitGTK crash), and its Tauri-HTTP assertion needs an `http:default` grant in the `webdriver-remote` capability that I could not verify.

## Summary by CodeRabbit

- **New Features**
  - Added audiobook playback from OPDS catalogs, including streaming, resume position, and duration handling.
  - Added BookOrbit audiobook support with automatic ebook pairing, synchronized progress, and read-along narration.
  - Added desktop window controls where supported.

- **Improvements**
  - Library audiobook durations now display readable hour, minute, and second units.
  - Added localized playback, catalog, duration, and audiobook error messages.

- **Bug Fixes**
  - Improved OPDS navigation alignment and playback error handling.
  - Improved handling of incomplete audiobook track information.


chrox:
Thanks — worked through all 15. Twelve were real and are fixed in 660ff9bd0; three I'm declining, with reasons below.

## Fixed

**Credential scoping (the critical one, plus the two related).** Track hrefs come out of the feed, so an entry can name any host, and the catalog's `Authorization` / custom headers followed it there. `OpdsAudioAuth` now carries the catalog's origin, and a single `trackHeaders()` gate attaches credentials only on a match — covering `fetchOpdsAudioBlob`, the ranged probe, **and** `probeAuth`, which you were right to call out separately since it is itself a credentialed request against a feed-chosen URL. A foreign host legitimately serving the media authenticates by signed URL, not these headers, so withholding them costs nothing.

**Uncredentialed media probe on an authenticated track.** Fixed, and this one had teeth: it is exactly the 401-then-WebView-Basic-auth-dialog bug this PR set out to kill, which had crept back in as the fallback path. It now reports the duration as unknown instead.

**ID3 tag larger than the probe window.** Fixed as you suggested — read the declared tag size, then a second bounded range at the first audio byte. Embedded cover art makes this common, and the old path fell through to the media element and pulled the whole track.

**No timeout on the ranged probe.** Fixed; the `AbortController` only fired once bytes arrived. Both reads now start a timer before the fetch and clear it in `finally`.

**Dropping an unprobeable track.** Agreed the old behaviour was worse than I'd reasoned: it closed the gap and shifted every later track earlier, losing a chapter and putting every seek and saved position in the wrong place. A fallback duration isn't available (that is the problem), so the session is now refused with a message rather than played silently wrong. New `OpdsAudioIncompleteError` + toast.

**Chapter IDs by array index.** Correct and a nice catch — `manifestChapters` sorts by sequence while `manifest.chapters` keeps the server's order. Now keyed by sequence. Added a test with a reversed manifest that fails against the old code.

**Position dropped while a push is in flight.** Correct: losing a tick is harmless since another follows, but the pause or end that stops playback has nothing after it. The latest position is now queued and flushed when the write completes.

**Unvalidated decoded payload.** Fixed — the string survives in the library across releases, so it is validated at the boundary now instead of throwing on `data.tracks.length`.

**Unstable library hash.** Correct. `md5(filePath)` covered title and author, so a catalog fixing a typo stranded the listening progress on the old row. Hash is now over catalog ID + ordered track URLs; full metadata stays in `filePath`. (I used a named `opdsAudioIdentity()` helper rather than the inline JSON, so the intent is testable.)

12 tests added; `pnpm test` is 11358 passing.

## Not changing

**TLS `acceptInvalidCerts` / `acceptInvalidHostnames`** (2 comments). Real in the abstract, but this is a deliberate project-wide decision, not something this PR introduced — the same flag is set in `opdsReq.ts` (4 sites), `audiobookshelf/client.ts`, `KOSyncClient.ts`, `pseStream.ts` and the pre-existing `BookOrbitClient.ts`. Readest targets self-hosted servers, which commonly have self-signed certificates. Removing it in the new files alone would be inconsistent and would break exactly those setups. If the project wants to revisit it, that is one change across all of them with a per-server trust store, not a drive-by here.

**Require HTTPS before creating the client.** Same reasoning: the app supports plain-HTTP self-hosted servers on a LAN, which is how the BookOrbit instance this was developed against is configured (`http://localhost:13380`). Rejecting `http:` would break working installs. The BookOrbit row is shared with the KOReader sync integration, which already accepts it.

**Don't buffer the complete asset before playback.** Known and deliberate for now — see the PR description. Two constraints: BookOrbit sends `Cross-Origin-Resource-Policy: same-origin`, and under CEF a custom scheme cannot serve a range starting anywhere but byte 0 (Chromium 40739128), so a blob is currently the only seekable source on Linux. The lazy version is a planned follow-up via MediaSource — `audio/mpeg` plus ranged reads reaches first sound in ~100 KB, measured in the desktop build. Not folded in here to keep this reviewable.

Worth noting for the m4b framing in that comment: this build ships without AAC (`canPlayType('audio/mp4; codecs="mp4a.40.2"')` is `""`), so an m4b fails with `MEDIA_ERR_SRC_NOT_SUPPORTED` however the bytes arrive. Separate codec issue.

Issue #6216, opened 2026-09-14
Audiobookshelf playback fails before requesting the audio stream


alucero270:
# Audiobookshelf playback fails before requesting the audio stream

## Bug Description

Audiobookshelf library synchronization works, but every synchronized audiobook fails to play.

Pressing Play displays:

`Audio playback interrupted. Press play to retry.`

Retrying does not start playback.

The issue occurs on both Android and Windows with Readest 0.12.8. The same audiobooks play successfully in Audiobookshelf itself.

Server-side access logs show that Readest successfully requests the item metadata, user details, and a playback session. Audiobookshelf returns HTTP 200 for every request, including `POST /api/items/<item-id>/play`.

Readest does not make a subsequent GET request for an audio file. Playback appears to fail after receiving the successful playback-session response but before requesting the audio stream.

## Steps to Reproduce

1. Connect Readest to a self-hosted Audiobookshelf 2.36.0 server using Settings > Integrations > Audiobookshelf.
2. Enable the audiobook library and select Sync Now.
3. Open any synchronized audiobook.
4. Press Play.
5. Observe the `Audio playback interrupted` error.
6. Press Play again.
7. Observe that playback still does not begin.

The behavior was reproduced with multiple audiobooks. The same files play normally in Audiobookshelf.

## Expected Behavior

After Audiobookshelf returns a successful playback-session response, Readest should request the selected audio track and begin streaming it.

## Reproduction File

Not applicable. The issue occurs with every tested audiobook.

## Operating System

- Android
- Windows

## OS Version

- Samsung Galaxy A55, One UI 8.5
- Windows 11 25H2, build 26200.9168

## Readest Version

0.12.8 on Android and Windows

## Additional Context

- Audiobookshelf version: 2.36.0
- Audiobookshelf is accessed through an HTTPS reverse proxy.
- Relevant sanitized server-side request sequence:
  - `GET /api/items/<item-id>?expanded=1` returned 200.
  - `GET /api/me` returned 200.
  - `POST /api/items/<item-id>/play` returned 200.
  - `POST /api/session/<session-id>/sync` returned 200.
  - No subsequent audio-file GET request was received.
- Custom CSS: None.

alucero270:
Additional troubleshooting narrows this down further:

- The Traefik certificate for `library.home.arpa` was installed in the Windows Current User Trusted Root store.
[...]

PR #6308, opened 2026-09-19
feat(audiobook): stream BookOrbit audio through the loopback media proxy (#6224)

chrox:
Follow-up to #6301. Part of #6224.

#6301 landed BookOrbit audiobooks on whole-track blobs: playing a chapter downloaded 20–45 MB before the first sample. A media element cannot be given BookOrbit's own URL at all — it sends `Cross-Origin-Resource-Policy: same-origin` — and the native URI scheme I tried there had to be reverted, because under CEF a custom scheme cannot serve a range starting anywhere but byte 0 (Chromium 40739128).

**The media proxy from #6216 is the right transport.** It is a real `http://127.0.0.1:<port>/<secret>` origin, so Chromium treats it as ordinary network traffic and `Range` works; it fetches upstream in Rust, where CORP does not apply. Both the player and read-along narration stream through it now, falling back to whole-track blobs only where it is unavailable — the web build and iOS.

## Two things the proxy needed

**It could carry no credential.** Audiobookshelf keeps its token in the upstream URL's own query; BookOrbit's asset route is bearer-only (I measured `?token=` and `?access_token=` — both 401). The allowlist becomes a map of origin → optional `Authorization`, so the credential lives proxy-side rather than in a URL handed to a media element. That also makes refresh free: re-registering with a new token leaves the track URL valid. It matters because the bytes are fetched by the proxy, so an expiring token would surface as a track stopping mid-chapter rather than as a 401 the client could retry.

**It sent no `Cross-Origin-Resource-Policy`.** Every route but `/s` is served `Cross-Origin-Embedder-Policy: require-corp` (`middleware.ts`, for Turso's SharedArrayBuffer), which defaults a cross-origin subresource to same-origin. The WebView refused every proxied response:

```
net::ERR_BLOCKED_BY_RESPONSE.NotSameOriginAfterDefaultedToSameOriginByCoep
```

The proxy exists to serve that page, so it now opts in. **This applies equally to Audiobookshelf playback through the proxy** — worth a look from whoever owns #6216, since it would block ABS streaming on any COEP-isolated route too.

## Also

The pairing dialog described a BookOrbit pairing as "Manage the Audiobookshelf audiobook paired with this ebook." The string predates BookOrbit; it is source-neutral now, and the server's own name is shown in the row below either way. Translated across all 34 locales.

## Verification

Unit tests for the credentialed allowlist, the CORP header, the proxied narration tracks, and the clock choice (a source with a blob loader must still play real URLs directly rather than downloading them whole). `pnpm test`, `pnpm lint`, `pnpm format:check`, `pnpm fmt:check`, `pnpm clippy:check` and `pnpm test:rust` (170 passing) all green.

End to end in the desktop build against a live BookOrbit server:

- Player resumes at its stored position, plays in real time, and **seeks** — the request trace shows non-zero-start ranges answered `206`, e.g. `bytes=14614528-` → `bytes 14614528-28374489/28374490`. That is precisely what the reverted custom scheme could never do.
- Read-along opened at a later chapter ("The Five Orange Pips", section 6) loads that chapter's asset alone (`duration 2479`, not the first asset's 3907) and plays it.

## Not included

A Tauri e2e test for the WebView/CORS facts is written but still held back: the harness cannot run on this machine (pre-existing WebKitGTK crash), and its Tauri-HTTP assertion needs an `http:default` grant in the `webdriver-remote` capability that I could not verify.

## Summary by CodeRabbit

- **New Features**
  - Added streamed audiobook playback with improved seeking through a media proxy.
  - Added support for authenticated audiobook streams with refreshed access credentials.
  - Added fallback playback when proxy streaming is unavailable.

- **Improvements**
  - Updated audiobook pairing descriptions across supported languages to refer to streamed audiobooks.
  - Improved media compatibility for isolated pages.

- **Tests**
  - Added coverage for proxy streaming, authentication, direct playback, seeking, and fallback behavior.


chrox:
Thanks — four of the five inline comments were right and are fixed in 43574a38e. (The review summary says 6 actionable but only 5 posted inline; if one was dropped, happy to look at it.)

## Translations — fixed, and then some

All four were correct, and checking them against the **sibling string in the same dialog** (`Manage the local recording paired with this ebook.`) turned up the same class of drift in six more locales the review didn't flag. That sibling is the house style for this dialog, so every locale now matches it:

| | flagged | also fixed |
| --- | --- | --- |
| ebook term | `ms` (`ebuku` → `e-buku`), `ro` (`acest ebook` → `această carte electronică`) | `fr`, `pt`, `pt-BR` (`ebook` → `e-book`); `th` (Latin `ebook` → `อีบุ๊ก`, which the sibling localises) |
| wording | `el` (`ηχητικού βιβλίου ροής` → `που μεταδίδεται μέσω ροής`), `tr` (`akış` → `akış üzerinden sunulan`) | `fr`, `de`, `pt` verb mood — the siblings are imperative (`Gérez`, `Verwalten Sie`, `Faça a gestão`), these were infinitive |
| pairing verb | — | `ms` → `dipadankan`, `pt-BR` → `vinculado`, `fa` → `جفت‌شده`, matching each sibling |

Both consistency claims were verifiable and checked out before I changed anything: `ms` had `e-buku` 7× vs `ebuku` 1× (mine), `ro` had `carte electronică` 6× vs `ebook` 2×. Also re-ran a guard afterwards: no locale now carries a bare Latin `ebook` in this string where its sibling localises the word.

## Not changing: TLS validation on the proxy client

Real in the abstract, but it is neither new here nor specific to the credential this PR adds.

`build_client()` and its lenient TLS came from #6216 (730f68a6c), and **that lenient policy is the entire point of the feature** — the PR title is "stream audiobook tracks through a loopback media proxy so self-signed servers play". Requiring validation would re-break exactly the case it was written for.

Credentials already traverse that client, too: main forwards Audiobookshelf's access token in the upstream URL, as its own comment notes — "The query, which carries the ABS access token, is never logged." BookOrbit's bearer would likewise have gone over a lenient-TLS connection anyway, since `BookOrbitClient` sets the same flags, as do `opdsReq.ts` (4 sites), `audiobookshelf/client.ts`, `KOSyncClient.ts` and `pseStream.ts`. So this PR moves an existing credential onto an existing path rather than opening a new one.

If the project wants to revisit it, that is one deliberate change across all of those with a per-server trust store — not a drive-by on the one call site that happens to be in this diff.

Issue #6138, opened 2026-09-07
BUG: WebDAV reports 'NETWORK ERROR' when none exists


Doctor-Electron:
### Bug Description

WebDAV sync sometimes fails with a 'NETWORK ERROR' message, but curl shows that the network is fully reachable and the error message is always cleared by a Readest restart.

### Steps to Reproduce

A WebDAV sync.

### Expected Behavior

An error message that describes the actual problem. 
(The user shouldn't be urged to perform a needless check on a functioning Network.

### Operating System

macOS

### OS Version

13

### Readest Version

0.12.1

chrox:
Please provide information about your tested WebDAV services/providers as it might be highly related to the providers.

Issue #6137, opened 2026-09-07
BUG: WebDAV sync fails to complete library transfer


Doctor-Electron:
### Bug Description

I am yet to see a successful WebDAV sync complete from PC to server. 
I have had the same problems on:

- GoogleDrive
- Cloudflare WebDAV
- My personal server (wsgidav with both Cheroot and Gevent)

WebDav sync makes multiple passes (commonly 6-8) but this does not result in a complete transfer of the library and Readest then fails silently with variable numbers of files still to transfer. (This evening it was 180 folders of 195 books, then 130 folders for 195 books)

The problem seems to be around how Readest deals with PDFs and large files.

### Steps to Reproduce

Use any WebDAV server with a mix of large files and PDFs.

### Expected Behavior

A WebDAV sync to complete successfully in one pass.

### Operating System

macOS

### OS Version

13

### Readest Version

0.12.1

chrox:
Please provide information about your tested WebDAV services/providers as it might be highly related to the providers.

Issue #6136, opened 2026-09-07
BUG: WebDAV sync fails with no warning


Doctor-Electron:
### Bug Description

WebDAV sync habitually fails to complete a full sync, even after 6-8 passes. But there is no report to the user that the sync has failed, or how many files remain to be synced. It is up to the user to manually check files on the server.

Occasionally, on the first pass only, there may be a flashed up tally "130 failures; 20 successful" but users need a screen warning that remains until dismissed, so we know if the sync has a) finished and b) been successful or not.

### Steps to Reproduce

Any WebDAV sync will fail to give adequate warning of success or failure.

### Expected Behavior

A dialog that gives success or failure and which displays until dismissed.

### Operating System

macOS

### OS Version

13

### Readest Version

0.12.1 (because 0.12.6 doesn't display PDFs)

chrox:
Please provide information about your tested WebDAV services/providers as it might be highly related to the providers.

Doctor-Electron:
Thanks for the quick reply and I'm very happy to provide as much supporting evidence as I can.
 
However, in my last post about this bug (#6044), you asked for a clean test and a full WebDAV log, which I provided, but you made no reference to it, so I don't know if you found it useful, irrelevant, or didn't look at it.

It's also a bit tricky to provide evidence for something that *isn't* happening: the absence of a warning to say sync has succeeded or failed.  I've said I (sometimes) see a flashed up tally after the initial sync pass, but nothing after that and, crucially, not when sync gives up with an incomplete transfer.

I did over 6 hours of testing yesterday on my wsgidav/cheroot/gevent server. Very happy to share the log if you think that would be useful. AFAIK, there were no errors with a full set of 200 or 201 confirmations.

But, first, to be clear, if everything is working as intended, what would you expect me to see at the end of each sync pass and at the point where sync stops (successfully or otherwise)?

chrox:
Could you test it with Readest 0.12.8?

Doctor-Electron:
Yes, certainly, I will. But can you tell me what i am expected to see?
I'm not just looking for a successful transfer, I'd like to understand if Readest flags a successful or unsuccessful transfer. Without that, I will never know if a transfer has been successful.

chrox:
For the full sync with book files a sync will mirror your local Books directory with the remote Books directory. Incremental sync will only sync the books that has been uploaded (new progress, notes or new book cover).

Doctor-Electron:
[...]

PR #6311, opened 2026-09-19
fix(sync): preserve WebDAV errors and report incomplete syncs

chrox:
WebDAV failures could lose their native error details, appear only in a short toast, and still advance “last synced.” This preserves transfer diagnostics and HTTP status codes, keeps failed runs visible in a dismissible report, and records success only when both book operations and the shared index complete. It also closes native file handles after buffered book/cover reads, including rejected reads.

Addresses #6136 and #6138; fixes a confirmed resource leak relevant to #6137. The reporter’s exact macOS failure was not reproduced, so this does not claim that every cause of incomplete transfers is resolved.

Validation:
- Tauri/CEF over CDP with a real local WsgiDAV server: imported an EPUB and two PDFs, including a 48 MiB PDF; all uploaded files matched their source SHA-256 hashes.
- Injected book-upload and index-upload failures: persistent details, correct success/failure counts, unchanged last-success timestamp, and successful retry. Checked the report in e-ink mode.
- Exercised the NativeFile fallback: the original loader retained 200 handles after 200 reads; the fixed loader retained no additional handles.
- Regression tests cover transfer error preservation, handle cleanup, partial-sync timestamps, failure reporting, summary counts, and keyboard dismissal of queued provider reports. Full unit suite: 11,532 passed, 16 skipped (965 test files passed). TypeScript, Biome lint, and formatting checks pass.

The native UI test used a fresh profile and temporarily disabled the local premium gate; entitlement/authentication and the OS file picker were outside its scope. The gate was restored and the diagnostic route removed before committing.

## Summary by CodeRabbit

- **New Features**
  - Added persistent sync-failure reports with provider details and actionable error messages.
  - Reports remain available across navigation until manually dismissed.
  - Added clearer reporting for library-index upload failures and individual book errors.

- **Bug Fixes**
  - Incomplete synchronizations are no longer marked successful.
  - Improved WebDAV error details, status handling, and retry behavior.
  - Ensured local file handles close after sync operations.

- **Localization**
  - Completed translations for report dismissal and library-index upload failures across supported locales.


chrox:
Addressed the untranslated-string findings in fdf4427ed. Both new report strings now have translations in all 34 affected locales. `pnpm check:translations` passes with no placeholder markers remaining; formatting, TypeScript, lint, and all 11,532 unit tests also pass locally. CI is rerunning on that commit.

